PR #13, opened 2025-02-27
fix: remove binary code

kk-xuhj:


xdotli:
@kk-xuhj shall we also change the schema of the crag benchmark?

PR #136, opened 2026-04-15
fix: resolve YAML config paths against CWD, not config parent (#135)

### LINKED ISSUE #135: `jobs_dir` resolution inconsistent between CLI flags and YAML config
## Problem

`jobs_dir` (and `tasks_dir`, `skills_dir`) resolves differently depending on whether it's passed via CLI flag or YAML config:

- **CLI `--jobs-dir jobs/my-run`** → resolves relative to CWD (`./jobs/my-run/`)
- **YAML `jobs_dir: jobs/my-run`** via `Job.from_yaml()` → resolves relative to config file parent (`<config-dir>/jobs/my-run/`)

This means the same logical value produces different output locations depending on how it's passed.

## Root cause

`Job.from_yaml()` in `job.py` computes `base_dir = path.parent` and prepends it to all relative paths:

```python
# _from_native_yaml, job.py:253
jobs_dir = base_dir / raw.get("jobs_dir", "jobs")

# _from_harbor_yaml, job.py:315
jobs_dir = base_dir / raw.get("jobs_dir", "jobs")
```

But the CLI path in `cli/main.py` passes the raw string directly to `SDK.run()`, which wraps it in `Path(jobs_dir)` — resolved against CWD.

## Repro

```bash
# From repo root:
benchflow job --jobs-dir jobs/my-run ...   # creates ./jobs/my-run/
benchflow job -f experiments/foo/bar.yaml  # with jobs_dir: jobs/my-run → creates experiments/foo/bar/jobs/my-run/
```

## Suggested fix

Resolve YAML relative paths against CWD, not `base_dir`. This also matches [Harbor's](https://github.com/harbor-framework/harbor) behavior, where all relative paths in config are CWD-relative.

EYH0602:
## Summary

- Fixes #135. `Job.from_yaml` previously prepended the YAML file's parent directory to relative `tasks_dir` / `jobs_dir` / `skills_dir` values, while CLI flags resolved them against CWD — same string, different output location.
- Drops the `base_dir` prefix from both `_from_native_yaml` and `_from_harbor_yaml` so YAML paths are passed through verbatim. CLI semantics, the example in `docs/cli-reference.md`, and Harbor convention all expect CWD-relative resolution.
- Absolute paths in YAML still work — `Path` preserves them under `truediv` for free.

## Test plan

- [x] `pytest tests/test_yaml_config.py` — 13 passed (4 new regression tests + 9 existing)
- [x] `pytest tests/` — 479 passed, 1 deselected (live test)
- [x] `ruff format --check src/ tests/` — clean
- [x] `ruff check src/ tests/` — clean
- [x] `ty check src/` — clean

New regression tests cover both native and Harbor formats:
- `test_native_yaml_paths_are_cwd_relative_not_config_relative`
- `test_harbor_yaml_paths_are_cwd_relative_not_config_relative`
- `test_absolute_yaml_paths_are_preserved`
- `test_get_task_dirs_finds_cwd_relative_tasks` (end-to-end check that would have caught #135)

Two pre-existing tests (`test_native_yaml_with_skills_dir`, `test_exclude_parsed`) were updated — they hard-coded the old config-relative behavior.

### DIFF
--- a/src/benchflow/job.py
+++ b/src/benchflow/job.py
@@ -243,14 +243,19 @@ class Job:
 
         # Detect format: Harbor uses "agents" + "datasets", benchflow uses "agent"
         if "agents" in raw or "datasets" in raw:
-            return cls._from_harbor_yaml(raw, path.parent, **kwargs)
-        return cls._from_native_yaml(raw, path.parent, **kwargs)
+            return cls._from_harbor_yaml(raw, **kwargs)
+        return cls._from_native_yaml(raw, **kwargs)
 
     @classmethod
-    def _from_native_yaml(cls, raw: dict, base_dir: Path, **kwargs) -> "Job":
-        """Parse benchflow-native YAML."""
-        tasks_dir = base_dir / raw["tasks_dir"]
-        jobs_dir = base_dir / raw.get("jobs_dir", "jobs")
+    def _from_native_yaml(cls, raw: dict, **kwargs) -> "Job":
+        """Parse benchflow-native YAML.
+
+        Relative paths (``tasks_dir``, ``jobs_dir``, ``skills_dir``) are
+        resolved against the current working directory, matching CLI flag
+        semantics. Absolute paths are passed through unchanged. See #135.
+        """
+        tasks_dir = raw["tasks_dir"]
+        jobs_dir = raw.get("jobs_dir", "jobs")
 
         # Parse prompts — YAML null becomes Python None
         prompts = raw.get("prompts")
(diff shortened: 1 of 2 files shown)


EYH0602:
This is a behavior change for any YAML that uses relative paths with the old config-parent semantics. 

@xdotli two follow-ups intentionally left out of this PR:

1. **Migrate `benchmarks/*.yaml`** — all four configs (`tb2_single-codex-gpt54.yaml`, `tb2_multiturn-codex-gpt54.yaml`, `skillsbench-codex-gpt54.yaml`, `skillsbench-claude-glm5.yaml`) use `../.ref/...` and `../jobs/...`, which work today only because `benchmarks/../.ref/...` happens to resolve to `.ref/...` from repo root. Under the new semantics they need to drop the `../` prefix (assuming run from repo root, which `docs/getting-started.md` already documents).
2. **`CHANGELOG.md` entry under `[Unreleased]`** — flag this as a breaking change for users with YAML configs that depend on the old config-relative resolution.

Please let me know if these two sounds good in this PR.

xdotli:
@claude review this

EYH0602:
resolved by #138

kywch:
I'm sorry I missed the PR.

PR #154, opened 2026-04-18
fix: preserve provider prefix in Harbor YAML model parsing

### LINKED ISSUE #151: _from_harbor_yaml strips provider prefix from model_name, breaking provider detection
## Bug

`_from_harbor_yaml` in `job.py` naively strips the first `/`-delimited component from `model_name`:

```python
model = agent_cfg.get("model_name", "")
if "/" in model:
    model = model.split("/", 1)[1]
```

This runs **before** `resolve_provider_env()` / `find_provider()`, so provider prefixes like `vllm/` are lost before the provider system ever sees them.

## Example

```yaml
agents:
  - name: pi-acp
    model_name: vllm/Qwen/Qwen3.5-35B-A3B
```

| Stage | Value |
|-------|-------|
| YAML input | `vllm/Qwen/Qwen3.5-35B-A3B` |
| After `_from_harbor_yaml` | `Qwen/Qwen3.5-35B-A3B` (vllm/ lost) |
| `find_provider("Qwen/...")` | `None` — no `qwen/` provider registered |
| `strip_provider_prefix` generic fallback | `Qwen3.5-35B-A3B` (wrong — HF model ID is `Qwen/Qwen3.5-35B-A3B`) |
| `BENCHFLOW_PROVIDER_MODEL` | `Qwen3.5-35B-A3B` (should be `Qwen/Qwen3.5-35B-A3B`) |

vLLM expects the full HuggingFace model ID (`Qwen/Qwen3.5-35B-A3B`), so API calls fail.

## Impact

All vLLM experiments launched via Harbor-format YAML silently get wrong model names and no provider-specific handling. Observed as 0/58 scores on InfiniteBench with pi-acp + vLLM (Qwen3.5-35B-A3B).

## Proposed fix

Remove the stripping in `_from_harbor_yaml`. Let `strip_provider_prefix` (which knows about registered providers) handle prefix stripping downstream:

```python
# Before (broken):
model = agent_cfg.get("model_name", "")
if "/" in model:
    model = model.split("/", 1)[1]

# After (fixed):
model = agent_cfg.get("model_name", "")
```

This is safe for all model formats — `strip_provider_prefix` already handles `anthropic/claude-...`, `openai/gpt-...`, `vllm/Qwen/...`, etc.

EYH0602:
## Summary
- Removed naive `split("/", 1)[1]` in `_from_harbor_yaml` that stripped provider prefixes (e.g. `vllm/`) before the provider system could resolve them
- This caused `vllm/Qwen/Qwen3.5-35B-A3B` to lose both the provider prefix and the HuggingFace org, producing the wrong model ID downstream

## Test plan
- [x] Existing `test_from_harbor_yaml` updated to expect full `anthropic/claude-haiku-4-5-20251001` in `config.model`
- [x] New `test_harbor_yaml_preserves_provider_prefix` covers the vllm multi-slash case (`vllm/Qwen/Qwen3.5-35B-A3B`)
- [x] All 12 yaml config tests pass

Fixes #151

### DIFF
--- a/src/benchflow/job.py
+++ b/src/benchflow/job.py
@@ -283,11 +283,8 @@ class Job:
         agent_cfg = agents[0] if agents else {}
         agent_name = agent_cfg.get("name", DEFAULT_AGENT)
 
-        # Model — Harbor uses "anthropic/model-name" format
-        model = agent_cfg.get("model_name", "")
-        if "/" in model:
-            model = model.split("/", 1)[1]
-        model = model or DEFAULT_MODEL
+        # Model — keep provider prefix intact for downstream resolution
+        model = agent_cfg.get("model_name", "") or DEFAULT_MODEL
 
         # Environment
         env_cfg = raw.get("environment", {})
--- a/tests/test_yaml_config.py
+++ b/tests/test_yaml_config.py
@@ -80,13 +80,32 @@ def test_from_harbor_yaml(harbor_yaml):
     cfg = job._config
 
     assert cfg.agent == "claude-agent-acp"
-    assert cfg.model == "claude-haiku-4-5-20251001"
+    assert cfg.model == "anthropic/claude-haiku-4-5-20251001"
     assert cfg.environment == "daytona"
     assert cfg.concurrency == 8
     assert cfg.retry.max_retries == 1  # n_attempts=2 → max_retries=1
     assert cfg.agent_env.get("ANTHROPIC_API_KEY") == "test-key"
     assert job._tasks_dir == Path("tasks")
     assert job._jobs_dir == Path("output")
+
+
+def test_harbor_yaml_preserves_provider_prefix(tmp_path):
+    """Provider prefix must survive _from_harbor_yaml for downstream resolution."""
+    tasks = tmp_path / "tasks" / "task-a"
+    tasks.mkdir(parents=True)
+    (tasks / "task.toml").write_text('version = "1.0"')
+
+    config = tmp_path / "config.yaml"
+    config.write_text("""
+agents:
+  - name: pi-acp
+    model_name: vllm/Qwen/Qwen3.5-35B-A3B
+datasets:
+  - path: tasks
+""")
+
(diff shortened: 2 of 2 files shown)


EYH0602 (commented):
## Code Review Summary
Reviewed 2 files, 16 lines changed across 3 review aspects (bug detection, test coverage, guidelines compliance).

**No issues found at confidence ≥75. LGTM.**

### Analysis

**Bug Detection:** The new `agent_cfg.get("model_name", "") or DEFAULT_MODEL` correctly handles all edge cases — missing key (returns `""`→default), `None` value (falsy→default), empty string (falsy→default), and prefixed model strings (passed through verbatim). The downstream provider resolution in `providers.py` (`find_provider`, `strip_provider_prefix`) is built to receive prefixed model strings.

**Test Coverage:** The updated assertion in `test_from_harbor_yaml` (line 83) is justified — the fixture input was always `anthropic/claude-haiku-4-5-20251001`, and the old assertion tested the now-removed stripping. The new `test_harbor_yaml_preserves_provider_prefix` covers the distinct multi-slash case (`vllm/Qwen/Qwen3.5-35B-A3B`) from the bug report. The missing-key default is covered by pre-existing `test_from_harbor_yaml_defaults`.

**Guidelines Compliance:** Minimal fix (4 lines removed, 1 added). No new hardcoded paths or special cases. Test update tracks the behavioral change per CLAUDE.md convention. PR is on a feature branch.

### Positive Observations
- Clean, minimal fix — removes complexity rather than adding it
- New test targets the exact failure mode from #151 (multi-segment HuggingFace path)
- Good use of `or` idiom to collapse the fallback logic

xdotli:
@EYH0602 Merged! ✅

**Review summary:** Clean minimal fix — removes naive `/` stripping in `_from_harbor_yaml`, correctly lets `strip_provider_prefix` handle it downstream. No callers depended on the old (stripped) behavior. Tests verified locally (54/54 passed).

PR #155, opened 2026-04-18
fix: preserve provider prefix in set_model for pi-acp

### LINKED ISSUE #153: _acp_run strip_provider_prefix before set_model breaks pi-acp with HuggingFace model IDs
## Bug

`_acp_run.py` strips the BenchFlow provider prefix before calling `set_model`:

```python
# _acp_run.py:85-87
acp_model_id = strip_provider_prefix(model)
await acp_client.set_model(acp_model_id)
```

For `vllm/Qwen/Qwen3.5-35B-A3B`, this strips `vllm/` and sends `set_model("Qwen/Qwen3.5-35B-A3B")` to pi-acp. This breaks pi-acp's provider routing.

## Root cause: pi-acp splits on the first `/` to extract provider

[`pi-acp` (`agent.ts:968-999`)](https://github.com/svkozak/pi-acp/blob/main/src/acp/agent.ts) handles `set_model` by splitting on the first `/`:

```typescript
async unstable_setSessionModel(params: { sessionId: string; modelId: string }) {
    // Accept either:
    //  - "provider/model" (preferred, matches how we advertise)
    //  - "model" (fallback, we try to resolve via available models)
    if (params.modelId.includes('/')) {
      const [p, ...rest] = params.modelId.split('/')
      provider = p
      modelId = rest.join('/')
    } else {
      modelId = params.modelId
    }
    // ...
    await session.proc.setModel(provider, modelId)
}
```

Pi's internal RPC ([`rpc-types.ts`](https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/src/modes/rpc/rpc-types.ts)) then does an exact two-field match:

```typescript
const model = models.find(m => m.provider === command.provider && m.id === command.modelId);
```

## What happens

| BenchFlow sends | pi-acp parses | Pi looks up | Result |
|---|---|---|---|
| `vllm/Qwen/Qwen3.5-35B-A3B` (correct) | provider=`vllm`, modelId=`Qwen/Qwen3.5-35B-A3B` | `provider="vllm" && id="Qwen/Qwen3.5-35B-A3B"` | **Match** (if models.json has vllm provider) |
| `Qwen/Qwen3.5-35B-A3B` (after strip) | provider=`Qwen`, modelId=`Qwen3.5-35B-A3B` | `provider="Qwen" && id="Qwen3.5-35B-A3B"` | **No match** → error -32603 |

The `/` in the HuggingFace model name (`Qwen/Qwen3.5-35B-A3B`) is misinterpreted as a provider separator when the real provider prefix (`vllm/`) has already been stripped.

## Scope

This affects any ACP agent whose `set_model` handler uses `provider/model` format and any model with a `/` in its bare ID (all HuggingFace model names: `Qwen/...`, `meta-llama/...`, `mistralai/...`, etc.).

`claude-agent-acp` and `codex-acp` are unaffected because their model IDs don't contain `/` after provider stripping (e.g. `claude-haiku-4-5-20251001`, `gpt-5-mini`).

## Proposed fix

Skip `strip_provider_prefix` for agents that understand BenchFlow provider prefixes (currently pi-acp). Other agents (claude-agent-acp, codex-acp) don't understand `vllm/` prefixes, so stripping is correct for them.

Alternatively, pass the provider and model ID as separate fields in the ACP `set_model` call rather than relying on string parsing conventions.

## Relation to #151 and #152

This is the third bug in the pi-acp + vLLM chain. Even after #151 (preserve `vllm/` prefix through `_from_harbor_yaml`) and #152 (generate `models.json` so Pi knows about the vLLM endpoint), `strip_provider_prefix` removes `vllm/` before `set_model` — causing Pi to misparse the remaining HuggingFace model ID.

EYH0602:
## Summary

- Add `preserve_provider_prefix` flag to `AgentConfig` (default `False`) to control whether `strip_provider_prefix` is applied before ACP `set_model`
- Set `preserve_provider_prefix=True` for `pi-acp`, whose `set_model` handler splits on the first `/` to extract the provider
- Skip stripping in `_acp_run.py` when the flag is set, so `vllm/Qwen/Qwen3.5-35B-A3B` reaches pi-acp intact instead of being misparsed as `provider=Qwen`

Closes #153

## Test plan

- [x] All 492 unit tests pass (1 pre-existing failure in `test_from_harbor_yaml` unrelated to this change)
- [x] `ty check` passes on modified files
- [ ] Manual verification with pi-acp + vLLM + HuggingFace model ID (e.g. `vllm/Qwen/Qwen3.5-35B-A3B`)

### DIFF
--- a/src/benchflow/_acp_run.py
+++ b/src/benchflow/_acp_run.py
@@ -25,6 +25,7 @@ from benchflow._sandbox import build_priv_drop_cmd
 from benchflow._trajectory import _capture_session_trajectory
 from benchflow.acp.client import ACPClient
 from benchflow.acp.container_transport import ContainerTransport
+from benchflow.agents.registry import AgentConfig
 from benchflow.process import DaytonaProcess, DockerProcess
 
 logger = logging.getLogger(__name__)
@@ -40,6 +41,7 @@ async def connect_acp(
     trial_dir: Path,
     environment: str,
     agent_cwd: str,
+    agent_cfg: AgentConfig | None = None,
 ) -> tuple[ACPClient, object, str]:
     """Create ACP transport, connect, init session, set model. Return (client, session, agent_name)."""
     # Resolve agent binary path for non-docker environments
@@ -83,9 +85,7 @@ async def connect_acp(
 
     if model:
         from benchflow.agents.providers import strip_provider_prefix
-        from benchflow.agents.registry import AGENTS
 
-        agent_cfg = AGENTS.get(agent)
         if agent_cfg and agent_cfg.preserve_provider_prefix:
             acp_model_id = model
         else:
--- a/src/benchflow/_acp_run.py
+++ b/src/benchflow/_acp_run.py
@@ -83,8 +83,13 @@ async def connect_acp(
 
     if model:
         from benchflow.agents.providers import strip_provider_prefix
+        from benchflow.agents.registry import AGENTS
 
-        acp_model_id = strip_provider_prefix(model)
+        agent_cfg = AGENTS.get(agent)
+        if agent_cfg and agent_cfg.preserve_provider_prefix:
+            acp_model_id = model
+        else:
+            acp_model_id = strip_provider_prefix(model)
         try:
             await asyncio.wait_for(acp_client.set_model(acp_model_id), timeout=60)
             logger.info(f"Model set to: {acp_model_id} (from {model})")
--- a/src/benchflow/agents/registry.py
+++ b/src/benchflow/agents/registry.py
@@ -129,6 +129,12 @@ class AgentConfig:
     home_dirs: list[str] = field(default_factory=list)
     # Extra dot-dirs under $HOME to copy to sandbox user (for dirs not
     # derivable from skill_paths or credential_files, e.g. ".openclaw").
+    preserve_provider_prefix: bool = False
+    # When True, ACP set_model receives the full model string including the
+    # BenchFlow provider prefix (e.g. "vllm/Qwen/Qwen3.5-35B-A3B").
+    # When False (default), the prefix is stripped first (e.g. "Qwen/Qwen3.5-35B-A3B").
+    # Agents whose set_model handler uses "provider/model" routing (e.g. pi-acp)
+    # need the prefix to avoid misparsing HuggingFace model IDs that contain "/".
     subscription_auth: SubscriptionAuth | None = None
     # Host CLI login that can substitute for an API key (e.g. OAuth tokens
     # from `claude login`). Detected automatically; API keys take precedence.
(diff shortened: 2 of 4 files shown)


src/benchflow/_acp_run.py
```diff
@@ -83,8 +83,13 @@ async def connect_acp(
 
     if model:
         from benchflow.agents.providers import strip_provider_prefix
+        from benchflow.agents.registry import AGENTS
 
-        acp_model_id = strip_provider_prefix(model)
+        agent_cfg = AGENTS.get(agent)
```
EYH0602: **Bug (confidence: 92):** `AGENTS.get(agent)` uses the raw `agent` string without normalizing it. The established pattern in `_agent_setup.py:35-36` is:

```python
agent_base = agent.split()[0]
agent_cfg = AGENTS.get(agent_base)
```

If `agent` ever contains extra arguments (e.g. `"pi-acp --some-flag"`), `AGENTS.get(agent)` returns `None`, the `preserve_provider_prefix` guard is silently bypassed, and the prefix gets stripped — re-introducing the exact bug this PR fixes.

Alternatively, since `agent_cfg` is already resolved at `sdk.py:539` before the `connect_acp` call, consider threading it through as a parameter instead of re-looking it up:

```python
async def connect_acp(
    ..., agent_cfg: AgentConfig | None, ...
):
    ...
    if agent_cfg and agent_cfg.preserve_provider_prefix:
        acp_model_id = model
    else:
        acp_model_id = strip_provider_prefix(model)
```

EYH0602 (commented):
## Review Summary
Reviewed 2 files, 14 lines changed across 3 review aspects (bug detection, test coverage, guidelines compliance).

## Critical Issues (confidence 90-100)

See inline comment on `_acp_run.py:88`.

## Important Issues (confidence 75-89)

**No test coverage for the new branching logic.** The `preserve_provider_prefix` flag creates three code paths (unknown agent → strip, known agent with flag=False → strip, known agent with flag=True → preserve), and none are tested. A future refactor could silently re-introduce the bug with no failing test to catch it. The model-ID selection logic is pure and testable without ACP transport — a unit test patching `AGENTS` and asserting the value passed to `set_model` would be sufficient.

## Positive Observations
- Clean "registry over hardcode" design: the flag is data-driven via `AgentConfig`, not agent-name-driven.
- Good inline documentation of the field purpose and examples.
- Default `False` preserves backward compatibility for all existing agents.

src/benchflow/_acp_run.py
EYH0602: Fixed in 54d7bff. Threaded `agent_cfg` from `sdk.py` into `connect_acp` as a parameter instead of re-looking it up. Also added 4 parametrized tests covering preserve=True, preserve=False, and agent_cfg=None paths.

xdotli:
@EYH0602 Merged! ✅

**Review summary:** `preserve_provider_prefix` flag approach is clean and well-tested (62/62 passed locally). 

**One minor note for follow-up:** Three other callers of `connect_acp` don't pass `agent_cfg` yet — `experiments/reviewer_ablation.py`, `tests/conformance/proof_multi_agent.py`, `benchmarks/followup-bench/runner.py`. They default to `None` so they won't crash, but if those scripts are ever run with pi-acp + HuggingFace models, the prefix will still get stripped. Low priority since they're experimental scripts.

PR #156, opened 2026-04-18
fix: make pi-acp multi-protocol so OpenAI-compatible providers work

### LINKED ISSUE #152: pi-acp env_mapping hardcodes Anthropic vars, breaks OpenAI-compatible providers (vLLM)
## Bug

The `pi-acp` registry entry has a static `env_mapping` that always translates `BENCHFLOW_PROVIDER_*` into `ANTHROPIC_*` env vars:

```python
# registry.py, pi-acp entry
env_mapping={
    "BENCHFLOW_PROVIDER_BASE_URL": "ANTHROPIC_BASE_URL",
    "BENCHFLOW_PROVIDER_API_KEY": "ANTHROPIC_AUTH_TOKEN",
    "BENCHFLOW_PROVIDER_MODEL": "ANTHROPIC_MODEL",
},
```

When the provider is vLLM (or any OpenAI-compatible endpoint), Pi receives `ANTHROPIC_BASE_URL` pointing to a vLLM server and sends **Anthropic Messages API** requests. vLLM only speaks the **OpenAI Chat Completions API** — completely different wire format. Pi's first API call fails silently: 0 tool calls, <0.5s execution, no error reported.

## Reproduction

```yaml
agents:
  - name: pi-acp
    model_name: vllm/Qwen/Qwen3.5-35B-A3B
environment:
  env:
    - BENCHFLOW_PROVIDER_BASE_URL=http://172.17.0.1:8080/v1
    - BENCHFLOW_PROVIDER_API_KEY=unused
```

Result: Pi exits immediately with 0 tool calls. Score: 0.

## Root cause

`env_mapping` is static per agent, but the correct mapping depends on the **provider's wire protocol**:
- Anthropic provider → `ANTHROPIC_BASE_URL` (correct — Anthropic Messages API)
- vLLM/OpenAI provider → Pi needs `models.json` with `"api": "openai-completions"`, not just env vars

Pi uses `models.json` to decide which wire protocol to use for each provider. Env vars alone (`ANTHROPIC_BASE_URL`) can only configure the Anthropic path. For OpenAI-compatible endpoints, Pi requires a `models.json` entry with the correct `api` field.

## Relation to #151

This bug compounds with #151 (`_from_harbor_yaml` stripping the `vllm/` prefix). Even after #151 is fixed so that `find_provider` detects vLLM, the `env_mapping` still translates everything to `ANTHROPIC_*` — which is the wrong protocol for vLLM.

## Possible approaches

1. **Dynamic env_mapping**: make `env_mapping` provider-aware so it maps to different agent vars based on the detected provider's `api_protocol`.
2. **Agent-side shim**: generate Pi's `models.json` at launch time from the `BENCHFLOW_PROVIDER_*` vars, selecting the correct `api` field based on `BENCHFLOW_PROVIDER_PROTOCOL`. This is what we're doing as a local workaround (wrapper script that intercepts `pi-acp` launch).
3. **Protocol-keyed env_mapping**: allow agents to declare separate mappings per `api_protocol`, e.g. `env_mapping_by_protocol={"openai-completions": {...}, "anthropic-messages": {...}}`.

Note: `claude-agent-acp` has the same static `env_mapping` pattern but only supports Anthropic wire format natively, so it doesn't hit this issue. Any future agent that supports multiple wire protocols via env vars would face the same problem.

EYH0602:
## Summary

- Adds a launch wrapper (`pi_acp_launcher.py`) that bridges `BENCHFLOW_PROVIDER_*` env vars to Pi's native config based on the detected protocol:
  - **Anthropic**: sets `ANTHROPIC_*` env vars (unchanged behavior)
  - **OpenAI-compatible** (vLLM, etc.): generates `~/.pi/agent/models.json` with the provider registered as `openai-completions`
- Changes pi-acp registry: `api_protocol=""` (multi-protocol), `requires_env=[]` (inferred at runtime), `env_mapping={}` (launcher handles it)

Closes #152

## Test plan

- [x] All 494 unit tests pass
- [x] ruff check + format clean
- [ ] Manual: run pi-acp with `vllm/` model against a local vLLM server — verify models.json is generated and Pi sends OpenAI Chat Completions requests
- [ ] Manual: run pi-acp with an Anthropic model — verify ANTHROPIC_* env vars are set (regression check)

### DIFF
--- a/src/benchflow/agents/pi_acp_launcher.py
+++ b/src/benchflow/agents/pi_acp_launcher.py
@@ -57,7 +57,10 @@ def setup_provider():
                 existing.setdefault("providers", {}).update(config["providers"])
                 config = existing
             except (json.JSONDecodeError, OSError):
-                pass
+                print(
+                    f"Warning: could not parse {models_path}, overwriting.",
+                    file=sys.stderr,
+                )
         models_path.write_text(json.dumps(config, indent=2))
     else:
         # Anthropic mode — set native env vars that Pi reads directly
--- a/src/benchflow/agents/pi_acp_launcher.py
+++ b/src/benchflow/agents/pi_acp_launcher.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python3
+"""Launch wrapper for pi-acp — bridges BENCHFLOW_PROVIDER_* to Pi config.
+
+Pi natively reads ANTHROPIC_* env vars for Anthropic providers. For
+OpenAI-compatible providers (vLLM, etc.), Pi requires a ``models.json``
+config file at ``~/.pi/agent/models.json`` that declares the provider's
+wire protocol.  This wrapper generates it from BENCHFLOW_PROVIDER_* env
+vars injected by the SDK, then execs ``pi-acp``.
+
+See https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/docs/models.md
+"""
+
+import json
+import os
+import sys
+from pathlib import Path
+
+
+def setup_provider():
+    """Configure Pi for the detected provider protocol."""
+    protocol = os.environ.get("BENCHFLOW_PROVIDER_PROTOCOL", "")
+    base_url = os.environ.get("BENCHFLOW_PROVIDER_BASE_URL", "")
+    api_key = os.environ.get("BENCHFLOW_PROVIDER_API_KEY", "")
+    model = os.environ.get("BENCHFLOW_PROVIDER_MODEL", "")
+    provider_name = os.environ.get("BENCHFLOW_PROVIDER_NAME", "custom")
+
+    if protocol == "openai-completions" and base_url:
+        # Pi uses ~/.pi/agent/models.json to discover non-Anthropic providers.
+        # Register the provider so Pi routes API calls through the OpenAI
+        # Chat Completions wire format instead of Anthropic Messages.
+        config = {
+            "providers": {
+                provider_name: {
+                    "baseUrl": base_url,
+                    "api": "openai-completions",
+                    "apiKey": api_key or "unused",
+                    "models": [
+                        {
+                            "id": model,
+                            "name": model,
+                            "reasoning": False,
+                            "input": ["text"],
+                            "contextWindow": 128000,
+                            "maxTokens": 16384,
+                        }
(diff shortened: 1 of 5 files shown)


EYH0602 (commented):
## Self-Review Summary
Reviewed 4 files, ~100 lines changed across 4 aspects (bug detection, test coverage, error handling, guidelines compliance).

**No blocking issues found.** Two minor observations:

### pi_acp_launcher.py — No unit tests for `setup_provider()`
The five code paths (OpenAI write, merge with existing, corrupt fallback, Anthropic env vars, no-op) have zero unit tests. The `openclaw_acp_shim.py` setup functions follow the same untested pattern, so this is consistent — but the merge logic (line 57: `existing.setdefault("providers", {}).update(...)`) is the key correctness claim and would benefit from a targeted test.

Not blocking — manual test plan covers it and this follows the existing codebase pattern.

### pi_acp_launcher.py:59 — Silent overwrite of corrupt `models.json`
When an existing `models.json` is unparseable, `except (json.JSONDecodeError, OSError): pass` silently falls through and overwrites with only the new provider entry. Behavior is correct (can't merge with corrupt JSON), but a `print(..., file=sys.stderr)` warning would help operators debug vanishing providers.

### Positive Observations
- Clean separation: launcher handles protocol-dependent config, registry stays declarative
- Merge-with-existing logic preserves user-added providers
- Test updates match the new shape without rewriting unrelated tests
- Follows CLAUDE.md conventions: minimal fix, registry-only change, established shim pattern

xdotli:
@EYH0602 Merged! ✅

**Review summary:** Well-structured, follows the `_OPENCLAW_SHIM` pattern exactly. Full test suite passed (499/499). Also tested combined merge of all 3 PRs (#154 + #155 + #156): 504 passed, ruff + ty all clean.

**Minor notes for future consideration:**
- Hardcoded `contextWindow` (128k) and `maxTokens` (16k) in the generated `models.json` are reasonable defaults but could be made configurable via env vars if users report issues with different models.
- The `apiKey: "unused"` string may still be sent as a Bearer token to vLLM — worth verifying Pi strips it when auth isn't needed.

Nice work on the whole chain — the vLLM + pi-acp pipeline should be unblocked now. 🎉

Issue #204, opened 2026-04-25
Skills loading model: how benchflow + claude-agent-acp compares to harbor + openhands_sdk


Yiminnn:
# Skills loading model: how benchflow + claude-agent-acp compares to harbor + openhands_sdk

## TL;DR

Both benchflow and harbor mount task-bundled `SKILL.md` files into the agent's
home directory. From there the two diverge:

- **benchflow + claude-agent-acp** delegates to the official Claude Code skill
  mechanism: each `SKILL.md` becomes a `Skill` tool whose **description** field
  is the frontmatter `description`, surfaced in the model's tool list.
  The skill **body** loads only when the model calls the Skill tool.
  This is *progressive disclosure* and matches `code.claude.com/docs/en/skills`.

- **harbor + openhands_sdk** delegates to OpenHands SDK's `AgentContext(skills=)`,
  which is OpenHands SDK's native skill mechanism — *not* something harbor wrote
  for SkillsBench. For modern AgentSkills-format SKILL.md files it's also
  progressive (descriptions in `<available_skills>` block of the system prompt).
  For *legacy* OpenHands skills with `trigger=None`, the **full body is dumped
  unconditionally into `<REPO_CONTEXT>`** of the system prompt every turn — this
  is where the "prompt injection" feel comes from.

So benchflow conforms to the Anthropic Skill design end-to-end. Harbor's path
inherits OpenHands SDK's design, which is *mostly* progressive but has a legacy
escape hatch that bypasses the model.

## 1. Does benchflow put the SKILL.md frontmatter into the prompt?

**Yes — transitively, via the same path Claude Code itself uses.**

benchflow's job stops at:

```python
# benchflow/agents/registry.py: claude-agent-acp config
skill_paths=["$HOME/.claude/skills"],
```

i.e. "deploy SKILL.md files here." It does not touch the prompt.

Then `claude-agent-acp` (the npm package `@zed-industries/claude-agent-acp`,
which wraps the same Anthropic Claude Code SDK shipped to end users) auto-scans
that directory at startup, parses each `SKILL.md` frontmatter, and registers
the `description` field as the description of a `Skill(<name>)` tool variant.
Per `code.claude.com/docs/en/skills`:

> In a regular session, skill descriptions are loaded into context so Claude
> knows what's available, but full skill content only loads when invoked.

The frontmatter description ends up in the system prompt's tool-list section,
capped at 1,536 chars per skill, with the global tool-list budget at 1% of the
[...]

PR #207, opened 2026-04-26
feat: add optional skill nudge to prompt resolution

zjw49246:
## Summary

- Add optional `BENCHFLOW_SKILL_NUDGE` env var that prepends skill discovery hints to task instructions
- Three modes: `name`, `description`, `full` — default off (no behavior change)
- Skill path auto-resolved per agent from registry (e.g. `~/.claude/skills`, `~/.agents/skills`)

## Problem

BenchFlow uses each agent's native skill mechanism (progressive disclosure for claude-agent-acp), but agents often fail to discover and invoke skills. This creates a parity gap with Harbor, which injects skill content directly into the system prompt. See #204 for full analysis.

## Solution

When `BENCHFLOW_SKILL_NUDGE=name` is set, prepend one line to the instruction:

```
Skills available at ~/.claude/skills: xlsx, pdf-editing. Read them before starting.
```

This is a pure text prepend — no agent-specific mechanism needed, works across all agents.

## Experiment Results (10 tasks, single-run)

| Agent / Model | No nudge | With nudge=name |
|---|---|---|
| Claude Opus 4.7 | 50% | **80%** |
| Codex GPT 5.5 | 70% | **80%** |

Trajectory analysis shows Claude reads SKILL.md files directly after being told the path. Without the nudge, it ignores skills entirely on most tasks.

## Usage

```bash
# CLI
BENCHFLOW_SKILL_NUDGE=name bench eval create -t tasks/my-task -a claude-agent-acp -s tasks/my-task/environment/skills/

# Python SDK
await sdk.run(..., agent_env={"BENCHFLOW_SKILL_NUDGE": "name"})

# YAML config
environment:
  env:
    - BENCHFLOW_SKILL_NUDGE=name
```

## Modes

| Value | Effect |
|---|---|
| `""` (default) | Off — no change |
| `"name"` | `Skills available at <path>: X, Y. Read them before starting.` |
| `"description"` | Names + YAML frontmatter descriptions |
| `"full"` | Entire SKILL.md content prepended |

## Test plan

- [x] Default behavior unchanged (empty string = no nudge)
- [x] `name` mode prepends skill names with correct agent-specific path
- [x] `description` mode includes frontmatter descriptions
- [x] `full` mode injects complete SKILL.md content
- [x] Invalid values silently ignored
- [x] All 5 registered agents resolve correct paths

---

<a href="https://app.devin.ai/review/benchflow-ai/benchflow/pull/207" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/benchflow/sdk.py
+++ b/src/benchflow/sdk.py
@@ -330,13 +330,68 @@ class SDK:
 
     @staticmethod
     def _resolve_prompts(
-        task_path: Path, prompts: list[str | None] | None
+        task_path: Path,
+        prompts: list[str | None] | None,
+        skills_dir: str | Path | None = None,
+        skill_nudge: str = "",
+        agent: str | None = None,
     ) -> list[str]:
-        """Read instruction.md and resolve prompt list."""
+        """Read instruction.md and resolve prompt list.
+
+        skill_nudge: "", "name", "description", or "full".
+        """
         instruction_path = task_path / "instruction.md"
         if not instruction_path.exists():
             raise FileNotFoundError(f"Task missing instruction.md: {task_path}")
         instruction = instruction_path.read_text().strip()
+
+        if skill_nudge:
+            from benchflow.agents.registry import AGENTS
+
+            skill_display_path = "~/.claude/skills"
+            if agent:
+                agent_cfg = AGENTS.get(agent)
+                if agent_cfg and agent_cfg.skill_paths:
+                    skill_display_path = agent_cfg.skill_paths[0].replace("$HOME", "~")
+
+            skills = []
+            for src in [skills_dir, task_path / "environment" / "skills"]:
+                if src and Path(src).is_dir():
+                    for d in sorted(Path(src).iterdir()):
+                        if d.is_dir() and (d / "SKILL.md").exists():
+                            content = d.joinpath("SKILL.md").read_text()
+                            name = d.name
+                            desc = ""
+                            if content.startswith("---"):
+                                parts = content.split("---", 2)
+                                if len(parts) >= 3:
+                                    import yaml
+                                    try:
+                                        fm = yaml.safe_load(parts[1])
+                                        desc = fm.get("description", "") if fm else ""
+                                    except Exception:
+                                        pass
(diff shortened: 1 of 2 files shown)


xdotli:
@zjw49246 thanks for the PR! 
> BenchFlow uses each agent's native skill mechanism (progressive disclosure for claude-agent-acp), but agents often fail to discover and invoke skills. This creates a parity gap with Harbor, which injects skill content directly into the system prompt. See https://github.com/benchflow-ai/benchflow/issues/204 for full analysis.
as for this, I have a small question: so as for skill loading mechanism, i'm curious what's diff between:
1. original agent
2. benchflow before this fix
3. benchflow after this fix
4. and harbor. 
And my main focus is which one would be closest to the native agent (the one you would use in your terminal / editor).

zjw49246:
@xdotli The four modes differ in how skills reach the agent:

1. **Native agent** (e.g. Claude Code in terminal): skill descriptions sit in the tool list via progressive disclosure. Users can type `/skillname` to force-load, and often mention skills in their prompts naturally.
2. **BenchFlow before this PR**: same progressive disclosure, but nobody tells the agent about skills. It often ignores them entirely.
3. **BenchFlow after this PR**: prepends one line to the prompt telling the agent where skills are and to read them. Similar to a user who knows skills are available.
4. **Harbor**: OpenHands SDK injects skill content into the system prompt. The agent can't miss them, but no production agent works this way.

(2) is closest to native when users don't mention skills. (3) is closest to native when they do. (4) goes beyond native.

It's opt-in because the two modes answer different questions: without nudge you're testing whether agents can discover skills on their own, with nudge you're testing whether skills actually help once the agent knows about them.

Also pushed a fix for the `agent_env is None` bug that Devin caught.

xdotli (approved):
This looks great. Thanks for the clear breakdown of the four modes in your comment - it makes total sense. Implementing this opt-in nudge mechanism at the prompt resolution layer gives us exactly the tool we need to evaluate skill effectiveness vs. pure discovery without polluting the core environment logic. LGTM!

PR #212, opened 2026-04-28
fix: dynamic --rootdir in PYTEST_ADDOPTS based on task workspace

### LINKED ISSUE #211: Three task-author papercuts: skill injection breaks home dir, --rootdir=/app forced, plugin autoload disabled silently breaks pytest-json-ctrf
Hit three issues in a row while building skill-using tasks against `bench eval create -t ... -a claude-agent-acp -s skills/`. Each is recoverable but each silently produced a broken benchmark.

## 1. Skill injection makes `\$HOME/.claude/` un-writable for sandbox user

`_env_setup.py:_inject_skills_into_dockerfile` appends to the task's Dockerfile:

```
COPY _deps/skills /skills/
RUN mkdir -p /home/agent/.claude && ln -sf /skills /home/agent/.claude/skills
RUN mkdir -p /home/agent/.codex   && ln -sf /skills /home/agent/.codex/skills
...
```

These run at image build as root, so `/home/agent/.claude/` ends up root-owned. `setup_sandbox_user` then creates the `agent` user at runtime, but the home subdirs are still root-owned. Claude Code's ACP shim tries to create `/home/agent/.claude/session-env` on first Bash invocation and fails with EACCES.

The agent message in the trajectory is:

> The Bash tool can't initialize because it needs to create \`/home/agent/.claude/session-env\`, which isn't writable. I can't run Python or recalc formulas without Bash.

The whole eval then either burns its budget on diagnostic Bash retries or dies idle. Reward looks legit (no crash) but the agent never got to do real work.

**Workaround in the task Dockerfile:**
```dockerfile
RUN mkdir -p /home/agent/.claude /home/agent/.codex /home/agent/.opencode /home/agent/.agents && \\
    chmod -R 777 /home/agent
```

But task authors won't know to do this — the standard Dockerfile template in CONTRIBUTING.md doesn't mention it.

**Suggested fix:** in `_inject_skills_into_dockerfile`, append a `chown -R 1000:1000 /home/agent` (or whatever the configured sandbox UID is) after the symlink lines. Alternatively, do the symlink at runtime in `setup_sandbox_user` after the user exists.

## 2. \`PYTEST_DISABLE_PLUGIN_AUTOLOAD=1\` silently breaks \`pytest-json-ctrf\`

The verifier hardening in `_sandbox.py:VERIFIER_ENV` sets `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1`. This means `pytest-json-ctrf` (installed via `uvx --with` or pip in test.sh) doesn't load and `--ctrf /path` becomes "unrecognized argument". Pytest exits with code 4.

The hardening comment says:
> Tasks that need specific plugins declare them in task.toml [verifier] pytest_plugins = [...]

But:
- This isn't documented in `docs/task-authoring.md`
- The example `test.sh` in `docs/task-authoring.md` itself uses `pytest --ctrf /logs/verifier/ctrf.json /tests/test_outputs.py` — which is exactly what fails.
- `bench tasks check` doesn't flag the mismatch

**Suggested fixes** (one of):
- Auto-allowlist `pytest_json_ctrf` since it's literally in the standard template
- Document the `[verifier] pytest_plugins` opt-out in `docs/task-authoring.md` next to the test.sh example
- Have `bench tasks check` warn when test.sh uses `--ctrf` without the corresponding plugin declaration

## 3. \`--rootdir=/app\` forced even when task uses \`WORKDIR /root\`

Hardening in `_sandbox.py:VERIFIER_ENV`:
```python
"PYTEST_ADDOPTS": (
    "-c /dev/null "
    "--confcutdir=/tests "
    "--rootdir=/app "
    ...
),
```

This anchors pytest's rootdir to `/app`. But:
- `docs/task-authoring.md` says "the working directory is the Dockerfile's `WORKDIR` (typically `/app/`)"
- Most task examples I've seen in the wild use `WORKDIR /root` (probably copied from older harbor templates)

If the Dockerfile doesn't `mkdir /app`, pytest fails:
```
ERROR: Directory '/app' not found. Check your '--rootdir' option.
```

The error is hard to debug from the verifier output — looks like pytest itself misconfigured rather than a missing dir.

**Suggested fix:** pick one path. Either:
- Auto-detect `WORKDIR` from the Dockerfile and use that as `--rootdir`
- Have `_inject_skills_into_dockerfile` (or a sibling helper) ensure `/app` exists
- Have `bench tasks check` fail when Dockerfile doesn't create `/app`

## Cumulative effect

Each papercut alone is small. Together, a task author's first run with `-s` will (a) silently fail tests because of plugin autoload, (b) silently report a fake reward because their test.sh doesn't propagate \$? past a tee pipe, (c) realize the agent never ran Bash, and (d) only then realize they also need to handle /app. That's a lot of debugging before the first signal.

Happy to send PRs for any of these — flagging together since they're related (all hit at the agent-with-skills-and-pytest path).

zjw49246:
## Summary

- PR #194 hardcoded `--rootdir=/app` in `VERIFIER_ENV["PYTEST_ADDOPTS"]`, which breaks **67 SkillsBench tasks** that use `WORKDIR /root` — pytest fails with `ERROR: Directory '/app' not found`
- Extract `_build_pytest_addopts(workspace, plugin_flags)` that sets `--rootdir` to the task's actual workspace (from Dockerfile `WORKDIR`), falling back to `/app` when workspace is `None`
- All security invariants preserved: `-c /dev/null`, `--confcutdir=/tests`, ADDOPTS re-pin after task env merge, `shlex.quote` on workspace path

## Changes

| File | What |
|------|------|
| `src/benchflow/_sandbox.py` | Remove static `--rootdir=/app` from `VERIFIER_ENV`; add `_build_pytest_addopts()` helper; simplify ADDOPTS assembly in `harden_before_verify()` |
| `tests/test_sandbox_hardening.py` | Update 6 existing assertions; add 13 new tests (workspace tracking, fallback, edge cases, security) |

## Test plan

- [x] 79 sandbox hardening tests pass (13 new)
- [x] 707 full test suite pass, 0 regressions
- [x] Oracle eval on `WORKDIR /root` tasks: weighted-gdp-calc (27/27), offer-letter-generator (18/18), earthquake-plate-calculation (8/8), parallel-tfidf-search (5/5) — all reward 1.0, `rootdir: /root`
- [x] Oracle eval on `WORKDIR /app` task: dialogue-parser — `rootdir: /app` confirmed

Closes #211 (point 3).

---

<a href="https://app.devin.ai/review/benchflow-ai/benchflow/pull/212" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/benchflow/_acp_run.py
+++ b/src/benchflow/_acp_run.py
@@ -115,6 +115,8 @@ async def connect_acp(
         logger.info(f"Agent sandboxed as: {sandbox_user}")
 
     acp_client: ACPClient | None = None
+    session: object | None = None
+    agent_name = agent
     for attempt in range(_ACP_CONNECT_MAX_RETRIES + 1):
         if attempt > 0:
             delay = _ACP_CONNECT_BASE_DELAY * (2 ** (attempt - 1))
@@ -174,6 +176,9 @@ async def connect_acp(
                 with contextlib.suppress(Exception):
                     await acp_client.close()
             raise
+
+    if acp_client is None or session is None:
+        raise RuntimeError("ACP connection did not initialize")
 
     agent_cfg = AGENTS.get(agent)
     if model and (agent_cfg is None or agent_cfg.supports_acp_set_model):
--- a/src/benchflow/_agent_env.py
+++ b/src/benchflow/_agent_env.py
@@ -235,7 +235,7 @@ def resolve_agent_env(
                 )
             )
         )
-        if has_agent_native_bridge_key:
+        if has_agent_native_bridge_key and mapped_provider_key is not None:
             # Only pre-existing same-provider aliases or explicit generic bridge
             # keys can satisfy provider auth. Values synthesized by env_mapping
             # or inherited from another provider context must not bypass the
--- a/src/benchflow/_daytona_patches.py
+++ b/src/benchflow/_daytona_patches.py
@@ -6,7 +6,7 @@ SDK is only touched when a Daytona environment is actually being built.
 
 import asyncio
 import logging
-from typing import Any
+from typing import Any, cast
 
 logger = logging.getLogger(__name__)
 
@@ -97,6 +97,7 @@ def apply() -> None:
         )
         return SessionCommandLogsResponse(output="", stdout="", stderr="")
 
-    AsyncProcess.get_session_command_logs = _patched_get_session_command_logs  # type: ignore[method-assign]
+    async_process_cls = cast(Any, AsyncProcess)
+    async_process_cls.get_session_command_logs = _patched_get_session_command_logs
     _PATCHED = True
     logger.debug("daytona SDK patches applied")
--- a/src/benchflow/_sandbox.py
+++ b/src/benchflow/_sandbox.py
@@ -351,9 +351,7 @@ _RUNTIME_PATH_PREFIXES = ("/tmp", "/var/tmp", "/logs", "/testbed")
 _DEFAULT_ROOTDIR = "/app"
 
 
-def _build_pytest_addopts(
-    workspace: str | None = None, plugin_flags: str = ""
-) -> str:
+def _build_pytest_addopts(workspace: str | None = None, plugin_flags: str = "") -> str:
     """Build PYTEST_ADDOPTS with a dynamic --rootdir based on the workspace.
 
     Without an explicit --rootdir, -c /dev/null causes pytest to fall back to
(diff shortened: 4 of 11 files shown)


zjw49246:
**CI status:** the `test` workflow fails on two steps — both are **pre-existing failures on main**, not introduced by this PR.

1. **`ruff format`** — `sdk.py` and `trial.py` have formatting issues on main. This PR only touches `_sandbox.py` and `test_sandbox_hardening.py`, both of which pass `ruff format --check`.

2. **`ty check`** — 24 type errors across 10 files (`_acp_run.py`, `_agent_setup.py`, `_daytona_patches.py`, `_env_setup.py`, `mcp/reviewer_server.py`, `runtime.py`, `sdk.py`, `skill_eval.py`, `trial.py`, `trial_yaml.py`). None in `_sandbox.py` or `test_sandbox_hardening.py`.

Verified by checking out `main` and running the same checks — identical failures.

Yiminnn:
Closing because this fix has been superseded by merged PR #215, which rolled up the related BenchFlow/OpenHands benchmark stability fixes. See merge commit 1427f62dc073ab475f3b427b581b04e96395b111.

PR #217, opened 2026-04-30
fix: preserve verifier log mount

Yiminnn:
## Summary
- Preserve the /logs/verifier directory during verifier hardening so Docker bind mounts stay attached
- Clear only verifier log contents before verification
- Reuse the same cleanup command in soft verification

## Test Plan
- uv run ruff format --check src/benchflow/_sandbox.py src/benchflow/trial.py tests/test_sandbox_hardening.py
- uv run ruff check src/benchflow/_sandbox.py src/benchflow/trial.py tests/test_sandbox_hardening.py
- uv run pytest tests/test_sandbox_hardening.py tests/test_sandbox_verifier_workspace.py tests/test_sdk_lockdown.py

### DIFF
--- a/src/benchflow/_sandbox.py
+++ b/src/benchflow/_sandbox.py
@@ -642,11 +642,16 @@ async def _trusted_verifier_pythonpath(
     return ":".join(e for e in extras if isinstance(e, str))
 
 
-# Wipe and recreate /logs/verifier/ before the verifier runs.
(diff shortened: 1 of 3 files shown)


Yiminnn:
Closing this draft after local reproduction showed the missing reward was caused by using a /tmp jobs directory that is not bind-mounted back to the host on this Docker setup. SkillsBench validation now uses a Docker-shared /Users jobs path and BenchFlow main passes the sanity verifier.

PR #230, opened 2026-05-02
fix: pass effective task path to deploy_skills so already-injected ch…

### LINKED ISSUE #229: deploy_skills double-deploys when skills_dir is set, causing 'cannot overwrite directory "/skills/..."' on container cp
## Summary

When `skills_dir` is passed (either via CLI `--skills-dir`, `skills_dir: auto`, or a path in YAML), benchflow deploys skills **twice**: once baked into the image via `_inject_skills_into_dockerfile`, and again at runtime via `deploy_skills` → `env.upload_dir(..., "/skills")`. The runtime upload fails when the source contains a symlink whose name collides with an entry already baked at `/skills/...`, with errors like:

```
docker compose ... cp <task>/environment/skills/. main:/skills
Error response from daemon: cannot overwrite directory "/skills/book-nav" with non-directory "/skills"
```

(For sources that are plain directories with the same name as a baked entry, the second copy may silently overlay the baked files — still a redundant, surprising step.)

Reproduced on `benchflow==0.3.3.dev0` running an InfiniteBench experiment with `skills_dir: auto` where each task's `environment/skills/` contains a per-task symlink (`book-nav -> ../../../data/...`). 16/16 trials failed with the error above; all 16 succeed after the patch below.

## Root cause

`Trial._setup` (`benchflow/trial.py:387–403`) creates a temp copy `effective_task_path` of the task and calls `_inject_skills_into_dockerfile(effective_task_path, ...)`, which:
- writes `<temp>/environment/_deps/skills/` (real copy via `shutil.copytree`, follows symlinks)
- appends `COPY _deps/skills /skills/` to the **temp** Dockerfile

The image is then built from the temp path, so `/skills/<entry>` is baked in.

Later, `Trial.install_agent` (`trial.py:512–520`) calls:

```python
await deploy_skills(
    self._env,
    cfg.task_path,            # ← original, NOT effective
    cfg.skills_dir,
    ...
)
```

`deploy_skills` (`benchflow/_agent_setup.py:217–229`) detects whether injection already happened by reading the Dockerfile at the path it was given:

```python
if skills_dir:
    dockerfile = task_path / "environment" / "Dockerfile"
    already_injected = (
        dockerfile.exists()
        and "COPY _deps/skills /skills/" in dockerfile.read_text()
    )
    if not already_injected:
        ...
        await env.upload_dir(skills_path, "/skills")
```

Because it receives `cfg.task_path` (original, never modified) instead of `effective_task_path` (the temp copy with the injected `COPY` line), `already_injected` is **always `False`**. The runtime cp then fires on top of the already-baked `/skills/...` and either errors (symlink-vs-dir mismatch) or redundantly overwrites.

## Reproduction

1. Task with a symlink under `environment/skills/`:
   ```
   datasets/<bench>/<task>/environment/skills/<skill> -> ../../../<somewhere>/<skill>
   ```
2. Run any experiment with `skills_dir: auto` (or any non-null `skills_dir`).
3. Observe `docker compose ... cp` failure during `Trial.install_agent`:
   ```
   cannot overwrite directory "/skills/<skill>" with non-directory "/skills"
   ```

Confirmed via `tail /tmp/benchflow-task-*/.../environment/Dockerfile` showing the injected `COPY _deps/skills /skills/` block in the temp copy, while the original task `Dockerfile` is unchanged — making the `already_injected` check structurally unable to detect injection.

## Suggested fix

Pass the effective (post-injection) task path to `deploy_skills` so its `already_injected` check reads the same Dockerfile that was actually built into the image.

Minimal patch (verified — 16/16 trials now pass):

```diff
--- a/benchflow/trial.py
+++ b/benchflow/trial.py
@@ -402,6 +402,8 @@ class Trial:
         if cfg.skills_dir:
             _inject_skills_into_dockerfile(effective_task_path, Path(cfg.skills_dir))

+        self._effective_task_path = effective_task_path
+
         self._env = _create_environment(
             cfg.environment,
             self._task,
@@ -511,7 +513,7 @@ class Trial:

         await deploy_skills(
             self._env,
-            cfg.task_path,
+            getattr(self, "_effective_task_path", cfg.task_path),
             cfg.skills_dir,
             self._agent_cfg,
             cfg.sandbox_user,
```

Alternatively, thread an explicit `already_injected: bool` flag through `Trial._setup → deploy_skills` so the decision doesn't depend on re-parsing the Dockerfile string at all.

## Environment

- benchflow `0.3.3.dev0`
- harbor `docker` backend
- Python 3.12, Linux

Happy to send a PR if a maintainer prefers either fix shape.

EYH0602:
## Summary
- Fixes #229  — `deploy_skills` was receiving `cfg.task_path` (original) instead of the post-injection temp copy, so its `already_injected` check always read False and the runtime `env.upload_dir(..., "/skills")` fired on top of skills already baked into the image. With a colliding symlink in the source, the cp failed with `cannot overwrite directory "/skills/<entry>" with non-directory "/skills"`.
- Stash the post-injection path on `self._effective_task_path` in `Trial._setup` and pass it to both `deploy_skills` call sites (oracle + agent).
- Add a regression test in `tests/test_agent_setup.py` that writes a Dockerfile carrying the injected `COPY _deps/skills /skills/` line and asserts `env.upload_dir` is not called.

## Test plan
- [x] `pytest tests/test_agent_setup.py tests/test_trial_install_agent_timeout.py tests/test_env_setup.py tests/test_verify.py` — passes (81 tests).
- [x] `ruff check` clean.
- [x] `ty check src/` clean.
- [x] Re-run reporter's InfiniteBench run with `skills_dir: auto` to confirm 16/16 trials succeed end-to-end.

### DIFF
--- a/src/benchflow/_agent_setup.py
+++ b/src/benchflow/_agent_setup.py
@@ -240,6 +240,7 @@ async def deploy_skills(
                 logger.warning(f"Skills dir not found: {skills_path}")
         else:
             logger.info("Skills already injected via Dockerfile")
+            effective_skills = "/skills"
 
     # Distribute to agent-specific discovery paths
     if agent_cfg is not None:
--- a/src/benchflow/trial.py
+++ b/src/benchflow/trial.py
@@ -402,6 +402,8 @@ class Trial:
         if cfg.skills_dir:
             _inject_skills_into_dockerfile(effective_task_path, Path(cfg.skills_dir))
 
+        self._effective_task_path = effective_task_path
+
         self._env = _create_environment(
             cfg.environment,
             self._task,
@@ -474,7 +476,7 @@ class Trial:
             )
             await deploy_skills(
                 self._env,
-                cfg.task_path,
+                getattr(self, "_effective_task_path", cfg.task_path),
                 cfg.skills_dir,
                 None,
                 cfg.sandbox_user,
@@ -520,7 +522,7 @@ class Trial:
 
         await deploy_skills(
             self._env,
-            cfg.task_path,
+            getattr(self, "_effective_task_path", cfg.task_path),
             cfg.skills_dir,
             self._agent_cfg,
             cfg.sandbox_user,
--- a/tests/test_agent_setup.py
+++ b/tests/test_agent_setup.py
@@ -92,6 +92,51 @@ async def test_deploy_skills_uploads_runtime_skills_and_links_shared_tree(tmp_pa
 
 
 @pytest.mark.asyncio
+async def test_deploy_skills_skips_runtime_upload_when_dockerfile_already_injected(
+    tmp_path,
+):
+    """Guards the fix from PR for issue #11 against the regression where
+    skills got deployed twice — once baked into the image via
+    `_inject_skills_into_dockerfile`, and again at runtime via
+    `env.upload_dir(..., "/skills")`. The runtime cp failed with
+    `cannot overwrite directory "/skills/<entry>" with non-directory "/skills"`
+    when the source contained a symlink colliding with a baked entry.
+
+    The detection works only when `deploy_skills` receives the *effective*
+    task path — the temp copy whose Dockerfile actually carries the
+    injected `COPY _deps/skills /skills/` line.
+    """
+    env = MagicMock()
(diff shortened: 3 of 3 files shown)


EYH0602:
landed in 0.3.4

Issue #387, opened 2026-05-24
ManifestEnvironment snapshot/restore ignores failed sandbox commands and records bogus checkpoints


xdotli:
## Summary

`ManifestEnvironment.snapshot()` and `restore()` run the SQLite backup / copy commands but never inspect the returned `ExecResult`. If `sqlite3 .backup` or `cp` fails, `snapshot()` still returns a `StateSnapshot` and `restore()` still returns success.

That breaks the Branch/value-function path: a branch can record a checkpoint that never existed, restore from a failed copy, then score children from corrupted or unchanged state.

## Reproduction

```bash
uv run python - <<'PY'
import asyncio
from benchflow.environment.manifest import EnvironmentManifest
from benchflow.environment.manifest_env import ManifestEnvironment
from benchflow.environment.protocol import StateSnapshot
from benchflow.sandbox.protocol import ExecResult

manifest = EnvironmentManifest.model_validate_toml('''
[environment]
name = "stateful"
base_image = "x:latest"
owns_lifecycle = false
[[environment.services]]
name = "svc"
command = "svc --port 9001"
port = 9001
[environment.state]
kind = "sqlite"
paths = ["/data/missing.db"]
''')

class FailingSandbox:
    def __init__(self): self.calls=[]
    async def exec(self, cmd, *, user='root', timeout_sec=30):
        self.calls.append(cmd)
        return ExecResult(return_code=1, stdout='', stderr='boom')

async def main():
    sb = FailingSandbox()
    env = ManifestEnvironment(manifest, sandbox=sb)
    snap = await env.snapshot()
    print('snapshot_returned', snap)
    await env.restore(StateSnapshot(id='bad', path='/tmp/nope'))
    print('restore_returned_without_error', True)
    print('calls', sb.calls)

asyncio.run(main())
PY
```

Observed:

```text
snapshot_returned StateSnapshot(id='...', path='/tmp/benchflow-snapshots/...')
restore_returned_without_error True
calls ['mkdir -p ... && sqlite3 /data/missing.db ".backup ..."', 'cp /tmp/nope/missing.db /data/missing.db']
```

Expected: `snapshot()` and `restore()` should raise with stderr/stdout and command context when the sandbox command returns non-zero.

## Code evidence

- `src/benchflow/environment/manifest_env.py`: `snapshot()` awaits `self._sandbox.exec(...)` and returns `StateSnapshot` without checking `return_code`.
[...]

PR #486, opened 2026-05-25
fix: lazy-import optional deps + manifest env snapshot integrity (#358 #387)

### LINKED ISSUE #358: Base install fails to import benchflow because optional Daytona deps are imported at package import time
## Summary

A base/core install of `benchflow` cannot import the package or run `bench --help` unless optional Daytona dependencies are present. The import chain reaches `benchflow.sandbox.daytona`, which imports `tenacity`, but `tenacity` only lives in the `sandbox-daytona`/`sandbox-modal` extras.

This breaks the README install path (`uv tool install benchflow`) and any Python user who only installs the core package.

## Repro

From this checkout:

```bash
uv run --no-project --with . python -c "import benchflow; print('import ok', benchflow.__version__)"
```

Observed:

```text
ModuleNotFoundError: No module named 'tenacity'
```

I also saw the same chain initially from the local project env when running:

```bash
uv run benchflow --help
uv run bench agent list
uv run python -c "import benchflow"
```

## Expected

Core install/import and `bench --help` should work without `sandbox-daytona`, `sandbox-modal`, or `judge` extras. Optional sandbox implementations should fail only when the user selects that sandbox.

## Actual

Import chain:

- `src/benchflow/__init__.py` imports `rollout_config_from_yaml`
- `src/benchflow/_utils/yaml_loader.py` imports `RolloutConfig`
- `src/benchflow/rollout.py` imports `SandboxStartupError` from `benchflow.sandbox.daytona`
- `src/benchflow/sandbox/daytona.py` imports `tenacity`
- `pyproject.toml` does not include `tenacity` in core deps

## Impact

P0 packaging/release blocker: the documented install path can produce an installed package whose import surface is unusable.

## Suggested fix

- Move `SandboxStartupError` out of `sandbox/daytona.py` into a core module that has no optional deps, or import it lazily only around Daytona startup handling.
- Add an isolated packaging smoke test that installs core only and runs:
  - `python -c "import benchflow"`
  - `bench --help`
  - `bench agent list`

xdotli:
Fixes #358, #387.

## #358 — Base install fails to import benchflow

A bare `pip install benchflow` (no `sandbox-daytona` / `sandbox-modal` extras) used to fail at `import benchflow` because the import chain reached `benchflow.sandbox.daytona`, which imported `tenacity` and the `daytona` SDK at module top.

**Fix:**
- Moved `SandboxStartupError` into `benchflow.sandbox.protocol` (no optional deps). `rollout.py` imports it from there; `daytona.py` re-exports for backward compatibility, so `from benchflow.sandbox.daytona import SandboxStartupError` still works.
- Replaced module-top `tenacity` and Daytona-SDK imports with lazy loaders (`_load_daytona_sdk`, `_tenacity_retry`) that materialize the optional packages on first `DaytonaSandbox` instantiation / first decorated-method call. Clear `ImportError` with install hint if the extra is missing when the SDK is actually needed.
- Added `tests/test_base_install_imports.py` guarding the structural contract so a future move can't silently regress #358 (one test even hides `tenacity` from `sys.modules` and re-imports the daytona module to assert it loads).

**Verified:**
```
$ uv venv /tmp/bf_smoke --python 3.12
$ uv pip install --python /tmp/bf_smoke/bin/python .
$ /tmp/bf_smoke/bin/python -c 'import benchflow; print(benchflow.__version__)'  # ok
$ /tmp/bf_smoke/bin/bench --help                                                  # ok
$ /tmp/bf_smoke/bin/bench agent list                                              # ok
```

## #387 — ManifestEnvironment snapshot/restore ignored failed sandbox commands

`ManifestEnvironment.snapshot()` (via `_capture_baseline`) and `restore()` ran the SQLite backup / `cp` commands but never inspected `ExecResult.return_code`. A failed `sqlite3 .backup` returned a bogus `StateSnapshot`; a failed `cp` returned silently. Branch then scored children against state that never existed (or was never restored), letting corrupted child rewards / estimated `V(parent)` leak into release evidence or training data.

**Fix:**
- New `EnvironmentSnapshotError` (typed `RuntimeError`) carrying `command`, `exit_code`, `stdout`, `stderr`.
- Both `_capture_baseline` and `restore` now check `return_code` and raise `EnvironmentSnapshotError` with a truncated stderr snippet in the message.
- This propagates out of `branch()` (`_checkpoint_branch` / `_restore_branch` don't catch), so Branch fails closed instead of scoring children against invalid state.

**Regression tests:** snapshot failure, restore failure, and the baseline-capture path inside `provision()` for stateful manifests.

## Quality gate

- `uv run python -m pytest tests/` — 2346 passed, 2 skipped, 1 deselected
- `uv run ruff check .` — All checks passed
- `uv run ty check src/` — All checks passed

---

> [!NOTE]
> **Medium Risk**
> Changes the core import graph and branch rollback substrate (`snapshot`/`restore`); behavior is safer (fail-closed) but any missed error path could still break branching or base-install consumers.
> 
> **Overview**
> This PR fixes two install/runtime gaps: **base installs** can import core packages without optional Daytona deps, and **manifest environment** snapshot/restore no longer pretends success when in-sandbox commands fail.
> 
> For **#358**, `SandboxStartupError` moves to `benchflow.sandbox.protocol` so `rollout` does not pull in `daytona`/`tenacity` at import time. `daytona.py` lazy-loads the SDK via `_load_daytona_sdk()` on first `DaytonaSandbox()` construction, stubs `tenacity` when the extra is missing, and keeps a backward-compatible re-export. New smoke tests lock the import contract.
> 
> For **#387**, `ManifestEnvironment` adds `EnvironmentSnapshotError` and checks `ExecResult.return_code` after SQLite `.backup` and `cp` restore (including baseline capture during `provision`). Failed checkpoints now fail closed instead of letting `Rollout.branch()` score children on missing or stale state. Regression tests cover snapshot, restore, and provision baseline paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1cb88450700e5092466bfd389c071a440c6c2eab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/benchflow/environment/manifest_env.py
+++ b/src/benchflow/environment/manifest_env.py
@@ -42,6 +42,26 @@ from benchflow.environment.protocol import (
 )
 
 
+class EnvironmentSnapshotError(RuntimeError):
+    """Raised when a snapshot or restore sandbox command fails (issue #387).
+
+    Snapshot/restore are the substrate ``Rollout.branch()`` builds rollback on
+    — a silent infra failure here lets a Branch record a checkpoint that never
+    existed (or restore from a copy that did not happen), then score children
+    from corrupted state. Bubbling a typed error keeps that failure visible to
+    the caller so child rewards / estimated V(parent) never leak into release
+    evidence as if they were real.
+    """
+
+    def __init__(self, message: str, *, command: str, exit_code: int,
+                 stdout: str, stderr: str) -> None:
+        super().__init__(message)
+        self.command = command
+        self.exit_code = exit_code
+        self.stdout = stdout
+        self.stderr = stderr
+
+
 class ManifestEnvironment:
     """Runs a manifest-declared stateful environment over a Sandbox.
 
@@ -175,6 +195,11 @@ class ManifestEnvironment:
         """Copy the declared state files into a fresh in-sandbox snapshot dir.
 
         Caller guarantees ``self._manifest.state`` is not None.
+
+        Issue #387: if the sandbox command fails (sqlite3 missing, db file
+        absent, copy permission denied), raise ``EnvironmentSnapshotError``
+        rather than recording a checkpoint that never existed — Branch and
+        ``reset()`` depend on this being a real backup.
         """
         spec = self._manifest.state
         assert spec is not None  # caller checks
@@ -184,7 +209,18 @@ class ManifestEnvironment:
         for src in spec.paths:
             dest = f"{snap_dir}/{PurePosixPath(src).name}"
             cmds.append(f'sqlite3 {shlex.quote(src)} ".backup {shlex.quote(dest)}"')
-        await self._sandbox.exec(" && ".join(cmds), timeout_sec=120)
+        command = " && ".join(cmds)
+        result = await self._sandbox.exec(command, timeout_sec=120)
+        if result.return_code != 0:
+            raise EnvironmentSnapshotError(
+                f"snapshot failed for environment '{self._manifest.name}' "
+                f"(exit_code={result.return_code}): "
+                f"{(result.stderr or result.stdout or '').strip()[:500]}",
+                command=command,
+                exit_code=result.return_code,
+                stdout=result.stdout or "",
+                stderr=result.stderr or "",
+            )
         return StateSnapshot(id=snap_id, path=snap_dir)
 
     async def restore(self, snap: StateSnapshot) -> None:
@@ -193,6 +229,12 @@ class ManifestEnvironment:
         Copies each captured DB file from the snapshot directory back over
         its live path. The caller quiesces the agent and services first
         (the Branch lifecycle) so the copy is consistent.
+
+        Issue #387: if the sandbox ``cp`` command fails (snapshot directory
+        missing, destination not writable), raise
+        ``EnvironmentSnapshotError`` rather than silently returning success
+        — the live state is then unchanged and the caller must treat it as
+        an infra failure, not a successful rollback.
         """
         spec = self._manifest.state
         if spec is None:
@@ -205,7 +247,18 @@ class ManifestEnvironment:
         for dst in spec.paths:
             src = f"{snap.path}/{PurePosixPath(dst).name}"
             cmds.append(f"cp {shlex.quote(src)} {shlex.quote(dst)}")
-        await self._sandbox.exec(" && ".join(cmds), timeout_sec=120)
+        command = " && ".join(cmds)
+        result = await self._sandbox.exec(command, timeout_sec=120)
+        if result.return_code != 0:
+            raise EnvironmentSnapshotError(
+                f"restore failed for environment '{self._manifest.name}' "
+                f"snapshot {snap.id!r} (exit_code={result.return_code}): "
+                f"{(result.stderr or result.stdout or '').strip()[:500]}",
+                command=command,
+                exit_code=result.return_code,
+                stdout=result.stdout or "",
+                stderr=result.stderr or "",
+            )
 
     async def reset(self) -> None:
         """Return the environment to its per-task initial state.
--- a/src/benchflow/rollout.py
+++ b/src/benchflow/rollout.py
@@ -78,7 +78,6 @@ from benchflow.providers.runtime import (
 from benchflow.rewards.validation import validate_reward_map
 from benchflow.rollout_branch import ChildRunner
 from benchflow.rollout_branch import branch as _branch_engine
-from benchflow.sandbox.daytona import SandboxStartupError
 from benchflow.sandbox.lockdown import (
     _resolve_locked_paths,
     _seed_verifier_workspace,
@@ -86,6 +85,7 @@ from benchflow.sandbox.lockdown import (
     lockdown_paths,
     setup_sandbox_user,
 )
+from benchflow.sandbox.protocol import SandboxStartupError
 from benchflow.sandbox.setup import (
     _create_environment,
     _inject_skills_into_dockerfile,
--- a/src/benchflow/sandbox/daytona.py
+++ b/src/benchflow/sandbox/daytona.py
@@ -16,11 +16,37 @@ import os
 import re
 import shlex
 from abc import abstractmethod
-from collections.abc import Callable
-from functools import wraps
 from pathlib import Path
-from typing import Any, TypeVar
+from typing import Any
 from uuid import uuid4
+
+try:
+    from tenacity import (
+        retry,
+        retry_if_exception_type,
+        stop_after_attempt,
+        wait_exponential,
+    )
+except ImportError:  # base install without ``sandbox-daytona`` extras (#358)
+    # ``tenacity`` ships under the ``sandbox-daytona`` extra. The fallbacks below
+    # let this module import cleanly in a base install — ``DaytonaSandbox()`` is
+    # what requires the real SDK (and tenacity along with it).
+
+    def retry(*_args: Any, **_kwargs: Any) -> Any:
+        def _decorator(fn: Any) -> Any:
+            return fn
+
+        return _decorator
+
+    def stop_after_attempt(*_args: Any, **_kwargs: Any) -> Any:
+        return None
+
+    def wait_exponential(*_args: Any, **_kwargs: Any) -> Any:
+        return None
+
+    def retry_if_exception_type(*_args: Any, **_kwargs: Any) -> Any:
+        return None
+
 
 from benchflow._paths import iter_safe_tree
 from benchflow.sandbox._base import (
@@ -76,9 +102,9 @@ def _ensure_daytona_anyio_compat() -> None:
     anyio.AsyncContextManagerMixin = _AsyncContextManagerMixin  # type: ignore[attr-defined]
 
 
-# Module-level handles for the Daytona SDK + ``tenacity``. These optional deps
-# are pulled in lazily — see ``_load_daytona_sdk`` — so importing this module
-# in a base install does not require the ``sandbox-daytona`` extra (issue #358).
+# Module-level handles for the Daytona SDK. The SDK is shipped under the
+# ``sandbox-daytona`` extra and is pulled in lazily — see ``_load_daytona_sdk``
+# — so importing this module in a base install does not require it (#358).
 _DAYTONA_SDK_LOADED = False
 AsyncDaytona: Any = None
 AsyncSandbox: Any = None
@@ -91,8 +117,6 @@ Image: Any = None
 Resources: Any = None
 SessionExecuteRequest: Any = None
 SnapshotState: Any = None
-
-_F = TypeVar("_F", bound=Callable[..., Any])
 
 
 def _load_daytona_sdk() -> None:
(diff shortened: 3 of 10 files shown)


xdotli:
**Verdict: REQUEST_CHANGES**

The two underlying fixes are correct (lift `SandboxStartupError` into `protocol`, surface snapshot/restore exec failures). But the lazy-tenacity machinery introduced to defer the optional `tenacity` import is a magical wrapper that re-implements parts of tenacity's API, leaks a test-only attribute into prod, and adds a defensive `_load_daytona_sdk()` call inside an exec method to patch over a test-time `__new__` pattern. There is a much simpler reframing that deletes the entire `_tenacity_retry` helper.

### 1. Code-judo / structural: `_tenacity_retry` is overbuilt — just guard the import

`src/benchflow/sandbox/daytona.py:242-294` introduces a custom retry-decorator wrapper that:
- Re-invents a mini DSL (`stop_after_attempt=3`, `wait_exponential={"multiplier": 2, ...}`) so callers don't import `tenacity`.
- Materializes the real `tenacity.retry` on first call and caches it in a `dict["fn"]`.
- Exposes a `retry_config` attribute purely so a test can verify the retry contract without invoking the wrapper.

This is a lot of structure for "import tenacity only if it's installed." The reframing:

```python
try:
    from tenacity import retry, stop_after_attempt, wait_exponential
except ImportError:  # base install — Daytona unused at runtime
    def retry(**_kw):
        def deco(fn): return fn
        return deco
    def stop_after_attempt(*_a, **_kw): return None
    def wait_exponential(*_a, **_kw): return None
```

Decorators stay at the call sites in their original `@retry(stop=stop_after_attempt(3), ...)` form. Module imports cleanly without tenacity. A base install that never instantiates `DaytonaSandbox` never notices. A base install that *does* try to construct one fails at `_load_daytona_sdk()` (the existing, correct gate) with a clear error message — the no-op retry is never exercised in practice.

This deletes:
- The entire `_tenacity_retry` function (~50 lines).
- The `retry_config = retry_kwargs` test-observability hack.
- The "tests construct via `__new__` so we call `_load_daytona_sdk` again inside `_sandbox_exec`" comment + extra call at line 1354.
- The need for the updated `test_create_sandbox_retry_count_is_three` test to grovel into a custom attribute.

The retry-count test goes back to `retry_obj = DaytonaSandbox._create_sandbox.retry; assert retry_obj.stop.max_attempt_number == 3` — the *real* tenacity contract — exactly as the pre-fix test had it.

**Strongly prefer this reframing.** It deletes one magical mechanism and restores a direct, boring contract. Please collapse before merge.

### 2. Spaghetti: test-only branch inside `_sandbox_exec`

`src/benchflow/sandbox/daytona.py:1350-1355`:

```python
# Tests construct DaytonaSandbox via __new__ to exercise this
# method in isolation — that skips __init__'s _load_daytona_sdk()
# call. Ensure the SDK handles ... are materialized before we touch them
_load_daytona_sdk()
```

Prod code carrying test scaffolding is a smell. The test in `tests/test_daytona_command_polling.py:32` does `DaytonaSandbox.__new__(DaytonaSandbox)` to skip `__init__`. The right fix is for the test to call `_load_daytona_sdk()` itself, not for every `_sandbox_exec` invocation in production to defensively re-check that the SDK loaded.

If you adopt the reframing in (1) (top-level `try/except ImportError` with a no-op `retry`), then `_load_daytona_sdk()` is the only thing left to lazy-load, and you can move *that* call into `_validate_definition` or `start()` — both of which always run before `_sandbox_exec` in real flows. The test-pattern branch then disappears.

### 3. Duplication: snapshot vs restore error blocks

`src/benchflow/environment/manifest_env.py:209-222` and `:247-260` are near-identical (12 lines each, same error construction, only the message verb differs). Extract a `_run_state_cmd(self, cmds: list[str], *, operation: str, snap_id: str | None = None)` helper that runs the command and raises `EnvironmentSnapshotError` with a templated message. Two callers, ~24 lines collapse to ~12.

Minor (not a blocker), but the cleaner shape is right there.

### Things I checked and they're fine

- **`EnvironmentSnapshotError`** as a typed exception is the right call — it makes the "silent infra failure -> corrupted Branch checkpoint" failure mode loud and lets callers `except` on the specific class.
- **`SandboxStartupError` move into `protocol`** + re-export from `daytona` for compat — clean.
- **Test file `test_base_install_imports.py`** is a smart structural guard. The `test_daytona_module_imports_without_tenacity` import-blocking pattern is the right level of evidence.

### File-size

`sandbox/daytona.py` is already 1489 lines (pre-existing). This PR adds ~100 net — most of it the `_tenacity_retry` machinery that should be deleted per (1). Net should be roughly flat after the reframing.

### Summary

Bug fixes are sound. The lazy-import scaffolding around tenacity is over-engineered and leaks observability into prod for tests. Two structural simplifications (top-level `try/except ImportError` for tenacity; extract the snapshot/restore error helper) delete code rather than add it. Please apply (1) before merge; (2) and (3) are nice-to-have.

xdotli:
## Independent verification — verdict: PASS

Verified against base `5934433` (pre-merge tip of v0.5-integration).

### Issue #358 — base install can't import benchflow
- Pre-fix repro: fresh venv, `uv pip install .` (no extras), `python -c "import benchflow"` → `ModuleNotFoundError: No module named 'tenacity'` originating from `benchflow/sandbox/daytona.py:22` (exact path from issue)
- Post-fix repro: fresh venv, `uv pip install .` (no extras):
  - `python -c "import benchflow; print(benchflow.__version__)"` → `0.5.0.dev0`
  - `bench --help` → renders normally
  - `bench agent list` → renders registered agents
- Author's regression test exercises bug: **YES** — `tests/test_base_install_imports.py` has 4 tests, all FAIL on base, PASS on PR; one test (`test_daytona_module_imports_without_tenacity`) actively hides `tenacity` from `sys.modules` and re-imports daytona to assert it loads — strong mutation-killing coverage

### Issue #387 — ManifestEnvironment snapshot/restore ignored failures
- Pre-fix repro (verbatim from issue with `FailingSandbox` returning `return_code=1`):
  - `snapshot returned (BUG): StateSnapshot(id='60524efb9696', ...)`
  - `restore returned (BUG)` (silent success)
- Post-fix repro:
  - `snapshot raised: EnvironmentSnapshotError snapshot failed for environment 'stateful' (exit_code=1): boom`
  - `restore raised: EnvironmentSnapshotError restore failed for environment 'stateful' snapshot 'bad' (exit_code=1): boom`
- Author's regression test exercises bug: **YES** — `test_snapshot_raises_when_sandbox_command_fails`, `test_restore_raises_when_sandbox_command_fails`, `test_provision_baseline_capture_failure_surfaces` all use a failing sandbox stub

### Test/lint
- pytest: **2346 passed, 2 skipped, 1 deselected**
- ruff: clean
- ty: clean

No merge / no push — verification only.

xdotli:
Applied thermo-nuclear audit remedies in 1cb8845:

1. **Replaced `_tenacity_retry` with standard top-level fallback.** `@retry(...)` from tenacity is now used directly. A top-level `try/except ImportError` provides no-op fallbacks so `benchflow.sandbox.daytona` still imports cleanly in a base install (the #358 guarantee — verified by `test_daytona_module_imports_without_tenacity`). Deleted the 50-line wrapper, the `_F` TypeVar, and `functools.wraps` / `collections.abc.Callable` imports.

2. **Removed `retry_config` test-scaffolding attribute.** `test_create_sandbox_retry_count_is_three` now inspects the live tenacity decorator config via `DaytonaSandbox._create_sandbox.retry.stop.max_attempt_number` — tenacity's documented introspection hook. Test adapts to prod API.

3. **Removed defensive `_load_daytona_sdk()` from `_sandbox_exec`.** It only existed to support tests that construct `DaytonaSandbox` via `__new__`. The three such tests (`test_sandbox_exec_passes_no_env_kv_argv`, `test_exec_times_out_when_daytona_command_never_exits`, `test_daytona_sdk_upload_dir_skips_symlinked_files`) now call `_load_daytona_sdk()` themselves — the same lazy-load hook `__init__` uses — so the prod hot path stays clean.

**Verification:**
- `uv run python -m pytest tests/`: **2346 passed, 2 skipped** (unchanged from previous head)
- `uv run ruff check .`: All checks passed
- `uv run ty check src/`: All checks passed
- Base-install import without tenacity/daytona: confirmed working via the manual sys.modules-blocking script (and `test_daytona_module_imports_without_tenacity`)
- `grep -rn 'retry_config|_tenacity_retry' src/ tests/`: no matches

**Diff:** 86 insertions, 100 deletions across 5 files (net -14).

bingran-you (approved):
## Review: APPROVE ✅

Both issues fixed with focused regression coverage. Real Daytona e2e verified using `DAYTONA_API_KEY` from local `.env`.

### Verification

**#358 — base install** (fresh venv, no `tenacity` / `daytona` SDK):
- `import benchflow` → `OK 0.5.0.dev0`
- `bench --help` / `bench agent list` → OK
- `import benchflow.sandbox.daytona` → loads without `tenacity`
- `DaytonaSandbox(api_key='x')` (no extras) → clean `ImportError: ... pip install 'benchflow[sandbox-daytona]'`
- Backward-compat `from benchflow.sandbox.daytona import SandboxStartupError` and canonical `from benchflow.sandbox.protocol import SandboxStartupError` resolve to the same class

**#387 — snapshot/restore error propagation** (unit + real Daytona ubuntu:22.04 + sqlite3):
- Failing `sqlite3 .backup` → `EnvironmentSnapshotError(exit_code=1, stderr='boom')` ✅
- Failing `cp` → `EnvironmentSnapshotError` ✅
- Happy-path snapshot → mutate → restore → verify round-trip works ✅
- Exception propagates through `branch.checkpoint()` (verified by inspection + ad-hoc test)

Real Daytona cost: ~$0.01, 2 sandbox lifecycles, ~90s wall-clock.

### Tests
- `tests/test_base_install_imports.py` → **4 passed**
- `tests/environment/test_manifest_env.py` → **23 passed**
- `tests/test_acp.py` → **60 passed**
- Full suite → **2346 passed, 2 skipped, 1 deselected**
- `ruff check .` / `ty check src/` → clean

### Thermo-nuclear findings (non-blocking, suggested follow-ups)

1. **`EnvironmentSnapshotError` placement** — currently in `environment/manifest_env.py:45`. For consistency with this PR's own choice to put `SandboxStartupError` in `sandbox/protocol.py`, the env snapshot error should live in `environment/protocol.py` next to the `snapshot()`/`restore()` contract it formalizes. A second `Environment` impl shouldn't have to import from `manifest_env`.

2. **Inconsistent defensive `_load_daytona_sdk()`** — defended in `_sandbox_exec` (`daytona.py:1353-1357`) but NOT in `_create_sandbox`, `_stop_sandbox`, `_sdk_upload_file`, `_sdk_upload_dir`, `_sdk_download_*`, `_get_session_command_with_retry`, `_get_session_command_logs_with_retry`. Either nothing defends or everything defends — current asymmetry is the worst of both.

3. **`aw.retry_config = retry_kwargs  # type: ignore[attr-defined]`** at `daytona.py:186` is a test-only attribute on a production wrapper. `tests/test_acp.py:924` could introspect via `inspect.unwrap` instead.

4. **`src/benchflow/sandbox/daytona.py` is now 1587 lines** (up from 1489). Pre-existing, not this PR's problem to solve, but worth flagging — `_DaytonaDirect` / `_DaytonaDinD` strategies could live in separate modules.

5. **Missing 5-line regression** — no test covers `EnvironmentSnapshotError` propagating through `branch.checkpoint()` / `rollout_branch.branch()`, which is the core impact claim in #387. Verified manually but a sealed regression would close the loop.

None of these block merge. Flipping `reviewing` → `reviewed` on #358 and #387.

PR #485, opened 2026-05-25
fix: skill-eval correctness sweep (#392 #393 #406 #424 #425 #426)

xdotli:
Fixes #392, #393, #406, #424, #425, #426. Each is independently testable. See commit for details.

## Per-issue summary

- **#392** — `cases[].environment` now lands in the generated task's `[environment.env]` block (and `tests/case.json`) instead of being silently dropped.
- **#393** — `dataset.skill_name` and per-case env keys/values are run through a new `_toml_quote` helper before interpolation. The exact issue repro (skill name containing `" ]\n[agent]\n…`) now produces valid TOML with a single `[agent]` section.
- **#406** — `test.sh.tmpl` now copies `/tests/judge_result.json` into `/logs/verifier/` where the collector reads it, so `CaseResult.rubric_results` is populated when the judge wrote details.
- **#424** — `evals.json` is validated by a strict Pydantic schema (`_EvalsJsonModel` / `_EvalCaseModel` / `_EvalDefaultsModel`) before any TOML/Python/Dockerfile generation. Bad `timeout_sec`, unsafe `judge_model`, `expected_behavior` written as a string, unknown fields, and non-string env values are rejected with one `ValidationError` instead of a downstream parse failure.
- **#425** — `SkillEvaluator._run_job` collects the rollout's `trajectory/acp_trajectory.jsonl` and pairs it with the prompt; `export_gepa_traces` writes `trajectory`, `tool_calls`, and `prompt` into each `traces/*.json` so the artifact actually contains an execution trace.
- **#426** — GEPA trace + summary writes route through a `_json_safe_dumps` helper that recursively scrubs NaN/±Infinity to `null` and sets `allow_nan=False` (same pattern as PR #442 for the trajectory JSONL exporter).

## Tests

New `tests/test_skill_eval_sweep.py` with 22 regression tests, including the exact repros from the linked issues.

```
uv run python -m pytest tests/            # 2361 passed, 2 skipped, 1 deselected
uv run ruff check .                       # All checks passed!
uv run ty check src/                      # All checks passed!
```

---

> [!NOTE]
> **Medium Risk**
> Changes the skill-eval artifact contract (schema, generated tasks, collector paths, GEPA trace shape) used by eval runs and downstream trainers; mitigated by focused regression tests but worth careful review of rollout path assumptions.
> 
> **Overview**
> This PR hardens the **skill-eval** pipeline end-to-end: input validation, safe artifact generation, richer rollout collection, and strict JSON for GEPA/trainer exports.
[...]

PR #487, opened 2026-05-25
fix: harden trace-to-task generated verifiers

### LINKED ISSUE #359: Trace-to-task generation emits weak, unsafe, and stale benchmark artifacts
## Summary

The trace-to-task generator can emit benchmark tasks whose verifier/oracle artifacts are too weak, false-negative, stale, or unsafe. This is high risk because these generated tasks can become evaluation data and then silently reward wrong solutions or fail correct ones.

## Verified repro: content requirement passes with wrong content

Generate the fixture task:

```bash
uv run bench tasks generate \
  --from-file tests/examples/traces/minimal-claude.jsonl \
  --output /tmp/benchflow-generated-tasks \
  --limit 1
```

Generated instruction:

```text
Create hello.txt with exactly: Hello from trace import
```

Generated verifier only checks existence:

```bash
if [ ! -f hello.txt ]; then
  echo "Missing: hello.txt"
  PASS=0
fi
```

Local proof with wrong content:

```bash
mkdir -p /tmp/benchflow-verifier-repro/work /tmp/benchflow-verifier-repro/logs
printf 'wrong\n' > /tmp/benchflow-verifier-repro/work/hello.txt
sed 's#/logs/verifier#/tmp/benchflow-verifier-repro/logs#g' \
  /tmp/benchflow-generated-tasks/create-hello-txt-with-exactly-hello-from-b2c25762/tests/test.sh \
  | (cd /tmp/benchflow-verifier-repro/work && bash)
cat /tmp/benchflow-verifier-repro/logs/reward.txt
```

Actual: `1.0`.

Expected: `0.0`, because the task says the file must contain exactly `Hello from trace import`.

### Git verifier false-fails newly created files

For traces with `git.repo` and `commit_before`, generated `tests/test.sh` uses:

```bash
git diff --name-only HEAD new.txt
```

Untracked newly-created files do not appear in `git diff --name-only HEAD`, so a correct agent/oracle that creates `new.txt` can get reward `0.0`.

Affected code: `src/benchflow/traces/task_gen.py` around `_build_test_sh`.

### Checkout failure is masked, so tasks can benchmark the wrong base commit

Generated Dockerfile code uses:

```dockerfile
git checkout <commit> || true
```

If the commit is outside the shallow depth, private, typoed, or otherwise unavailable, Docker still builds from clone HEAD. The task then evaluates against the wrong repo state.

Affected code: `src/benchflow/traces/task_gen.py` around `_build_dockerfile`.

### SSH remotes are preserved into clean Docker builds

```bash
uv run python -c 'from benchflow.traces.task_gen import _github_clone_url; print(_github_clone_url("git@github.com:octocat/Hello-World.git"))'
```

Actual: `git@github.com:octocat/Hello-World.git`.

A clean Docker build usually cannot use host SSH auth. Either normalize public GitHub SSH remotes to HTTPS or reject unsupported/private remotes explicitly.

### `Edit`/`MultiEdit` oracle solutions write fragments as whole files

`_solution_writes_from_trace()` treats `new_string` as final file content, and `solve.sh` writes it with `cat > file`. For edit tools, `new_string` is often just the replacement fragment, not the whole file.

This can produce broken oracle evidence while the paired verifier only checks “file changed.”

### Trace-controlled metadata is interpolated into shell/Dockerfile without robust escaping

Subagents reproduced trace IDs/newlines escaping comments in generated `tests/test.sh` / `solution/solve.sh`, and unsafe `git.repo` / `commit_before` interpolation into Dockerfile `RUN`.

### `overwrite=True` leaves stale files

`generate_task(..., overwrite=True)` overwrites known files but does not clear the existing task directory. Stale tests/assets can survive into the regenerated task.

## Impact

P1 generated-benchmark correctness/security issue. Trace-generated tasks can:

- award 1.0 for incorrect answers,
- fail correct new-file solutions,
- build against the wrong git state,
- carry broken oracle evidence,
- retain stale artifacts after overwrite,
- and interpolate untrusted trace metadata into executable artifacts.

## Suggested fix

- Generate content-aware checks when write content is known; otherwise mark the verifier as manual/0.0 instead of auto-passing existence.
- For git-backed tasks, distinguish tracked modifications from new files and fail if checkout cannot reach the exact base commit.
- Normalize/reject SSH/private remotes unless credentials are explicitly supported.
- Only emit oracle `solve.sh` for replayable whole-file writes or apply real patches for edits.
- Sanitize/reject trace-controlled metadata before shell/Dockerfile generation.
- Make `overwrite=True` replace the task directory or explicitly preserve only documented files.
- Add regression tests that run generated verifiers against wrong file content, untracked new files, failed checkout, edit-tool traces, unsafe metadata, and overwrite staleness.

bingran-you:
Fixes #359.

## Summary

- Generate exact-content verifier checks for whole-file trace writes instead of existence-only checks.
- Fail closed for edit-only traces without git context, and use `git status --porcelain` for git-backed edit checks so untracked new files count as changed.
- Stop masking git checkout failures in generated Dockerfiles; validate commit metadata before rendering executable artifacts.
- Normalize public GitHub SSH remotes to HTTPS and reject unsupported SSH/non-GitHub remotes.
- Emit oracle `solve.sh` only for replayable whole-file writes so `Edit.new_string` fragments are not treated as full file contents.
- Make `overwrite=True` replace the task directory after rendering succeeds, preventing stale generated files from surviving.

## Verification

```bash
uv sync --extra dev --extra sandbox-daytona --locked
uv run python -m pytest tests/test_traces_task_gen.py tests/test_trace_task_gen_traversal.py
uv run ruff check .
uv run ty check src/
uv run python -m pytest tests/
```

Full test result: `2358 passed, 11 skipped, 1 deselected`.

CLI/e2e verifier repro:

```bash
uv run bench tasks generate --from-file tests/examples/traces/minimal-claude.jsonl --output /tmp/benchflow-generated-tasks-359 --limit 1
# wrong hello.txt content -> reward 0.0
# exact hello.txt content -> reward 1.0
```

Observed: wrong=`0.0`, ok=`1.0`.

This PR targets `v0.5-integration` and should remain open for review.

---

<a href="https://app.devin.ai/review/benchflow-ai/benchflow/pull/487" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/benchflow/traces/task_gen.py
+++ b/src/benchflow/traces/task_gen.py
@@ -7,14 +7,18 @@ with a ``test.sh`` verifier derived from the trace outcome.
 
 from __future__ import annotations
 
+import base64
 import hashlib
+import json
 import logging
 import os
 import re
 import shlex
+import shutil
 import textwrap
+from dataclasses import dataclass
 from pathlib import Path
-from typing import cast
+from urllib.parse import urlparse
 
 from benchflow.traces.models import ParsedTrace
 
@@ -49,6 +53,49 @@ _TIMEOUT_BY_DIFFICULTY = {
     "hard": 1200,
     "expert": 1800,
 }
+
+_WHOLE_FILE_WRITE_TOOLS = {"Write", "write_to_file"}
+_AMBIGUOUS_EDIT_TOOLS = {"Edit", "MultiEdit", "edit_file"}
+_GITHUB_REPO_RE = re.compile(r"^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$")
+_GIT_COMMIT_RE = re.compile(r"^[0-9a-fA-F]{7,40}$")
+
+
+@dataclass(frozen=True)
+class _FileExpectation:
+    """A generated verifier/oracle target extracted from a trace."""
+
+    path: str
+    expected_content: str | None
+
+    @property
+    def has_dynamic_path(self) -> bool:
+        return _has_dynamic_segments(self.path)
+
+    @property
+    def target(self) -> str:
+        return _globify_path(self.path) if self.has_dynamic_path else self.path
+
+    def verifier_payload(self, *, has_git: bool) -> dict[str, str]:
+        if self.expected_content is not None:
+            return {
+                "kind": "content",
+                "path": self.path,
+                "target": self.target,
+                "content_b64": base64.b64encode(self.expected_content.encode()).decode(
+                    "ascii"
+                ),
+            }
+        if has_git:
+            return {
+                "kind": "git_status",
+                "path": self.path,
+                "target": self.target,
+            }
+        return {
+            "kind": "manual",
+            "path": self.path,
+            "target": self.target,
+        }
 
 
 def _estimate_difficulty(trace: ParsedTrace) -> str:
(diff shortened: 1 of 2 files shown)


bingran-you:
## Review: REQUEST_CHANGES 🛑

PR addresses every documented sub-claim in #359 with 9 focused regression tests, all passing. But one of the PR's own stated hardening goals — "sanitize/reject trace-controlled metadata before shell/Dockerfile generation" — is still broken on the TOML half: hand-rolled `_sanitize_toml_string` doesn't escape `\r`, `\t`, NUL, BEL, ESC. Confirmed live via `bench tasks check` producing `task.toml parse error: Illegal character '\r' (at line 11, column 23)` on a `\r`-bearing trace.

### Per-sub-claim verification

| Sub-claim | Status | Evidence |
|---|---|---|
| Existence-only verifier passes wrong content | ✅ | Issue repro: wrong=0.0, correct=1.0. `test.sh` does base64-decoded byte equality (`src/benchflow/traces/task_gen.py:473-481`) |
| Git verifier false-fails newly created files | ✅ | `git status --porcelain` (`task_gen.py:455`); `test_git_status_verifier_accepts_untracked_new_file` covers it |
| `git checkout \|\| true` masks failure | ✅ | Now uses `git fetch --depth 1 origin <commit> && git checkout --detach <commit> && case "$(git rev-parse --verify HEAD)" ...` (`task_gen.py:611-616`); no `\|\| true` in output |
| SSH remotes preserved | ✅ | `git@github.com:…` → `https://github.com/…`; non-GitHub SSH raises `ValueError` (`task_gen.py:632-647`) |
| `Edit.new_string` written as whole file | ✅ | `_file_expectations_from_trace` separates `_WHOLE_FILE_WRITE_TOOLS` from `_AMBIGUOUS_EDIT_TOOLS`; Edit-only trace produces `solve.sh` with "No replayable file writes" |
| Trace-controlled metadata in shell/Dockerfile | 🟡 partial | Shell comments sanitized; commit/repo regex-validated; JSON payloads safe. **TOML is not.** |
| `overwrite=True` leaves stale files | ✅ | `shutil.rmtree(task_dir)` after rendering succeeds (`task_gen.py:730-734`); `test_overwrite_removes_stale_files` |
| Failed render leaves half-regenerated benchmark | ✅ | `_validate_git_commit` raises before `mkdir`/`rmtree`; `test_unsafe_git_commit_rejected_before_overwrite` |

### Blocker — incomplete + duplicated TOML escaping

`src/benchflow/traces/task_gen.py:316-318` `_sanitize_toml_string` only escapes `\\`, `"`, `\n`. Live `bench tasks check` on a generated task with `trace_id = "abc\rdef"` reports `Illegal character '\r' (at line 11, column 23)`.

`tomli_w` is **already a project dep** and `src/benchflow/task/config.py:327-339` already uses `tomli_w.dumps` canonically. This is the same canonical-helper duplication pattern flagged in PR #485 review (`_toml_quote`).

**Fix:**
- Replace `_sanitize_toml_string` + the stringly-typed `toml_lines = [...]` block in `_build_task_toml` with a dict + `tomli_w.dumps(...)`. Removes ~30 lines of brittle string assembly **and** closes the unsanitized-metadata surface in one move.
- Add regression test: trace with `trace_id = "abc\rdef"` → `tomllib.loads(toml)` parses with round-trip equal.

```python
import tomli_w
data = {
    "version": "1.0",
    "task": {"name": task_name},
    "metadata": {"author_name": author, "difficulty": difficulty, ...},
    "agent": {"timeout_sec": timeout_sec},
    "verifier": {"timeout_sec": verifier_timeout_sec},
    "environment": {...},
}
return tomli_w.dumps(data)
```

### Thermo-nuclear findings

1. **(BLOCKING — canonical reuse + incomplete escape)** `_sanitize_toml_string` at `task_gen.py:316-318` reinvents `tomli_w.dumps`. See above.

2. **(Good code judo, noted positively)** `_file_expectations_from_trace` + the new dataclass at `task_gen.py:497-510` collapses three branches in `_build_test_sh` (compgen / git-diff / existence) into one payload-driven flow. Net deletion + clearer invariants. This is the kind of restructuring the rubric wants.

3. **(Minor, non-blocking)** `_build_dockerfile` template at `task_gen.py:599-622` has two branches that both concatenate to `base`. Small code-judo opportunity.

4. **(Minor, non-blocking)** The generated python verifier at `task_gen.py:434-493` is a long format-string. Works and is testable but a module-level template or shipped `.py` would let it be syntax-checked / unit-tested directly.

5. **File size**: 848 lines — under 1000, passes. No flag.

6. **Test design**: All 9 new tests are focused mutation-killers (named with issue/PR number, single behavior per test, positive + negative cases). Missing only the TOML hostile-metadata round-trip test, which would have caught finding #1.

### Tests / lint / type-check
- `tests/test_traces_task_gen.py + tests/test_trace_task_gen_traversal.py`: **56 passed**
- Full suite: **2367 passed, 2 skipped, 1 deselected** in 72s
- `ruff check .` / `ty check src/`: clean

### End-to-end
Ran issue #359's CLI repro verbatim: `uv run bench tasks generate --from-file tests/examples/traces/minimal-claude.jsonl --output /tmp/benchflow-generated-tasks-359 --limit 1`. Verified verifier→oracle→verifier round-trip = 1.0 in plain bash. No real model/Dockerfile build (would need Docker + cost).

### Open questions
- Is the `\r` / control-char metadata case in-scope for this PR? Issue #359 calls out "Trace-controlled metadata is interpolated into shell/Dockerfile without robust escaping" — TOML isn't named explicitly, but it's the same untrusted-input surface.
- `_relativize_path("/etc/passwd")` returns `"passwd"`. Not introduced by this PR but worth flagging — should absolute non-project paths be dropped outright?

Keeping `reviewing` on #359 pending the `tomli_w.dumps` switch + `\r` regression test.

PR #518, opened 2026-05-25
Expose skill invocation counts in rollout artifacts

### LINKED ISSUE #507: Expose skill invocation counts in rollout artifacts
Split from #204.

The #204 analysis suggested making skill usage visible without manually grepping trajectories. Current `v0.5-integration` does not expose a `n_skill_invocations` / `skill_invocations` field in `result.json`, `summary.json`, or the integration result checker.

Acceptance criteria:
- Add a stable skill-usage metric to the canonical rollout/result artifact layer, for example `n_skill_invocations` plus optional per-skill counts if the structured ACP/tool event stream supports it cleanly.
- Derive the metric from structured events rather than brittle text matching.
- Preserve existing result schema compatibility, and update summary/checker tests for the new optional field.
- Document the field in the artifact/reference docs if it becomes user-facing.

Triage evidence: on `origin/v0.5-integration` HEAD `0e8f5d15ddbbebfed8a5de489f873c476f3280fe`, `rg "n_skill_invocations|skill_invocations|Skill\\(|tool_use.*Skill" src tests docs README.md` returns no implementation/documentation matches.

bingran-you:
## Summary

- Add canonical structured trajectory helpers for skill invocation metrics.
- Persist `n_skill_invocations` in `result.json` and nested `agent_result`.
- Aggregate `total_skill_invocations` and `avg_skill_invocations` into `summary.json` and metrics summaries.
- Teach the integration result checker to validate optional skill invocation fields against ACP trajectory JSONL when present.
- Document the user-facing artifact fields.

The count is derived only from structured ACP events: `tool_call` events whose `kind` is exactly `skill`. It intentionally does not parse titles, messages, or other display text.

Closes #507.

## Validation

- `uv run ruff check .`
- `uv run ty check src/`
- `uv run python -m pytest tests/test_skill_invocation_artifacts.py tests/test_integration_check_results.py tests/test_metrics.py tests/test_usage_proxy.py tests/test_acp.py tests/test_capture_trajectory.py`
- `uv run python -m pytest tests/` -> 2394 passed, 2 skipped, 1 deselected
- Local ACP smoke with a stdio mock agent emitting one `kind: "skill"` tool call and one `kind: "bash"` tool call -> `{"artifact_n_skill_invocations": 1, "n_skill_invocations": 1, "n_tool_calls": 2}`

---

<a href="https://app.devin.ai/review/benchflow-ai/benchflow/pull/518" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/docs/running-benchmarks.md
+++ b/docs/running-benchmarks.md
@@ -408,10 +408,17 @@ The `result.json` contains:
 ```json
 {
   "rewards": {"reward": 0.48},
+  "n_tool_calls": 12,
+  "n_skill_invocations": 2,
   "passed": true,
   "verifier_output": "..."
 }
 ```
+
+`n_skill_invocations` is derived from structured ACP trajectory events: BenchFlow
+counts only `tool_call` events whose `kind` is `skill`. Job `summary.json`
+also includes `total_skill_invocations` and `avg_skill_invocations` across the
+rollouts in the run.
 
 List evaluations:
 ```bash
--- a/src/benchflow/_utils/evaluation_results.py
+++ b/src/benchflow/_utils/evaluation_results.py
@@ -9,12 +9,20 @@ from benchflow._utils.reward_events import (
     reward_event_to_dict,
 )
 from benchflow.models import RolloutResult
+from benchflow.trajectories.metrics import (
+    count_skill_invocations,
+    result_skill_invocations,
+)
 
 
 def agent_result_from_rollout(result: RolloutResult) -> dict[str, Any]:
     """Return the serialized agent_result block for an in-memory rollout result."""
+    n_skill_invocations = result.n_skill_invocations or count_skill_invocations(
+        result.trajectory
+    )
     return {
         "n_tool_calls": result.n_tool_calls,
+        "n_skill_invocations": n_skill_invocations,
         "n_prompts": result.n_prompts,
         "n_input_tokens": result.n_input_tokens,
         "n_output_tokens": result.n_output_tokens,
@@ -40,6 +48,9 @@ def rollout_result_payload(
     task_source = result.source_provenance or task_source_provenance(
         source_provenance, tasks_dir / task_name
     )
+    n_skill_invocations = result.n_skill_invocations or count_skill_invocations(
+        result.trajectory
+    )
     return {
         "task_name": result.task_name,
         "rewards": result.rewards,
@@ -47,6 +58,7 @@ def rollout_result_payload(
         "verifier_error": result.verifier_error,
         "export_error": result.export_error,
         "n_tool_calls": result.n_tool_calls,
+        "n_skill_invocations": n_skill_invocations,
         "agent_result": agent_result_from_rollout(result),
         **(
             {"reward_events": [reward_event_to_dict(event) for event in reward_events]}
@@ -91,4 +103,15 @@ def usage_summary(results: dict[str, dict]) -> dict[str, Any]:
             round(total_cost / len(covered), 10) if covered else None
         ),
         "telemetry_coverage": (len(covered) / len(completed) if completed else 0.0),
+    }
+
+
+def skill_invocation_summary(results: dict[str, dict]) -> dict[str, Any]:
+    """Aggregate structured skill invocation counts for summary.json."""
+    total = sum(result_skill_invocations(result) for result in results.values())
+    return {
+        "total_skill_invocations": total,
+        "avg_skill_invocations": (
+            round(total / len(results), 1) if results else 0.0
+        ),
     }
--- a/src/benchflow/acp/types.py
+++ b/src/benchflow/acp/types.py
@@ -78,6 +78,7 @@ class ToolKind(StrEnum):
     BROWSER = "browser"
     READ = "read"
     WRITE = "write"
+    SKILL = "skill"
 
 
 class ToolCallStatus(StrEnum):
--- a/src/benchflow/evaluation.py
+++ b/src/benchflow/evaluation.py
@@ -26,6 +26,7 @@ import yaml
 
 from benchflow._utils.evaluation_results import (
     rollout_result_payload,
+    skill_invocation_summary,
     usage_summary,
 )
 from benchflow._utils.learner_memory import (
@@ -1224,6 +1225,7 @@ class Evaluation:
             ),
             "memory": memory,
             "memory_scores": memory_scores,
+            **skill_invocation_summary(all_results),
             **usage_summary(all_results),
             **summary_source_fields(cfg.source_provenance, all_results),
         }
--- a/src/benchflow/metrics.py
+++ b/src/benchflow/metrics.py
@@ -20,6 +20,7 @@ from benchflow._utils.scoring import (
     pass_rate,
     pass_rate_excl_errors,
 )
+from benchflow.trajectories.metrics import result_skill_invocations
 
 logger = logging.getLogger(__name__)
 
@@ -31,6 +32,7 @@ class TaskMetrics:
     task_name: str
     reward: float | None = None
     n_tool_calls: int = 0
+    n_skill_invocations: int = 0
     n_prompts: int = 0
     error: str | None = None
     verifier_error: str | None = None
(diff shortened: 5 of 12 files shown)


bingran-you:
## Verdict: REQUEST_CHANGES

This review combines #507 verification, test results, and a thermo-nuclear pass. The plumbing is correct and tests pass, but the metric's *source of truth* is misaligned with the ACP spec, which makes the field functionally a constant `0` in practice. That's the blocking concern.

### #507 verification (fixture path)

Real Docker + Gemini rollout was skipped because (a) the structural issue below means a real rollout would still report `0`, and (b) fixture-level verification already exercises the full data path. Synthesized a 3-event trajectory (`bash`, `skill`, `skill`) through `_build_rollout_result` + `rollout_result_payload` + `skill_invocation_summary`:

```
result.json:
{
  "n_tool_calls": 3,
  "n_skill_invocations": 2,
  "agent_result.n_skill_invocations": 2
}
summary.json fragment: {
  "total_skill_invocations": 2,
  "avg_skill_invocations": 2.0
}
```

So the *plumbing* is correct given an event whose `kind == "skill"`.

### Tests / lint / type-check

- `uv run ruff check .` -> All checks passed
- `uv run ty check src/` -> All checks passed
- `uv run python -m pytest tests/` -> 2385 passed, 11 skipped, 1 deselected

### Thermo-nuclear findings (prioritized)

**1. BLOCKER: `kind: "skill"` is not in the ACP spec — counter will always be 0 in production.**

ACP schema (latest, v0.13.3) defines `ToolKind` as exactly:
```
read, edit, delete, move, search, execute, think, fetch, switch_mode, other
```
"skill" is not a member. The pinned `agent-client-protocol>=0.10` SDK (`acp/schema.py:16`) confirms the same literal. The PR's source of truth (`event.get("kind") == "skill"`) therefore only matches synthesized test fixtures — no ACP-compliant agent will ever emit it. A grep across `src/` confirms zero code paths write `kind="skill"` (only `acp/types.py:81` adds the enum member; nothing emits it).

The `openclaw_acp_shim` (line 441) writes `"kind": block.get("name", "tool")` — i.e. the raw upstream tool *name* (e.g. `"Read"`, `"Bash"`, `"Skill"`), not a normalized ACP kind. So if a Claude/Anthropic SDK ever surfaces a tool literally named `Skill` (capital S), the shim would write `kind: "Skill"` and the counter would *still* return 0 because the comparison is case-sensitive (`== "skill"`).

The acceptance criterion from #507 was "Derive the metric from structured events rather than brittle text matching." The current implementation satisfies the letter (no text matching) but not the spirit — the "structured" channel it depends on does not exist.

Concrete remedies (pick one):
- (a) Define what `kind: "skill"` actually means and *emit* it somewhere: e.g. detect Skill tool invocations in the shim/runtime by `block.name == "Skill"` (or whatever the upstream marker is) and normalize to `kind="skill"` before persisting.
- (b) Change the counter to match the real signal: e.g. `tc.title.startswith("Skill(")` or `(event.get("kind") or "").lower() == "skill"` plus title detection. This is "text matching" which the issue explicitly forbids — so (a) is the right path.
- (c) Don't ship this until the upstream ACP spec defines a skill kind (the PR is structurally premature). 

This finding alone is enough to block.

**2. Parallel/redundant counter — three computations of the same value per result.**

`count_skill_invocations(trajectory)` is invoked in:
- `rollout.py:520` (stores into `RolloutResult.n_skill_invocations`)
- `_utils/evaluation_results.py:20` (`agent_result_from_rollout`: `result.n_skill_invocations or count_skill_invocations(...)`)
- `_utils/evaluation_results.py:51` (`rollout_result_payload`: same pattern)

In the common case of zero skill invocations (which is *every* case in production today, per #1), `result.n_skill_invocations` is `0` (falsy) so the `or` triggers a second AND third full-trajectory scan per result. Trivial per-result, but a real (if small) waste because trajectories can be thousands of events for long rollouts.

The `or` fallback exists to handle test paths that construct `RolloutResult` without `n_skill_invocations` (e.g. `test_eval_source_provenance.py`). Cleaner shape:
- Make `RolloutResult.n_skill_invocations: int | None = None`, set it canonically at one site (`_build_rollout_result`), and have downstream serializers trust it (or 0). One source of truth, no `or` fallback.

The reviewer's "thermo-nuclear" lens: this is a parallel counter sitting alongside `n_tool_calls = len(session.tool_calls)`. The natural single-source-of-truth would be `sum(1 for tc in session.tool_calls if tc.kind == "skill")` computed once in the same place as `n_tool_calls` (i.e. derived from `ACPSession.tool_calls` directly, not from the post-processed dict trajectory). This also dovetails with finding #1: putting the count next to the ACPSession source forces the question of where `kind="skill"` actually originates.

**3. Schema drift: BenchFlow `ToolKind` enum diverges from ACP spec without justification.**

`src/benchflow/acp/types.py:72-81` defines:
```
OTHER, BASH, SEARCH, BROWSER, READ, WRITE, SKILL
```
ACP defines:
```
read, edit, delete, move, search, execute, think, fetch, switch_mode, other
```
Mismatched: BenchFlow has `BASH/BROWSER/WRITE/SKILL` which aren't in ACP; ACP has `EDIT/DELETE/MOVE/EXECUTE/THINK/FETCH/SWITCH_MODE` which aren't in BenchFlow's enum. The local enum is also functionally unused (only one consumer: a default in a dataclass field, never compared against). Adding `SKILL` to a vestigial enum doesn't fix this divergence and isn't referenced by the counter (which compares strings, not enum members).

Remedy: either delete the local enum and use the SDK's `ToolKind` literal directly, or document why BenchFlow's superset exists and have the counter use `ToolKind.SKILL.value` so the constant is at least linked.

**4. Test design — the two new test files have ~30% duplication.**

`tests/test_skill_invocation_artifacts.py` and `tests/test_integration_check_results.py` (new tests) each construct nearly identical `result.json` + `acp_trajectory.jsonl` fixtures by hand. Three of the new pytest functions in `check_results` (`test_check_results_accepts_skill_invocation_metrics`, `test_check_results_flags_skill_invocation_mismatch`, plus existing `_write_config` helper) each inline a 20-line `summary.json` dict that's 95% identical. Worth extracting a `_skill_invocation_fixture` helper.

The functional coverage is fine — both happy path and mismatch detection are tested. Just padded.

**5. Fan-out justification: 12 files is reasonable for this feature.**

Files split into:
- 1 new metric helper (`trajectories/metrics.py`)
- 1 schema (`acp/types.py` — though it does nothing, see #3)
- 4 data plumbing layers (`models.py`, `rollout.py`, `evaluation.py`, `metrics.py`) — necessary for the field to flow from `RolloutResult` to `summary.json`
- 2 serialization helpers (`_utils/evaluation_results.py`, `tests/integration/check_results.py`)
- 4 docs / tests

Boundary fan-out is fine; *count* fan-out (the same `count_skill_invocations` call appearing in 3 sites) is the spaghetti smell (see #2).

### Recommended remedies (concrete)

1. **Block on #1 until resolved.** Either (a) emit `kind="skill"` from a real producer in this repo (probably an Anthropic-skill-aware shim or a post-processing pass over `ACPSession.tool_calls` that recognizes the Skill tool by name), then keep the counter; or (b) document this PR as "scaffolding ahead of ACP adopting a skill kind" and leave it explicitly behind a feature flag / marked experimental in the docs. The current README claims `total_skill_invocations` is meaningful — it isn't, today.

2. **Collapse the counter to one site.** Compute `n_skill_invocations` once in `_build_rollout_result` (or, better, alongside `n_tool_calls` in `ACPSession`/`execute_prompts`). Remove the `or count_skill_invocations(...)` fallbacks in `_utils/evaluation_results.py`. Trust the field.

3. **Reconcile `acp/types.py:ToolKind` with the SDK.** Either delete it or document the superset and have the counter reference `ToolKind.SKILL.value` so grep finds the link.

4. **Extract the test fixture helper** in `tests/test_integration_check_results.py` to remove the 50+ lines of duplicated dict literals across the two new tests (and the existing tests that already build `summary.json` by hand).

Issue #555, opened 2026-05-25
check_results.py hard-fails passing local tasks-dir runs when source provenance is missing


bingran-you:
## Summary

`tests/integration/check_results.py` treats a fully passing local `--tasks-dir` eval as FAIL when BenchFlow cannot infer GitHub source provenance for that local directory. This contradicts `docs/v05-e2e-testing-guide.md` §13, which says source provenance warnings are expected for local runs.

The behavior is especially confusing because the same task and command shape passes when the local tasks directory happens to live under BenchFlow's recognized dataset cache (`.cache/datasets/benchflow-ai/skillsbench`), but fails when SkillsBench is cloned to an arbitrary local path such as `/tmp/.../skillsbench/tasks`.

## Repro

Use a normal local clone of SkillsBench outside BenchFlow's dataset cache:

```bash
TASKS=/tmp/benchflow-e2e-sidecar-20260525-113850/skillsbench-src/skillsbench/tasks

uv run bench eval create \
  --tasks-dir "$TASKS" \
  --include weighted-gdp-calc \
  --include shock-analysis-supply \
  --exclude shock-analysis-supply \
  --agent oracle \
  --sandbox daytona \
  --concurrency 1 \
  --jobs-dir /tmp/benchflow-e2e-sidecar-20260525-113850/jobs/oracle-include-exclude

uv run python tests/integration/check_results.py \
  /tmp/benchflow-e2e-sidecar-20260525-113850/jobs/oracle-include-exclude/2026-05-25__11-46-13/ \
  agent=oracle model=null environment=daytona concurrency=1
```

Observed output:

```text
Agent                        Score  Pass  Fail   Err   Status
--------------------------------------------------------------
2026-05-25__11-46-13        100.0%     1     0     0     FAIL
  ⚠ weighted-gdp-calc: result.json: missing source provenance
  ⚠ weighted-gdp-calc: config.json: missing source provenance
  ⚠ summary.json missing source provenance and not all results have source
--------------------------------------------------------------

Some agents had issues. See warnings above.
```

The underlying `result.json` has `reward=1.0`, no error, and valid verifier artifacts, but `source` is `null` because the arbitrary local clone path was not recognized as a source repo/cache path.

## Expected

Pick one contract and make it explicit:

1. If local `--tasks-dir` provenance warnings are expected per the guide, `check_results.py` should report them as warnings without making a fully passing local eval FAIL; or
[...]

PR #563, opened 2026-05-26
fix: persist Daytona sandbox IDs for audit and cleanup

### LINKED ISSUE #554: Daytona runs do not persist sandbox ids, making interrupted sandbox cleanup unauditable
## Summary

Interrupted Daytona runs can leave fresh `STARTED` sandboxes behind, but BenchFlow's local artifacts/logs do not record the Daytona sandbox id for each rollout attempt. After an interrupted or killed eval process, an operator cannot safely tell which live Daytona sandboxes belong to the just-interrupted BenchFlow run and which belong to unrelated work.

This is an auditability/cleanup gap in the real E2E path: the sandbox exists, the run logs show process ids and retry attempts, but the durable job artifact has no stable provider-side sandbox id to use for post-mortem cleanup.

## Repro

Run a Daytona eval and interrupt it after sandbox creation/agent startup, or kill a run that is retrying provider/transport failures:

```bash
export TASKS=.cache/datasets/benchflow-ai/skillsbench/tasks
uv run bench eval create \
  --tasks-dir "$TASKS" \
  --include weighted-gdp-calc \
  --agent opencode --model vllm/gpt-5.5 \
  --sandbox daytona \
  --jobs-dir /tmp/test-opencode-transport

# interrupt/kill while it is repeatedly starting Daytona processes
```

Observed in this E2E pass:

```text
Starting environment: weighted-gdp-calc
Installing opencode in sandbox (timeout=900s)...
Setting up sandbox user: agent
Sandbox user agent ready (workspace=/root)
Daytona process started (pid=94274)
Process terminated
ACP connect failed (attempt 1): Process closed stdout (rc=None) ...
ACP connect retry 1/3 after 2s
Daytona process started (pid=94591)
...
Daytona process started (pid=95305)
Agent connection lost: Process closed stdout (rc=None) ...
Retrying weighted-gdp-calc (attempt 2) ...
```

The durable artifacts have `result.json`, `config.json`, `timing.json`, `trajectory/acp_trajectory.jsonl`, and `agent/*.txt`, but no Daytona sandbox id. `transport_error_info` includes process pid/rc and sandbox probe fields, but not the sandbox id. `sandbox_startup_info.sandbox_id` is only populated for startup failures; successful sandbox creation followed by transport failure/interruption has no equivalent durable id.

## Expected

Every Daytona rollout attempt should persist the provider-side sandbox id in at least one stable artifact, ideally:

- `config.json` or a small `sandbox.json` written immediately after sandbox creation,
- `result.json.sandbox` / `result.json.sandbox_info` for completed/error attempts,
- `transport_error_info.sandbox_id` and/or a general attempt diagnostic when a transport failure happens after successful sandbox creation,
- logs should print the sandbox id alongside the local process pid.

This must be redaction-safe: sandbox ids are operational identifiers, not API keys, and are necessary for cleanup/audit.

## Why this matters

- Operators cannot safely clean up leaked/STARTED Daytona sandboxes after a local interruption without risking deleting someone else's live sandbox.
- Retry storms (#546/#538) multiply the number of possible orphan sandboxes, and the artifacts only show local pids, not provider ids.
- Audit zips are incomplete: they prove a Daytona run happened, but not which provider resource it used.
- Transport diagnostics can say the sandbox was reachable, but not identify which sandbox was probed.

## Suggested fix

Create a provider-neutral sandbox attempt metadata object and write it as soon as sandbox creation succeeds. For Daytona, include `sandbox_id`, provider name, creation/start timestamps, and cleanup status if known. Thread that metadata into `result.json` and diagnostics so even interrupted/error runs have enough information to reconcile against the Daytona dashboard/API.

## Found via

v0.5 E2E audit on `v0.5-integration` at `d8fb7af8379e1a71b3e2f170c8e15c64e9fcd026`, while running real Daytona tasks against SkillsBench `3f86918e2de85e07bf2a6b817a7d00e92aba76c1`. Sidecar audit also independently observed fresh `STARTED` Daytona sandboxes after interrupted runs with no local sandbox-id trail.

ElegantLin:
## Summary

- Daytona runs now write `sandbox.json` immediately after sandbox creation, ensuring the provider-side sandbox ID survives process interruptions.
- `result.json` also includes `sandbox_id` for completed runs, linking rollout artifacts to provider-side resources.
- Adds a `sandbox_id` property to `BaseSandbox` (returns `None` by default), overridden in `DaytonaSandbox` to return `self._sandbox.id`.
- Uses `isinstance(sandbox_id, str)` checks to safely handle mocked environments in tests.

Closes #554

## Test plan

- [x] `test_persist_sandbox_info_writes_sandbox_json` — verifies sandbox.json is written with correct ID and provider
- [x] `test_persist_sandbox_info_skips_when_no_sandbox_id` — Docker backends produce no sandbox.json
- [x] `test_persist_sandbox_info_skips_when_no_rollout_dir` — no crash on None rollout_dir
- [x] `test_result_json_includes_sandbox_id` — sandbox_id appears in result.json
- [x] `test_result_json_sandbox_id_null_for_docker` — sandbox_id is null when not provided
- [x] Full test suite passes (1371 passed, 9 pre-existing failures unrelated)

---

<a href="https://app.devin.ai/review/benchflow-ai/benchflow/pull/563" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/benchflow/rollout.py
+++ b/src/benchflow/rollout.py
@@ -805,7 +805,16 @@ def _persist_sandbox_info(env: Any, rollout_dir: Path | None) -> None:
     """Write sandbox.json with provider-side sandbox ID immediately after creation.
 
     This ensures the sandbox ID is persisted even if the run is interrupted
-    before result.json is written.
+    before result.json is written. Daytona also calls this at the moment its
+    sandbox is created (inside ``start()``), not only after ``start()`` returns
+    — closing the in-``start()`` interrupt window for DinD's long daemon-wait
+    (#554/#563).
+
+    The ``sandbox.json`` write is best-effort: a failure to write it (disk full,
+    read-only dir) must not abort an otherwise-healthy run, so it is wrapped and
+    logged rather than raised. The id-capture itself is load-bearing and lives in
+    the caller (``Rollout.start``); this helper only does the best-effort file
+    write.
     """
     sandbox_id = getattr(env, "sandbox_id", None)
     if not isinstance(sandbox_id, str) or not rollout_dir:
@@ -816,7 +825,11 @@ def _persist_sandbox_info(env: Any, rollout_dir: Path | None) -> None:
         "provider": provider,
         "created_at": str(datetime.now()),
     }
-    (rollout_dir / "sandbox.json").write_text(json.dumps(info, indent=2))
+    try:
+        (rollout_dir / "sandbox.json").write_text(json.dumps(info, indent=2))
+    except Exception as e:  # best-effort: never fail start() over an audit file
+        logger.warning(f"Failed to persist sandbox.json for {sandbox_id}: {e}")
+        return
     logger.info(f"Sandbox {sandbox_id} ({provider})")
 
 
--- a/src/benchflow/rollout.py
+++ b/src/benchflow/rollout.py
@@ -2703,6 +2703,8 @@ class Rollout:
                 runtime_skills_dir=self._config.skills_dir,
                 declared_sandbox_skills_dir=None,
             ),
-            sandbox_id=self._sandbox_id,
+            # getattr guard: cleanup/export paths build a Rollout skeleton via
+            # __new__ (no __init__), so _sandbox_id may be unset there (#563).
+            sandbox_id=getattr(self, "_sandbox_id", None),
             **self._usage_metrics,
         )
--- a/src/benchflow/rollout.py
+++ b/src/benchflow/rollout.py
@@ -521,6 +521,7 @@ def _build_rollout_result(
     source_provenance: dict[str, Any] | None = None,
     diagnostics: RolloutDiagnostics | None = None,
     skill_policy: TaskSkillPolicy | None = None,
+    sandbox_id: str | None = None,
 ) -> RolloutResult:
     """Build RolloutResult and write result.json, timing.json, prompts.json, trajectory.
 
@@ -640,6 +641,7 @@ def _build_rollout_result(
                     if source_provenance is not None
                     else {}
                 ),
+                "sandbox_id": sandbox_id,
             },
             indent=2,
         )
@@ -783,6 +785,25 @@ async def _start_env_and_upload(
         await env.upload_file(task_path / "instruction.md", "/instruction.md")
     if (task_path / "solution").is_dir():
         await env.upload_dir(task_path / "solution", "/solution")
+
+
+def _persist_sandbox_info(env: Any, rollout_dir: Path | None) -> None:
+    """Write sandbox.json with provider-side sandbox ID immediately after creation.
+
+    This ensures the sandbox ID is persisted even if the run is interrupted
+    before result.json is written.
+    """
+    sandbox_id = getattr(env, "sandbox_id", None)
+    if not isinstance(sandbox_id, str) or not rollout_dir:
+        return
+    provider = type(env).__name__
+    info = {
+        "sandbox_id": sandbox_id,
+        "provider": provider,
+        "created_at": str(datetime.now()),
+    }
+    (rollout_dir / "sandbox.json").write_text(json.dumps(info, indent=2))
+    logger.info(f"Sandbox {sandbox_id} ({provider})")
 
 
 async def _run_oracle(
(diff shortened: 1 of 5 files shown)


bingran-you (changes requested):

This PR is currently not mergeable with `origin/main` (`src/benchflow/rollout.py` conflicts), and the sandbox-id persistence is still too late to protect the failure mode it is meant to audit.

The current flow starts the env and uploads files before `_persist_sandbox_info()` runs. If upload fails, or the process is interrupted after Daytona creation but before persistence, there is still no `sandbox.json`, so cleanup/audit loses the Daytona sandbox id.

I reproduced this with a fake env: after a start/upload failure, `sandbox.json` was still absent; calling `_persist_sandbox_info()` manually afterwards created it. That means the happy path is covered, but the failure window remains.

Please persist the sandbox id immediately after the environment starts and before upload/setup steps that can fail.

Additional repo-convention issue: the new regression test docstrings mention PR #555; they should name this reviewed PR (#563) or the guarding commit.

Verification notes:
- Branch-head checks passed: `uv sync`, full `pytest`, `ty`, `ruff`, `git diff --check`.
- Manual Codex ACP smoke passed: reward `1.0`, `trajectory_source=acp`, 2 tool calls, usage `provider_response`, token/cost fields populated.
- Daytona happy-path smoke created `sandbox.json` and `result.json` with a sandbox id, but the failure-path reproduction above is the blocker.

bingran-you:

This PR is trying to close #554 by making Daytona provider sandbox IDs recoverable for audit/cleanup when a rollout is interrupted before the normal `result.json` write. The proposed solution adds a `sandbox_id` property to the sandbox abstraction, persists `sandbox.json`, and threads the captured ID into completed `result.json` artifacts.

The direction is right, but I still see one blocking lifecycle gap.

### Blocking finding

**`sandbox.json` is still not persisted immediately after Daytona sandbox creation.**

The new callback runs only after `await env.start(force_build=False)` returns in `src/benchflow/rollout.py:748-751`. For Daytona, provider creation is only the first part of `start()`: direct mode calls `_create_sandbox()` at `src/benchflow/sandbox/daytona.py:431`, then still runs `_sandbox_exec(...)` at `src/benchflow/sandbox/daytona.py:442`; DinD mode calls `_create_sandbox()` at `src/benchflow/sandbox/daytona.py:792`, then starts/waits for Docker, uploads compose/env files, builds, runs compose, and waits for the main container at `src/benchflow/sandbox/daytona.py:803-851`.

Any failure, timeout, or process interruption in that post-create setup window still leaves a live Daytona sandbox without `sandbox.json`, which is the exact audit/cleanup failure this PR is meant to eliminate. The regression test covers upload failure after `on_started` has fired, but it does not cover a failure inside `env.start()` after `_create_sandbox()` succeeds and before `start()` returns.

**Handoff:** move the persistence event to the actual provider-created boundary, not the upload boundary. The cleanest shape is probably a sandbox lifecycle hook/observer or a split create/prepare lifecycle so Daytona can invoke the callback immediately after `_create_sandbox()` and before the next awaited setup step. Add a regression where a fake/Daytona-like env sets `sandbox_id` during `start()` and then raises before returning; `sandbox.json` should already exist. Do not rely only on catching `env.start()` exceptions in rollout, because interruption during Daytona's post-create awaits would still miss the ID.

### Maintainability notes

`rollout.py` is already very large, and the new `on_started` callback adds another lifecycle special case to a busy orchestration file. I would be less concerned if it represented the real sandbox-created event, but right now the abstraction name and placement are misleading: it says “started,” while the artifact requirement is “provider resource allocated.” Put the boundary where that invariant is actually established.

No new file crosses from under 1k lines to over 1k lines, and `BaseSandbox.sandbox_id` itself is a reasonable small contract. The blocker is the lifecycle boundary, not the idea of recording the ID.

### Residual tests needed

After moving the hook, please cover Daytona/direct post-create setup failure, Daytona/DinD post-create setup failure, the existing upload-failure regression, Docker/no-provider-ID behavior, and completed Daytona runs with matching `sandbox.json` and top-level `result.json["sandbox_id"]`.

Verdict: request changes. The PR fixes the upload-failure window, but not the broader “sandbox created before persistence” window claimed in the description.

ElegantLin:
Follow-up (`2a74031`): the `on_started` callback only fired after `env.start()` returned, leaving a residual interrupt window inside `start()` (DinD daemon-wait, mkdir, uploads — tens of seconds) where the sandbox exists but no `sandbox.json`. Now `DaytonaSandbox` persists its id the moment it's created (`_on_sandbox_created()` right after `self._sandbox` is assigned, both the normal and CancelledError-recovery paths). Also made the write best-effort (non-fatal) and corrected the `sandbox_id` docstring. New test drives the real `_create_sandbox` wiring (verified to fail without the fix).

bingran-you (approved):

PR #564, opened 2026-05-26
fix: classify provider auth failures as non-retryable

### LINKED ISSUE #546: Provider auth failures retry three Daytona sandboxes instead of failing fast
## Summary

When Gemini returns a non-retryable provider auth failure (`403 PERMISSION_DENIED`, message: "Your API key was reported as leaked. Please use another API key."), BenchFlow treats it like a normal retryable rollout error and starts new Daytona sandboxes for attempts 2 and 3.

This wastes Daytona time/quota and delays the real operator action: rotate the provider key.

## Repro

Use a Gemini key that the provider has revoked/reported as leaked, then run any Gemini × Daytona task:

```bash
export TASKS=.cache/datasets/benchflow-ai/skillsbench__snapshots/<sha>/tasks
uv run bench eval create \
  --tasks-dir "$TASKS" \
  --include weighted-gdp-calc \
  --agent gemini --model gemini-2.5-flash \
  --sandbox daytona \
  --jobs-dir /tmp/test-provider-auth
```

Observed from the v0.5 E2E run:

```text
ACP error 403: ... "Your API key was reported as leaked. Please use another API key." ...
Retrying weighted-gdp-calc (attempt 2): ACP error 403 ...
Retry backoff: 2.0s before attempt 2
Starting environment: weighted-gdp-calc
...
Retrying weighted-gdp-calc (attempt 3): ACP error 403 ...
Retry backoff: 4.0s before attempt 3
Starting environment: weighted-gdp-calc
...
[ERR] weighted-gdp-calc (tools=0)
```

The same pattern reproduced for `quantum-numerical-simulation`, `weighted-gdp-calc`, and self-gen creator runs after the key was rejected.

## Expected

Provider authentication/authorization failures should be classified as non-retryable. The job should stop after the first failed attempt, mark the result with a specific category (for example `provider_auth`), and surface a clear operator action: rotate/fix credentials.

## Actual

BenchFlow performs the full retry budget and provisions a fresh Daytona sandbox for each attempt even though the provider message says the key must be replaced.

## Why this matters

- Burns remote sandbox quota and wall-clock time.
- Obscures the root cause behind repeated infra-looking retries.
- Makes long E2E suites much noisier after credentials are invalidated.

## Suggested fix

Add provider-auth classification for ACP errors containing provider 401/403 auth markers, and make those categories non-retryable in eval retry policy. Preserve the raw provider message in a redacted/sanitized diagnostic field, not as a generic `acp_error`/`other` string.

## Found via

v0.5 Daytona E2E run on `61fa09e6973f44f01e9a3d21773e302316257e39` after the provided Gemini key was rejected by the provider.

ElegantLin:
## Summary

- Provider auth errors (Gemini 403 `PERMISSION_DENIED`, Claude 401 `Failed to authenticate`, leaked/invalid API keys) were classified as generic `acp_error` and retried up to 3 times, each spinning up a fresh Daytona sandbox.
- Added `PROVIDER_AUTH` error category in `scoring.py` with auth-specific markers (`PERMISSION_DENIED`, `Invalid API key`, `Failed to authenticate`, `Invalid bearer token`, `leaked`, `Unauthorized`, `was rejected as invalid`).
- `classify_error()` now detects these markers within ACP errors and returns `provider_auth` instead of `acp_error`.
- `RetryConfig.exclude_categories` defaults to `{"timeout", "provider_auth"}`, so auth failures fail fast on the first attempt.

Closes #546

## Test plan

- [x] `test_provider_auth_gemini_403` — Gemini 403 PERMISSION_DENIED classified as provider_auth
- [x] `test_provider_auth_claude_401` — Claude 401 invalid bearer token classified as provider_auth
- [x] `test_provider_auth_invalid_api_key` — generic invalid API key classified as provider_auth
- [x] `test_provider_auth_rejected_as_invalid` — BenchFlow's own subscription auth message classified as provider_auth
- [x] `test_acp_error_non_auth_still_retryable` — generic ACP errors remain retryable
- [x] `test_should_not_retry_provider_auth_gemini_403` — RetryConfig rejects Gemini auth
- [x] `test_should_not_retry_provider_auth_claude_401` — RetryConfig rejects Claude auth
- [x] Full test suite passes (1373 passed, 9 pre-existing failures unrelated)

---

<a href="https://app.devin.ai/review/benchflow-ai/benchflow/pull/564" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/benchflow/_utils/scoring.py
+++ b/src/benchflow/_utils/scoring.py
@@ -10,7 +10,18 @@ ACP_ERROR = "acp_error"
 IDLE_TIMEOUT = "idle_timeout"
 INFRA_ERROR = "infra_failure"
 SANDBOX_SETUP = "sandbox_setup"
+PROVIDER_AUTH = "provider_auth"
 TIMED_OUT = "timeout"
+
+_PROVIDER_AUTH_MARKERS = (
+    "PERMISSION_DENIED",
+    "leaked",
+    "Failed to authenticate",
+    "Invalid bearer token",
+    "Invalid API key",
+    "was rejected as invalid",
+    "Unauthorized",
+)
 
 # Verifier error category constants
 VERIFIER_FAILED = "verifier_failure"
@@ -41,7 +52,9 @@ def classify_error(error: str | None) -> str | None:
         return INSTALL_FAILED
     if "closed stdout" in lower:
         return PIPE_CLOSED
-    if "ACP error" in error:
+    if "ACP error" in error or "was rejected as invalid" in error:
+        if any(m in error for m in _PROVIDER_AUTH_MARKERS):
+            return PROVIDER_AUTH
         return ACP_ERROR
     if "sandbox startup" in lower or "sandbox creation" in lower:
         return SANDBOX_SETUP
--- a/src/benchflow/_utils/scoring.py
+++ b/src/benchflow/_utils/scoring.py
@@ -13,14 +13,21 @@ SANDBOX_SETUP = "sandbox_setup"
 PROVIDER_AUTH = "provider_auth"
 TIMED_OUT = "timeout"
 
+# Matched case-insensitively against the error string. Covers the
+# human-authored markers plus the sanitized "provider auth failed (HTTP 401)"
+# marker injected at the rollout/provider boundary (#546/#564), where the real
+# 401/403 is visible only in the proxy trajectory, not the top-level ACP error.
 _PROVIDER_AUTH_MARKERS = (
-    "PERMISSION_DENIED",
+    "permission_denied",
     "leaked",
-    "Failed to authenticate",
-    "Invalid bearer token",
-    "Invalid API key",
+    "failed to authenticate",
+    "invalid bearer token",
+    "invalid api key",
     "was rejected as invalid",
-    "Unauthorized",
+    "unauthorized",
+    "provider auth failed",
+    "http 401",
+    "http 403",
 )
 
 # Verifier error category constants
@@ -53,7 +60,7 @@ def classify_error(error: str | None) -> str | None:
     if "closed stdout" in lower:
         return PIPE_CLOSED
     if "ACP error" in error or "was rejected as invalid" in error:
-        if any(m in error for m in _PROVIDER_AUTH_MARKERS):
+        if any(m in lower for m in _PROVIDER_AUTH_MARKERS):
             return PROVIDER_AUTH
         return ACP_ERROR
     if "sandbox startup" in lower or "sandbox creation" in lower:
--- a/src/benchflow/eval_worker.py
+++ b/src/benchflow/eval_worker.py
@@ -24,20 +24,11 @@ logging.basicConfig(level=logging.INFO, format="%(message)s")
 
 
 def _retry_config(raw: dict[str, Any]) -> RetryConfig:
-    retry = raw.get("retry") or {}
-    return RetryConfig(
-        max_retries=int(retry.get("max_retries", 2)),
-        retry_on_install=bool(retry.get("retry_on_install", True)),
-        retry_on_pipe=bool(retry.get("retry_on_pipe", True)),
-        retry_on_acp=bool(retry.get("retry_on_acp", True)),
-        retry_on_idle_timeout=bool(retry.get("retry_on_idle_timeout", True)),
-        retry_on_infra=bool(retry.get("retry_on_infra", True)),
-        retry_on_verifier_infra=bool(retry.get("retry_on_verifier_infra", True)),
-        wait_multiplier=float(retry.get("wait_multiplier", 2.0)),
-        min_wait_sec=float(retry.get("min_wait_sec", 1.0)),
-        max_wait_sec=float(retry.get("max_wait_sec", 30.0)),
-        exclude_categories=set(retry.get("exclude_categories", ["timeout"])),
-    )
+    # Centralized parsing: omitted fields fall back to RetryConfig's own
+    # defaults (which exclude provider_auth), not hard-coded literals, so a
+    # partial worker payload can't silently revert to retrying auth failures
+    # (#564 finding 2).
+    return RetryConfig.from_mapping(raw.get("retry"))
 
 
 def _environment_manifest(raw: dict[str, Any]):
--- a/src/benchflow/evaluation.py
+++ b/src/benchflow/evaluation.py
@@ -47,6 +47,7 @@ from benchflow._utils.scoring import (
     INFRA_ERROR,
     INSTALL_FAILED,
     PIPE_CLOSED,
+    PROVIDER_AUTH,
     VERIFIER_DEP_INSTALL,
     VERIFIER_INFRA,
     VERIFIER_TIMEOUT,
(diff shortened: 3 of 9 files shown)


bingran-you (changes requested):

This PR is currently not mergeable with `origin/main` (`src/benchflow/_utils/scoring.py` and `src/benchflow/evaluation.py` conflict), and the core auth-failure retry behavior is still not fixed.

The classifier only sees the top-level `result.error` and matches exact-case marker strings. In a real invalid-key Codex ACP run, the top-level error was just `ACP error -32603: Internal error`; the actual provider 401 details were only visible deeper in the provider trajectory. As a result, `RetryConfig().should_retry("ACP error -32603: Internal error")` still returned retryable and the eval retried 3 times.

I also checked representative strings locally; these auth-like failures are still misclassified as generic ACP errors:

```text
ACP error 401: unauthorized => acp_error
ACP error -32001: invalid api key => acp_error
ACP error -32603: failed to authenticate => acp_error
```

Please move the provider-auth classification closer to the provider/rollout boundary, or propagate a sanitized provider-auth diagnostic from telemetry/proxy/ACP into the retry classifier. More brittle top-level string matching will keep missing the real failure shape.

Additional repo-convention issue: the new regression test docstrings should name PR #564 / the guarding commit, not only the upstream issue.

Verification notes:
- Branch-head checks passed: `uv sync`, full `pytest`, `ty`, `ruff`, `git diff --check`.
- Healthy ACP smoke with `codex-acp` + `azure-foundry-openai/gpt-5.5` passed: reward `1.0`, `trajectory_source=acp`, usage `provider_response`, total tokens `25509`, `llm_trajectory.jsonl` present.
- Negative auth smoke with a fake OpenAI key produced 3 result files for one task, proving retries still occurred.

bingran-you:

Provider credential failures, especially Gemini `403 PERMISSION_DENIED` and Claude/OpenAI-style `401` auth errors, are currently classified as generic `acp_error`. Because `acp_error` is retryable, BenchFlow burns the full retry budget and starts fresh Daytona sandboxes even though the operator action is to fix/rotate credentials.

The proposed solution adds a new `provider_auth` error category, teaches `classify_error()` to detect auth markers, excludes `provider_auth` from `RetryConfig` by default, and adds a rollout-side `_provider_auth_status()` helper that scans captured provider HTTP exchanges for `401`/`403` and appends a sanitized marker to the ACP error.

### Blocking findings

1. **Daytona provider-auth failures are still classified before the sandbox proxy imports its captures.**

   `Rollout.run()` catches `AgentProtocolError` and immediately calls `_classify_acp_error()` at `src/benchflow/rollout.py:2065`. That helper calls `_provider_auth_status()` at `src/benchflow/rollout.py:2498`, which reads `server.trajectory.exchanges` at `src/benchflow/rollout.py:2509`.

   That works for the host-side proxy, where exchanges are appended in memory as requests complete. But Daytona uses `SandboxUsageProxy`, and that proxy does not populate the host-side `trajectory` until `SandboxUsageProxy.stop()` calls `_load_captures()` at `src/benchflow/providers/sandbox_usage_proxy.py:143`. `Rollout.cleanup()` only stops the usage runtime later, at `src/benchflow/rollout.py:1912`, after `_error` has already been classified and set.

   So for the real Daytona path, `_provider_auth_status()` can see an empty trajectory, return `None`, preserve the original generic `ACP error`, and the retry loop still sees `acp_error` and burns extra Daytona attempts.

   **Handoff:** move provider-auth enrichment to a point after sandbox usage captures have been imported, or make the usage runtime expose an explicit “flush/load captures now” operation that `_classify_acp_error()` can call before reading status codes. Keep this behind the usage/proxy runtime boundary rather than adding more ad hoc rollout branching. Add a regression where the fake sandbox proxy starts with empty `trajectory`, only populates during `stop()`, and the evaluation retry decision still sees `provider_auth`. Add a Daytona-shaped negative auth canary: one task, invalid provider key, exactly one rollout artifact, `error_category="provider_auth"`, no attempt 2.

2. **Retry defaults now drift at the worker payload boundary.**

   `RetryConfig` defaults now exclude `provider_auth` in `src/benchflow/evaluation.py:116`, but `src/benchflow/eval_worker.py:25` reconstructs retry config manually and falls back to `["timeout"]` at `src/benchflow/eval_worker.py:38` when `exclude_categories` is absent. Current sharded eval payloads serialize `exclude_categories`, but this is still a brittle contract boundary: direct worker payloads, older payloads, or partial payloads silently revert to retrying provider-auth failures.

   **Handoff:** centralize retry parsing, e.g. `RetryConfig.from_mapping()`, or default to `RetryConfig().exclude_categories` when the payload omits `exclude_categories`. Add a worker-config regression proving `_evaluation_config({... no retry.exclude_categories ...}).retry.should_retry(provider_auth_error)` is false.

### Residual risk

The core implementation relies on string markers and a sentinel suffix in `result.error`. A typed diagnostic emitted by the provider/usage proxy layer would be cleaner and less spaghetti-prone. Before merge, I would want integration checks for both Docker/host proxy and Daytona/sandbox proxy with invalid credentials, verifying attempt count, `result.error_category`, trajectory import, timing, and usage metadata shape.

ElegantLin:
Pushed the fix addressing your CHANGES_REQUESTED — the boundary-classification piece you flagged is now live on the branch (it had been committed locally but not pushed when you reviewed).

**Core finding (the real Daytona path):** the top-level `ACP error -32603: Internal error` genuinely lacks auth markers, and Daytona's `SandboxUsageProxy` only fills its trajectory on `stop()` — which runs in `cleanup()` *after* classification. So classification ran against an empty trajectory and the eval still retried.

Fix (`185d10a` + `06f34f8`):
- ACP-error classification is **deferred until after `cleanup()`**. `cleanup()` snapshots any provider 401/403 right after `stop_provider_runtime()` imports captures and before the runtime ref is dropped.
- The snapshot reads **only the integer status code** (never body/headers) and injects a sanitized `| provider auth failed (HTTP 401)` marker, which `classify_error` maps to `provider_auth` → not retryable.
- Works for both the host proxy (trajectory filled as requests complete) and the sandbox proxy (filled on `stop()`); regression `test_provider_auth_detection` proves a proxy that only populates on `stop()` still yields the 401.
- Also centralized retry parsing in `RetryConfig.from_mapping()` so a worker payload omitting `exclude_categories` still refuses to retry `provider_auth`.
- Removed a leftover unreachable `return None` in `_provider_auth_status`.

99 provider-auth/scoring/job/eval-worker tests pass; ruff + ty clean. Re-requested your review.

ElegantLin:
Follow-up (`2dbe90a`) — residual hardening after the boundary fix:
- run()'s `except AgentProtocolError` now sets a provisional `self._error` so `cleanup()`'s required-usage guard doesn't emit a misleading "no provider token usage captured" log on auth failures; the post-cleanup block still refines it to the `provider_auth` marker.
- `_classify_acp_error` reads `getattr(e, "message", str(e))` (base `AgentProtocolError` doesn't set `.message`).
- Documented the Bedrock / direct-AWS provider-auth coverage gap (`BedrockProxyServer` exposes no scannable 401/403).
Both new regression tests verified to fail without their fix.

bingran-you (approved):

Issue #250, opened 2026-05-11
ACP trials do not persist provider token/cost telemetry


Yiminnn:
## Summary

BenchFlow ACP trials can run agents and save rewards, timing, tool-call counts, errors, prompts, and ACP-visible trajectories, but they do not persist exact provider token/cost telemetry.

This means completed runs are usable for reward and trajectory analysis, but not for exact token/cost analysis unless external proxy/provider logs can be joined later.

## Current behavior

ACP trial artifacts include:

- `result.json`
- `timing.json`
- `prompts.json`
- `trajectory/acp_trajectory.jsonl`

`result.json` currently captures:

- task/trial metadata
- rewards
- agent/model metadata
- `n_tool_calls`
- `n_prompts`
- errors
- timestamps
- timing

It does not persist:

- input tokens
- output tokens
- cache tokens
- total tokens
- provider cost
- usage source
- price source

The ACP trajectory captures normalized session events such as tool calls, messages, thoughts, and stop reason. It is not a stable source of provider API `usage`.

## Prior related issues

- #90 already identified token counts and cost as missing Harbor-parity trial data, but the issue is closed.
  - https://github.com/benchflow-ai/benchflow/issues/90

- #183 identified that OpenHands CLI ACP works for execution but does not expose stable token/cost metrics.
  - https://github.com/benchflow-ai/benchflow/issues/183#issuecomment-4301874146

- #189 proposed using `TrajectoryProxy` to capture per-trial LLM token usage and write token counts into `result.json`.
  - https://github.com/benchflow-ai/benchflow/issues/189

## Root cause

BenchFlow currently treats ACP execution telemetry and provider LLM telemetry as the same artifact stream.

They are not the same:

- ACP gives BenchFlow agent/session events.
- Provider APIs return token usage at the LLM request/response layer.
- Without a proxy, SDK metrics bridge, or provider-log join, BenchFlow never sees reliable per-trial provider usage.

So this is not just a CSV/reporting issue. The runtime does not currently persist the required source data.

## Expected behavior

When provider usage is available, each trial should persist exact telemetry in `result.json`, for example:

```json
{
  "agent_result": {
    "n_input_tokens": 123,
    "n_output_tokens": 45,
    "n_cache_read_tokens": 0,
    "n_cache_creation_tokens": 0,
    "total_tokens": 168,
    "cost_usd": 0.0123,
    "usage_source": "provider_response",
    "price_source": "pricing_table_version"
  }
}
```

[...]

Issue #183, opened 2026-04-23
OpenHands integration: ACP CLI path does not expose stable token/cost metrics; need SDK-runner or equivalent metrics bridge


AmyTao:
## Summary

Benchflow's current `openhands` integration uses the OpenHands **CLI ACP** path (`openhands acp ...`). This works for install / launch / ACP execution, but it does **not** reliably expose the token / cost metrics that the older Harbor + OpenHands SDK runner path provided.

This becomes a problem if Benchflow is expected to replace Harbor in the OpenHands evaluation harness while preserving metrics such as:

- `agent_result.n_input_tokens`
- `agent_result.n_cache_tokens`
- `agent_result.n_output_tokens`
- `agent_result.cost_usd`

## Current behavior

Benchflow can run OpenHands through ACP successfully, but the ACP path does not provide a stable source of cumulative LLM metrics.

What we can get reliably from the current ACP session:

- agent name / version from `initialize`
- tool calls
- message / thought / tool trajectory
- stop reason
- timing

What we **cannot** rely on the ACP path to provide today:

- cumulative prompt tokens
- cumulative completion tokens
- cumulative cache tokens
- cumulative cost in USD

Benchflow now has result schema slots for these metrics, but under the current OpenHands CLI ACP integration they will usually remain `null` unless some future trajectory source includes usage/cost.

## Why this matters

The older Harbor/OpenHands SDK-runner path could collect these values directly inside the agent process via:

- `llm.metrics.accumulated_token_usage`
- `llm.metrics.accumulated_cost`

That means Harbor output could populate richer trial results, while the current Benchflow ACP integration cannot provide equivalent metrics with the same reliability.

## Expected behavior

Benchflow should have a supported way to run OpenHands that can emit stable metrics comparable to the old Harbor/OpenHands SDK-runner results.

At minimum, OpenHands runs in Benchflow should be able to populate:

```json
"agent_result": {
  "n_input_tokens": ...,
  "n_cache_tokens": ...,
  "n_output_tokens": ...,
  "cost_usd": ...
}
```

Suggested direction
Add an OpenHands SDK runner execution path in Benchflow, and collect metrics directly from the SDK runtime.
Or add an equivalent telemetry/metrics bridge for the CLI ACP path, if OpenHands CLI can emit cumulative usage/cost in a supported machine-readable way.
From an accuracy and maintenance perspective, the SDK-runner path appears preferable, since it reads metrics from the same execution layer that is actually calling the LLM.

[...]

PR #566, opened 2026-05-26
feat: stream ACP trajectory incrementally to disk

Yiminnn:
## Summary

`acp_trajectory.jsonl` is now written to disk **as the agent runs**, not just at end-of-run. Mid-run crashes leave a usable partial trajectory; followers can `cat` the file during the run to see live tool calls.

**Format and viewer are unchanged.** Each line is still a merged event from `_capture_session_trajectory`; the writer just snapshots the current cumulative state on every ACP update.

## Mechanism

- `ACPSession` gains an optional `on_change: Callable[[ACPSession], None]` invoked after `record_user_prompt`, `mark_prompt_end`, and any *recognized* `handle_update` branch.
- `TrajectoryWriter` (new) writes payloads atomically via `tmp + os.replace`; dedupes byte-identical re-snapshots.
- `_snapshot_session_trajectory` is non-destructive (does not flush `_pending_text`), so streamed chunks stay mergeable until `mark_prompt_end` runs and produces the canonical merged events.
- `Rollout._attach_trajectory_writer` wires the writer in `connect()` and `connect_as()` (multi-role), capturing the cumulative `self._trajectory` from prior scenes via `make_trajectory_sink(writer, prior)` — so scene-N's sink writes `prior_scenes + current_session`, not just the current session.
- Final write in `_build_rollout_result` now goes through `TrajectoryWriter.write_final` (idempotent atomic write — important for oracle / scraped-fallback paths where no live writer ran).

## Why this is safe

- **Backward compatible**: same JSONL output, same path, same consumers (`viewer.py`, `judge.py.tmpl`, `tests/test_smoke.py`, etc.).
- **Multi-scene**: cumulative-aware sink prevents earlier scenes from being wiped by the next scene's writer. New tests in `TestMultiSceneCumulativeStreaming` cover this.
- **Async**: `handle_update` is single-thread cooperative within the asyncio event loop, so the sync `write_text + os.replace` per update is consistent with the existing model. See follow-up note below for one MEDIUM finding deferred to a separate PR.
- **Crash recovery**: stale `.tmp` from a SIGKILLed writer is swept by the next `TrajectoryWriter.__init__`.
- **Unknown ACP updates**: `handle_update` skips `_notify_change` for unrecognized `sessionUpdate` types (no wasted snapshots, no risk if a future ACP version adds new event kinds).
- **Callback errors are visible**: `_notify_change` now logs at `ERROR` (was `WARNING`) — silent streaming degradation in a 64-concurrency log would otherwise be easy to miss.

## End-to-end verified

`openhands` + `gemini/gemini-3.1-flash-lite` on SkillsBench `citation-check` via Daytona — reward 1.0, no errors, 2 tool calls, `trajectory_source: acp`, `partial_trajectory: False`. File lands at t=0 with the `user_message` and grows through each tool-call status transition:

| t | bytes | event | what |
| --- | --- | --- | --- |
| 10:04:20 | 826 | 1 | `user_message` |
| 10:04:24 | 52,750 | 3 | tool `cat` completed |
| 10:04:48 | 82,612 | 5 | tool `edit` completed |
| 10:04:52 | 85,022 | 7 | final `agent_message` |

On `main` the file would have first appeared at the verifier-finished mark (~10:05:53 in this run).

## Test plan

- [x] 7 new unit tests in `tests/test_capture_trajectory.py`:
  - `TestTrajectoryWriter` — file appears on first event, tool_call pending → completed transitions visible, exceptions swallowed, atomic rewrites (no torn lines / leftover `.tmp`), `write_final` overwrites.
  - `TestSnapshotSessionTrajectory` — non-destructive snapshot preserves `_pending_text` until `mark_prompt_end`.
  - `TestMultiSceneCumulativeStreaming` — prior scenes preserved on disk during current scene; empty current session doesn't wipe history; caller-mutated cumulative list does not double-count.
  - `TestHandleUpdateUnknownType` — unknown `sessionUpdate` skips `on_change`; known types still notify.
  - `TestTrajectoryWriterStaleTmpCleanup` — stale `.tmp` swept on `__init__`.
- [x] Full local test suite green: **1312 passed, 9 skipped**, no regressions.
- [x] Live Daytona run on `openhands + gemini/gemini-3.1-flash-lite + citation-check`: reward 1.0.

## Known follow-ups (out of scope for this PR)

- **Event-loop blocking (MEDIUM)**: `TrajectoryWriter.flush` performs sync `write_text + os.replace` from the asyncio event loop (~1–3 ms per ACP update; ~100–300 ms cumulative per chunky prompt with 100+ streamed chunks). Bounded but not free. Would require making the `on_change` callback async-aware. Defer to a follow-up unless a real run shows measurable cost.
- **Token usage** is unaffected and unrelated: `openhands + daytona` reports `usage_source: unavailable` by design (the host-side proxy is unreachable from the remote sandbox). Tracked in #183 / #249 / #250.


bingran-you:

This PR is trying to make `trajectory/acp_trajectory.jsonl` useful before rollout finish. Today ACP trajectory is only written during result finalization, so a crash or timeout can leave followers with no readable ACP trail. The proposed solution adds an `ACPSession.on_change` callback, wires it from `Rollout.connect()` / `connect_as()`, and introduces `TrajectoryWriter` to atomically rewrite cumulative JSONL snapshots as ACP updates arrive. Final result writing is routed through the same writer.

The direction is good, but I do not think this is merge-ready yet.

### Blocking findings

1. **Finalization can overwrite the only partial trajectory from a failed later scene.**

   `_attach_trajectory_writer()` correctly streams `prior + current session` to disk, but the in-memory rollout trajectory is only extended after `execute_prompts()` returns successfully. If a multi-scene or multi-role rollout has completed one scene, then a later scene times out/crashes after emitting ACP events, the live writer can briefly put the right partial cumulative trajectory on disk. During cleanup, `_capture_partial_acp_trajectory()` immediately returns because `self._trajectory` is already non-empty, so the current failed session is never appended. Then `_build_rollout_result()` unconditionally calls `write_final(self._trajectory)`, overwriting the live file with the stale prior-scene-only trajectory.

   Relevant areas: `src/benchflow/rollout.py:536`, `:1506`, `:1576`, `:1637`.

   **Repro shape:** scene 1 succeeds, scene 2 connects and streams `user_message` / `tool_call`, `execute_prompts()` raises before returning, `disconnect()` skips partial capture because prior trajectory exists, and final result overwrites `acp_trajectory.jsonl` with only scene 1.

   **Handoff:** make partial capture append the current session delta whenever there is a live session, not only when `self._trajectory` is empty. Set `partial_trajectory=True` / `trajectory_source="partial_acp"` for that mixed prior+partial result. Add a rollout-level regression where scene 1 completes, scene 2 emits ACP events, `execute_prompts()` raises, and the final `acp_trajectory.jsonl` plus result metadata preserve both scenes.

2. **The new tests push `tests/test_capture_trajectory.py` past 1k lines.**

   This PR grows `tests/test_capture_trajectory.py` from roughly 735 lines to over 1k by appending writer, snapshot, multi-scene streaming, unknown update, and stale tmp tests into the existing capture test module. This is exactly the file-sprawl threshold this review bar tries to stop.

   **Handoff:** split the new classes into a focused module such as `tests/test_trajectory_writer.py` or `tests/test_trajectory_streaming.py`, and keep `test_capture_trajectory.py` scoped to capture semantics.

3. **`_snapshot_session_trajectory()` duplicates the canonical event-to-JSONL conversion.**

   The PR now has two near-identical implementations of the event serialization contract: `_snapshot_session_trajectory()` and `_capture_session_trajectory()` (`src/benchflow/trajectories/_capture.py:76`, `:167`). The safety claim is that the live file format is unchanged, so duplicating the schema mapping is fragile.

   **Handoff:** extract the shared “session events to trajectory events” conversion into one helper, with the only behavioral difference being whether pending text is flushed destructively or merged non-destructively. That makes “streaming format equals final format” a structural invariant instead of copy/paste discipline.

### Experiment-health gaps

The PR description reports one Daytona smoke run. Useful, but not enough for a trajectory artifact lifecycle change. Before relying on this for SkillsBench production data, I would want Docker/Daytona parity, with-skill/no-skill coverage, timeout/crash behavior including the later-scene failure case above, verifier lifecycle proof, and metadata completeness (`trajectory_source`, `partial_trajectory`, timing, token/usage where available, tool-call counts).

Verdict: request changes. The core idea is solid, but the current rollout finalization path can erase the partial trajectory this PR is meant to preserve.

Yiminnn:
Thanks for the thorough review. Pushed `1f91d1b` addressing all three blocking findings.

**Finding 1 — late-scene partial capture overwrite.** Confirmed and fixed. `_capture_partial_acp_trajectory` no longer early-returns when `self._trajectory` is non-empty; it now captures the live session and appends the delta beyond `_session_traj_count` (the pointer to events already extended by `execute()`). When a delta is appended, `partial_trajectory=True` and `trajectory_source='partial_acp'` are set. New regression test class `TestMultiScenePartialCaptureFix` in `tests/test_trajectory_streaming.py` covers four cases: late-scene delta appended on top of completed scene 1, no-delta no-op when execute already drained, single-scene crash still works, and missing live session is a safe no-op.

**Finding 2 — duplicated event→JSONL conversion.** Extracted `_events_to_trajectory(events)` in `src/benchflow/trajectories/_capture.py`. Both `_capture_session_trajectory` (destructive, flushes pending) and `_snapshot_session_trajectory` (non-destructive, merges pending without mutating) now delegate to it. Streaming-format == final-format is a structural invariant.

**Finding 3 — `test_capture_trajectory.py` over 1k lines.** Moved `TestSnapshotSessionTrajectory`, `TestTrajectoryWriter`, `TestMultiSceneCumulativeStreaming`, `TestHandleUpdateUnknownType`, `TestTrajectoryWriterStaleTmpCleanup`, and the new `TestMultiScenePartialCaptureFix` into `tests/test_trajectory_streaming.py`. `test_capture_trajectory.py` is back to capture semantics only (743 → 743 lines for the capture half, 455 lines for the streaming half).

**Experiment-health gaps.** I'll keep this PR scoped to the trajectory write-time change. Docker/Daytona parity, with/without-skill coverage, token-usage capture across environments, and the full crash-matrix evidence are tracked separately (token capture on `openhands + daytona` is the existing #183 / #249 / #250 gap, not something this PR introduces or fixes). Happy to bring receipts on those in a follow-up PR if you'd like a specific matrix.

Local: 2379 passed, 11 skipped, no regressions.

PR #572, opened 2026-05-27
fix: surface test-stdout.txt in verifier exception for dep_install classification (#540)

### LINKED ISSUE #540: ENG-151 verifier_error_category=dep_install classification cannot fire in production — classifier markers never match the production emitter
## Summary

The ENG-151 \"verifier dependency install classification\" path (`verifier_error_category == \"dep_install\"`) cannot fire on real evals: the only error string the production code emits for verifier dep failures matches **none** of the four markers the classifier looks for. The §16 unit tests pass because they hand-feed marker strings into `classify_verifier_error`; the production runtime never produces those strings.

A real torch dep-install failure in §6's recommended task (`simpo-code-reproduction`) is observed in `verifier/test-stdout.txt`, but `result.json.verifier_error_category` comes back `null` (or `\"verifier_other\"` if the verifier raises) — never `\"dep_install\"`.

## The disconnect

[`src/benchflow/_utils/scoring.py:85-97`](https://github.com/benchflow-ai/benchflow/blob/v0.5-integration/src/benchflow/_utils/scoring.py#L85-L97) classifier:
```python
def _looks_like_verifier_dep_install_error(error: str) -> bool:
    markers = (
        \"dependency install failed\",
        \"no solution found\",
        \"could not find a version\",
        \"resolution impossible\",
    )
    return any(marker in error for marker in markers)
```

[`src/benchflow/task/verifier.py:293-298`](https://github.com/benchflow-ai/benchflow/blob/v0.5-integration/src/benchflow/task/verifier.py#L293-L298) — the ONLY production emitter for verifier-side failures (when the reward file is missing):
```python
raise RewardFileNotFoundError(
    f\"verifier exited with rc={test_return_code}; no reward file \"
    f\"found at {...}\"
)
```

`\"verifier exited with rc=…\"` matches none of `_looks_like_verifier_dep_install_error`'s markers. The classifier never sees pip's actual `\"no solution found when resolving tool dependencies: torch==2.1.2+cpu\"` line because pip's output goes to `verifier/test-stdout.txt`, not into the exception message.

## Repro

```bash
export TASKS=$(pwd)/.cache/datasets/benchflow-ai/skillsbench/tasks

uv run bench eval create --tasks-dir $TASKS \\
  --include simpo-code-reproduction \\
  --agent gemini --model gemini-2.5-flash --sandbox daytona \\
  --jobs-dir /tmp/test-eng151

J=$(ls -d /tmp/test-eng151/20*/simpo-code-reproduction__*/ | head -1)

# Verifier actually saw a dep failure (confirmed in stdout):
grep \"no solution found\" $J/verifier/test-stdout.txt
# × No solution found when resolving tool dependencies: torch==2.1.2+cpu

# But the classifier didn't fire:
jq '.verifier_error_category' $J/result.json
# null  (or \"verifier_other\" if the verifier raised)
```

## Why §16 unit tests pass anyway

`tests/test_acp.py::*verifier_dep_install*` constructs `classify_verifier_error(\"... no solution found ...\")` directly with a marker-bearing string. The classifier function is exercised, but the *production code path* that should reach it (verifier failure → exception message → classifier) is never end-to-end tested. The function works; the wiring doesn't.

## Severity

P1.

- A named v0.5 diagnostic field (`verifier_error_category == \"dep_install\"`) is unreachable on real evals.
- The §6 live recipe (`simpo-code-reproduction`) is documented as the canonical exercise — observed to silently *not* trigger the classification.
- Operators triaging \"verifier failed\" can't filter for dep-install failures using the documented field.

## Proposed fix

Pick one (recommend #1 + #3):

1. **Surface pip's output into the exception.** In `task/verifier.py:293`, when raising `RewardFileNotFoundError`, also tail `verifier/test-stdout.txt` (or `test-stderr.txt`) for the last N lines and include them in the exception message. Then `classify_verifier_error` sees pip's `\"no solution found\"` and fires correctly.

2. **Detect dep-install failures at the verifier wrapper layer.** Before raising `RewardFileNotFoundError`, scan `test-stdout.txt` for the same markers and set a flag on the exception (`dep_install_detected = True`). Classifier reads the flag, not just the message.

3. **End-to-end test.** Add an integration test (Docker is fine — doesn't need Daytona) where the verifier runs `pip install nonexistent-package==1.0` and the test asserts `result.json.verifier_error_category == \"dep_install\"`. Today no such test exists.

## Related side-finding (`simpo-code-reproduction` itself)

The recommended task at §6 — `simpo-code-reproduction` — has two task-level problems independent of this bug. Useful context but not in benchflow's scope:

- `tests/test.sh` swallows the test return code via `if [ $? -eq 0 ]; … else echo 0 > reward.txt; fi` (no `set -e`). Even on a real dep failure, `reward.txt = 0` is written, the verifier exits 0, no exception is raised, and the rollout finishes \"successfully failed\". Probably a skillsbench task bug ([benchflow-ai/skillsbench](https://github.com/benchflow-ai/skillsbench)).
- `torch==2.1.2+cpu` is no longer resolvable from configured indexes (`× No solution found when resolving tool dependencies: torch==2.1.2+cpu`). Task perma-broken; reward=0 is now infra noise, not capability signal. Probably a skillsbench task bug.

These together mean §6 has no way to exercise the live `dep_install` path even if the benchflow classification bug above were fixed. A separate \"healthy dep-install-failing task\" should be selected (or constructed) for §6's recipe.

## Found via

v0.5 e2e validation §6 at `5a7a326`. 500s eval; observed `result.json.verifier_error_category=null` despite confirmed pip dep failure in stdout.

ElegantLin:
## Summary

- **Root cause:** `RewardFileNotFoundError` in `verifier.py` only included file paths in its message, never pip's actual error output. The `classify_verifier_error` function had `dep_install` markers (`"no solution found"`, etc.) but could never match them because they lived in `test-stdout.txt`, not the exception.
- **Fix:** When raising `RewardFileNotFoundError`, tail the last 30 lines of `test-stdout.txt` into the exception message. This lets `classify_verifier_error` see pip's output and correctly return `verifier_dep_install`.
- **No changes to `scoring.py`** — the classifier logic and markers were already correct (added in prior PRs); only the wiring was broken.

## Test plan

- [x] 4 new parametrized cases in `test_classify_verifier_error` covering all dep_install markers in exception messages
- [x] `test_dep_install_takes_precedence_over_generic_crash` — precedence test
- [x] `TestRewardFileNotFoundSurfacesStdout` — 4 end-to-end wiring tests: stdout tail present, classifier fires, missing file graceful, non-dep-install stays `verifier_failure`
- [x] All 96 verifier+scoring tests pass
- [x] ruff clean

Closes #540

---

<a href="https://app.devin.ai/review/benchflow-ai/benchflow/pull/572" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/benchflow/_utils/scoring.py
+++ b/src/benchflow/_utils/scoring.py
@@ -18,6 +18,16 @@ VERIFIER_INFRA = "verifier_infra"
 VERIFIER_TIMEOUT = "verifier_timeout"
 VERIFIER_DEP_INSTALL = "verifier_dep_install"
 
+# Canonical dependency-install markers shared by verifier stdout scanning and
+# verifier-error classification. Keep these lower-case; the helper below
+# performs case-insensitive matching.
+VERIFIER_DEP_INSTALL_MARKERS: tuple[str, ...] = (
+    "dependency install failed",
+    "no solution found",
+    "could not find a version",
+    "resolution impossible",
+)
+
 ScoreOutcome = Literal["passed", "failed", "errored", "verifier_errored"]
 ResultOutcome = Literal["passed", "failed", "errored", "verifier_errored", "unscored"]
 
@@ -77,7 +87,7 @@ def classify_verifier_error(verifier_error: str | None) -> str | None:
         return None
     lower = verifier_error.lower()
     if "verifier crashed" in verifier_error:
-        if _looks_like_verifier_dep_install_error(lower):
+        if contains_verifier_dep_install_marker(lower):
             return VERIFIER_DEP_INSTALL
         if _looks_like_verifier_infra_error(lower):
             return VERIFIER_INFRA
@@ -87,15 +97,15 @@ def classify_verifier_error(verifier_error: str | None) -> str | None:
     return "verifier_other"
 
 
-def _looks_like_verifier_dep_install_error(error: str) -> bool:
+def contains_verifier_dep_install_marker(text: str) -> bool:
     """Detect verifier dependency installation failures (ENG-151)."""
-    markers = (
-        "dependency install failed",
-        "no solution found",
-        "could not find a version",
-        "resolution impossible",
-    )
-    return any(marker in error for marker in markers)
+    lower = text.lower()
+    return any(marker in lower for marker in VERIFIER_DEP_INSTALL_MARKERS)
+
+
+def _looks_like_verifier_dep_install_error(error: str) -> bool:
+    """Backward-compatible internal alias for dep-install marker matching."""
+    return contains_verifier_dep_install_marker(error)
 
 
 def _looks_like_verifier_infra_error(error: str) -> bool:
--- a/src/benchflow/task/verifier.py
+++ b/src/benchflow/task/verifier.py
@@ -13,8 +13,10 @@ from __future__ import annotations
 
 import json
 import logging
+import re
 import shlex
 import shutil
+from collections import deque
 from pathlib import Path
 from typing import Any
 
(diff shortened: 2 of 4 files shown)


bingran-you (changes requested):

The dependency-install classification improvement is useful, and the real Claude ACP smoke passed, but the current implementation surfaces raw verifier stdout through `RewardFileNotFoundError` / `verifier_error`. That widens the secret-leak blast radius because `test-stdout.txt` can contain env dumps, install URLs, tokens, or other untrusted subprocess output that later lands in summaries/dashboards.

Please redact before appending stdout tails, or keep raw stdout out of result metadata and classify from a private diagnostic path instead.

Additional maintainability issues:
- `_tail_file()` reads/splits the whole file to return 30 lines. Please stream-tail it, e.g. with `collections.deque(f, maxlen=n)`.
- `tests/test_verify.py` grows past 1k lines. Please move these regression tests to a focused verifier-output module.
- Regression test docstrings should name the reviewed PR/fix explicitly, e.g. PR #572 / #540, per repo convention.

Verification notes:
- Branch-head checks passed: `uv sync`, full `pytest`, `ty`, `ruff`.
- Current-merge `pytest` and `ty` passed.
- Current-merge `ruff` failed only on existing main baseline `UP041` in `src/benchflow/sandbox/docker.py:408`, not this PR.
- Real `claude-agent-acp` Docker smoke passed: reward `1.0`, `trajectory_source=acp`, 5 ACP events, 1 tool call, usage `provider_response`, total tokens `64349`, `llm_trajectory.jsonl` present.

bingran-you:

PR #572 fixes a verifier classification gap from #540: dependency-install failures were visible in `test-stdout.txt`, but `RewardFileNotFoundError` only included reward-file paths. Because `classify_verifier_error()` only sees the persisted `verifier_error` string, dep-install markers like “no solution found” never reached the classifier in real runs.

The proposed solution tails the last 30 lines of `test-stdout.txt`, redacts known credential patterns, appends that tail to `RewardFileNotFoundError`, and adds tests for classifier behavior, tailing, and redaction. It also moves new tests into `tests/test_verifier_output.py`, keeping `tests/test_verify.py` under 1k lines.

### Blocking findings

1. **Redaction is still too narrow for surfacing verifier stdout into persisted metadata.**

   `src/benchflow/task/verifier.py:73` defines an allowlist of specific token regexes, `src/benchflow/task/verifier.py:117` applies only those patterns, and `src/benchflow/task/verifier.py:361` appends the resulting text into `RewardFileNotFoundError`, which becomes `verifier_error` and can land in summaries/dashboards. This can still leak common secret-bearing stdout lines such as `HF_TOKEN=...`, `GITHUB_TOKEN=...`, `AWS_SECRET_ACCESS_KEY=...`, and credentialed install URLs like `PIP_INDEX_URL=https://user:password@...`.

   **Handoff:** either avoid persisting raw stdout and extract only a safe dep-install diagnostic/category from the private verifier log, or move redaction into a canonical utility that redacts by secret-looking key names (`KEY`, `TOKEN`, `SECRET`, `PASSWORD`, `AUTH`, `COOKIE`, etc.), URL credentials, headers, and known provider token formats. Add regression tests for env-dump lines and credentialed package-index URLs.

2. **The new tests do not guard the production wiring this PR depends on.**

   `tests/test_verifier_output.py:49` tests classifier strings directly, `:70` tests `_tail_file()` directly, and `:140` synthesizes the final `verifier_error` string. None exercises `Verifier._verify_test_script()` at `src/benchflow/task/verifier.py:349`, where stdout tail is actually attached, nor the `rollout.py` wrapper that prefixes `verifier crashed:` before classification. These tests would still pass if the real verifier stopped appending stdout to the exception.

   **Handoff:** add an async verifier test with a fake mounted sandbox that writes `rollout_paths.test_stdout_path`, produces no reward file, raises `RewardFileNotFoundError`, and asserts the raised message is redacted and classifies as `VERIFIER_DEP_INSTALL` after the real `verifier crashed: {e}` wrapping. Ideally add the same for the non-mounted/download path, since experiment health depends on Docker/Daytona artifact parity.

3. **The “30 line” tail has no byte cap, so one long line can bloat `result.json` and dashboards.**

   `src/benchflow/task/verifier.py:105` streams by line with `deque(maxlen=n)`, but `src/benchflow/task/verifier.py:117` joins retained lines without a character/byte limit. A verifier can emit a single huge line, causing the full line to be persisted through `verifier_error` at `src/benchflow/task/verifier.py:361`.

   **Handoff:** cap the surfaced diagnostic by bytes/chars after redaction, include a clear truncation marker, and test the single-long-line case. The cleaner version is still to persist only safe extracted dep-install evidence instead of raw tail context.

Verdict: request changes. The implementation is much closer, but I would not merge while untrusted verifier stdout can still leak broad classes of secrets into persisted experiment metadata.

ElegantLin:
Follow-up (`65f116f`): beyond closing the secret-leak (now surfaces only a fixed diagnostic), the marker scan was only the last 30 lines — but dep-install runs at the START of test.sh, so a marker followed by >30 lines was silently dropped → misclassified as generic `verifier_failure`. Now scans the whole file in bounded 64K chunks (with overlap for boundary-straddling markers), still returning only a bool (no raw stdout leaves the function). Regression test covers the early-marker + long-trailing case.

bingran-you (approved):

PR #573, opened 2026-05-27
fix: isolate mirror tests from host env to prevent API key leak (#522)

### LINKED ISSUE #522: test_google_only_mirrors_to_gemini leaks GEMINI_API_KEY in assertion output and breaks §15 regression flow
## Summary

`tests/test_agent_env_resolution.py::TestAutoInheritEnvBidirectionalMirror::test_google_only_mirrors_to_gemini` is not isolated from the host process environment, with two compounding consequences:

1. **Guide-execution blocker.** Anyone who follows [`docs/v05-e2e-testing-guide.md`](https://github.com/benchflow-ai/benchflow/blob/v0.5-integration/docs/v05-e2e-testing-guide.md) is told to `export GEMINI_API_KEY=[REDACTED]` in §0 and then to run `uv run python -m pytest tests/ -x -q` in §15 expecting `1910+ passed, 0 failed`. With the host `GEMINI_API_KEY` exported, this test fails — making the documented regression baseline unachievable.

2. **Secret leaks into the pytest output.** The failing assertion is `assert env[\"GEMINI_API_KEY\"] == \"gk-test\"`, and pytest's failure rendering prints both operands. The real key value ends up verbatim in stdout (and any captured CI artifact or pasted bug report).

## Repro

```bash
git checkout v0.5-integration  # at 5a7a326
uv sync --extra dev --extra sandbox-daytona --locked
export GEMINI_API_KEY=[REDACTED]    # any non-empty value
uv run python -m pytest tests/test_agent_env_resolution.py::TestAutoInheritEnvBidirectionalMirror -v
```

Result:
\`\`\`
___ TestAutoInheritEnvBidirectionalMirror.test_google_only_mirrors_to_gemini ___
    env = {\"GOOGLE_API_KEY\": \"gk-test\"}
    auto_inherit_env(env)
    assert env[\"GOOGLE_API_KEY\"] == \"gk-test\"
>   assert env[\"GEMINI_API_KEY\"] == \"gk-test\"
E   AssertionError: assert '[REDACTED_GEMINI_API_KEY]...j35F3w_691mJo' == 'gk-test'
E       + [REDACTED_GEMINI_API_KEY]
\`\`\`

The redacted-style `'[REDACTED_GEMINI_API_KEY]...j35F3w_691mJo'` and the expanded full-string diff line both contain the real key. Pytest does not redact values for assertion failures.

## Root cause

[`src/benchflow/agents/env.py:66-110`](https://github.com/benchflow-ai/benchflow/blob/v0.5-integration/src/benchflow/agents/env.py#L66-L110) — `auto_inherit_env(agent_env)` defaults `source_env` to `os.environ` and `setdefault`s every known auth var (including `GEMINI_API_KEY`) into `agent_env` *before* the GOOGLE↔GEMINI mirror runs. The mirror at L112-115 then sees `GEMINI_API_KEY` already present and skips, so the host value wins.

`tests/test_agent_env_resolution.py::TestAutoInheritEnvBidirectionalMirror` (lines 47-68) declares three tests that each call bare `auto_inherit_env(env)` without `monkeypatch.delenv` or a `source_env={}` override. The sibling class `TestResolveAgentEnvGemini` (line 71+) consistently uses `_clear_keys(monkeypatch)` — so the hygiene exists, it just wasn't applied here.

Test \"happens to pass\" today only on hosts without `GEMINI_API_KEY`/`GOOGLE_API_KEY` in os.environ — which is everywhere except a developer following the v0.5 e2e guide.

## Severity

P0.

- Release-blocking secret exposure: every documented v0.5 verifier following §0 + §15 trips this.
- Secret-handling: a failing test should never reveal an exported credential. If this had been run in CI with a real key, it would be in the runs log.

## Proposed fixes (pick one or stack)

1. **Test-isolation (minimum).** Add an autouse fixture (or per-test `monkeypatch.delenv` like `_clear_keys`) on `TestAutoInheritEnvBidirectionalMirror` so the three tests start from a clean process env. Trivial.

2. **API discipline (better).** Make `auto_inherit_env`'s `source_env` a required arg — no implicit `os.environ` fallback — and have the one production call site pass `os.environ` explicitly. This eliminates the entire \"test forgot to isolate\" class of bug at the type signature.

3. **Secret-safe assertion (defense in depth).** Replace `assert env[\"GEMINI_API_KEY\"] == \"gk-test\"` with `assert env.get(\"GEMINI_API_KEY\") == \"gk-test\", \"mirror failed\"` and gate the failing-value message so real-looking keys are scrubbed in `conftest.py`. Even with #1 or #2, a future test could regress this.

(3) alone isn't enough — the assertion would still need a non-leaky failure render. Recommend (1)+(2) at minimum.

## Found via

v0.5-integration @ `5a7a326` — automated v0.5 e2e validation run following `docs/v05-e2e-testing-guide.md`.

ElegantLin:
## Summary

- `TestAutoInheritEnvBidirectionalMirror` called `auto_inherit_env()` without clearing `os.environ`, so real API keys (e.g. `GEMINI_API_KEY`) on the host leaked into the test dict
- Assertion failures then printed the real key value in pytest output — a secret exposure risk
- Added an `autouse` fixture using the existing `_clear_keys()` helper, matching the isolation pattern already used by `TestResolveAgentEnvGemini` in the same file

## Test plan

- [x] All 7 tests in `test_agent_env_resolution.py` pass (including the previously failing `test_gemini_only_mirrors_to_google`)
- [x] Verified the fix works with real `GEMINI_API_KEY` / `GOOGLE_API_KEY` exported in the environment
- [x] ruff clean

Closes #522

---

<a href="https://app.devin.ai/review/benchflow-ai/benchflow/pull/573" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/tests/test_agent_env_resolution.py
+++ b/tests/test_agent_env_resolution.py
@@ -45,11 +45,18 @@ def _clear_keys(monkeypatch):
 
 
 class TestAutoInheritEnvBidirectionalMirror:
-    """auto_inherit_env mirrors GEMINI_API_KEY ↔ GOOGLE_API_KEY both ways."""
+    """auto_inherit_env mirrors GEMINI_API_KEY ↔ GOOGLE_API_KEY both ways.
+
+    Pure mirror tests pass ``source_env={}`` to cut off host-env inheritance
+    entirely. Depending on (or partially clearing) the real process env would
+    let a host ``GEMINI_API_KEY``/``GOOGLE_API_KEY`` bleed into the assertion —
+    and a failing assert would print that real credential. ``source_env={}``
+    makes the test hermetic by construction.
+    """
 
     def test_gemini_only_mirrors_to_google(self):
         env = {"GEMINI_API_KEY": "gk-test"}
-        auto_inherit_env(env)
+        auto_inherit_env(env, source_env={})
         assert env["GEMINI_API_KEY"] == "gk-test"
         assert env["GOOGLE_API_KEY"] == "gk-test"
 
@@ -57,13 +64,13 @@ class TestAutoInheritEnvBidirectionalMirror:
         """Issue #342: setting GOOGLE_API_KEY alone must populate GEMINI_API_KEY
         so the in-sandbox Gemini CLI can authenticate."""
         env = {"GOOGLE_API_KEY": "gk-test"}
-        auto_inherit_env(env)
+        auto_inherit_env(env, source_env={})
         assert env["GOOGLE_API_KEY"] == "gk-test"
         assert env["GEMINI_API_KEY"] == "gk-test"
 
     def test_both_set_preserves_explicit_values(self):
         env = {"GEMINI_API_KEY": "gemini-val", "GOOGLE_API_KEY": "google-val"}
-        auto_inherit_env(env)
+        auto_inherit_env(env, source_env={})
         assert env["GEMINI_API_KEY"] == "gemini-val"
         assert env["GOOGLE_API_KEY"] == "google-val"
 
--- a/tests/test_resolve_env_helpers.py
+++ b/tests/test_resolve_env_helpers.py
@@ -47,117 +47,115 @@ class TestAutoInheritEnv:
             ),
         ],
     )
-    def test_inherits_key(self, monkeypatch, env_name, env_value):
-        monkeypatch.setenv(env_name, env_value)
+    def test_inherits_key(self, env_name, env_value):
         env = {}
-        auto_inherit_env(env)
+        auto_inherit_env(env, source_env={env_name: env_value})
         assert env[env_name] == env_value
 
-    def test_does_not_overwrite_explicit(self, monkeypatch):
-        monkeypatch.setenv("ANTHROPIC_API_KEY", "sk-host")
+    def test_does_not_overwrite_explicit(self):
         env = {"ANTHROPIC_API_KEY": "sk-explicit"}
-        auto_inherit_env(env)
+        auto_inherit_env(env, source_env={"ANTHROPIC_API_KEY": "sk-host"})
         assert env["ANTHROPIC_API_KEY"] == "sk-explicit"
 
     def test_gemini_mirrored_to_google(self):
         env = {"GEMINI_API_KEY": "gk-test"}
-        auto_inherit_env(env)
+        auto_inherit_env(env, source_env={})
         assert env["GOOGLE_API_KEY"] == "gk-test"
 
     def test_gemini_mirror_no_overwrite(self):
         env = {"GEMINI_API_KEY": "gk-test", "GOOGLE_API_KEY": "gk-explicit"}
-        auto_inherit_env(env)
+        auto_inherit_env(env, source_env={})
         assert env["GOOGLE_API_KEY"] == "gk-explicit"
 
-    def test_missing_host_key_not_added(self, monkeypatch):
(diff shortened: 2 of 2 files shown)


bingran-you (changes requested):

The branch fixes one mirror-test class, but the same host-env leak still exists in sibling tests in `tests/test_resolve_env_helpers.py`.

I reproduced both leaks locally:

```bash
GOOGLE_API_KEY=host-dummy-secret BENCHFLOW_DOTENV_PATH=/tmp/no-such-benchflow-env uv run python -m pytest tests/test_resolve_env_helpers.py::TestAutoInheritEnv::test_gemini_mirrored_to_google -q --tb=short
AZURE_RESOURCE=host-dummy-resource BENCHFLOW_DOTENV_PATH=/tmp/no-such-benchflow-env uv run python -m pytest tests/test_resolve_env_helpers.py::TestAutoInheritEnv::test_azure_endpoint_derives_resource -q --tb=short
```

Both failures print the host value in the assertion diff. The narrow class-level cleanup means future mirror/derive tests can still accidentally inherit real host credentials.

Recommended fix: pure mirror/derive tests should call `auto_inherit_env(env, source_env={})`; inheritance-specific tests should pass an explicit `source_env={...}` instead of depending on or partially clearing the real process env.

Verification notes:
- Branch-head checks passed: `uv sync`, full `pytest`, `ty`, `ruff`.
- Current-merge `pytest` and `ty` passed.
- Current-merge `ruff` failed only on existing main baseline `UP041` in `src/benchflow/sandbox/docker.py:408`, not this PR.
- I also attempted real ACP E2E; provider/auth setup prevented a healthy full run for this PR, so I am not treating E2E as passed here.

bingran-you:

This PR closes #522 by making `auto_inherit_env` tests hermetic. Several tests called `auto_inherit_env(env)` directly, which falls back to `os.environ`; if a developer or CI host has real credentials like `GEMINI_API_KEY`, `GOOGLE_API_KEY`, or related provider vars exported, those values can enter the test dict and then appear in pytest assertion output.

The implementation makes the unit tests explicit about their input source: inheritance tests now pass `source_env={...}`, and pure mirror/derive tests pass `source_env={}`. That keeps the tests focused on the contract under test while cutting off accidental host-env reads.

### Findings

No blocking findings.

The previous host-env leakage pattern appears fixed in the touched test surfaces:

- `tests/test_agent_env_resolution.py:57-75` now isolates Gemini/Google mirror tests with `source_env={}`.
- `tests/test_resolve_env_helpers.py:50-173` now separates inheritance tests from mirror/derive tests by passing explicit `source_env` mappings.
- This is the right abstraction boundary for these unit tests: they exercise `auto_inherit_env` deterministically without reaching into process-global state.

I do not see a maintainability regression. The PR does not add production branching, new helper layers, type looseness, abstraction leakage, or file-size risk. `tests/test_resolve_env_helpers.py` is still large, but this PR reduces net lines and does not push it over the 1k threshold.

### Residual risk / follow-up

The PR body says it added an autouse fixture using `_clear_keys()`, but the actual implementation uses explicit `source_env` arguments instead. I think the code is cleaner than the description, but the PR body should be updated if the description is meant to be exact.

This is test-only and does not touch SkillsBench, trajectories, verifiers, skill loading, Daytona, or Docker behavior. I do not think a Daytona/Docker canary is required for this merge; the relevant risk is unit-test hermeticity.

bingran-you (approved):

PR #576, opened 2026-05-28
Add mini-swe-agent harness via ACP shim

bingran-you:
## Summary
Integrates [mini-swe-agent](https://github.com/SWE-agent/mini-swe-agent) as a first-class benchflow agent (`mini-swe`, aliases `mini` / `minisweagent` / `mini-swe-agent`), following the same integration contract as the other supported agents.

mini-swe is a deliberately minimal, single-bash-tool harness for apples-to-apples model comparison. A new in-process ACP shim runs its `DefaultAgent` loop and loads its bundled `mini.yaml` **verbatim** (minus the interactive `mode` key), so the upstream guardrails are reproduced faithfully: single `bash` tool, shared system/instance templates, >10k output truncation, malformed-tool-call retry.

## Files
- `src/benchflow/agents/mini_swe_acp_shim.py` (new) — import-safe ACP shim (no top-level side effects; stdout isolation + minisweagent import happen in `main()`).
- `src/benchflow/agents/registry.py` — one additive `AgentConfig` + aliases; installs into an isolated `/opt/benchflow/mini-swe-venv`.
- `tests/integration/configs/mini-swe.yaml` + `tests/integration/run.sh` — agent integration matrix.
- `tests/conformance/run_conformance.py` — mini-swe smoke model + env keys.
- `tests/test_mini_swe_routing.py`, `tests/test_mini_swe_submit.py` — routing + submit-lifecycle tests.
- `src/benchflow/sandbox/docker.py` — drive-by fix of a pre-existing `ruff UP041` error (added in #575) that failed `ruff check src tests` for every PR off main. Behavior-preserving.

## Provider wiring
The shim reads `BENCHFLOW_PROVIDER_*` directly (like `openclaw`/`pi`/`opencode`/`harvey-lab`) — no `env.py` changes; the usage proxy is honored via the injected litellm `api_base`, so token usage is captured the same way as other agents.

`_litellm_prefix` reconstructs the litellm provider prefix from `BENCHFLOW_PROVIDER_PROTOCOL`. mini-swe drives `litellm.completion` (chat-completions / anthropic-messages, **not** the OpenAI Responses API); `openai-responses` only comes from aws-bedrock, whose proxy also exposes an anthropic-messages surface, so Anthropic models route there. This makes Azure (openai-completions) and Bedrock (Claude via the proxy's `/v1/messages`) both work.

## Review follow-ups (thermo-nuclear, two rounds)
- **Per-action ACP lifecycle**: `execute_actions` drives the env loop itself — each action emits start→result around its own `env.execute`; the submit action is closed with the submission; actions that never run (anything after submit in a multi-tool-call turn) emit nothing rather than being falsely marked completed. Fixes both the original dangling-submit bug and the multi-action pollution case.
- **Infra-error classification**: unexpected exceptions in `session/prompt` return a JSON-RPC error (not a successful `end_turn`), so auth/provider/protocol/runtime failures are classified as agent/infra errors instead of masquerading as task failures (matches openclaw). The agent's own task failures still return normally with an `exit_status`.
- **Import-safe shim**: stdout redirect + minisweagent import live in `main()`/a factory, so the routing policy is importable/unit-testable without the sandbox runtime and importing the module never clobbers stdout.
- **Parity coverage**: mini-swe added to the integration matrix (`configs/` + `run.sh`) and the conformance smoke map.

## Test plan
- [x] `ruff check src tests`, `ruff format --check src tests`, `ty check` — clean
- [x] Full unit suite: **2504 passed, 3 skipped, 0 failed**
- [x] `test_mini_swe_routing.py` (Azure/Bedrock/anthropic/empty routing); `test_mini_swe_submit.py` (multi-action submit lifecycle — executed action → real output completed, submit → submission completed, post-submit action → not emitted)
- [x] Real e2e on SkillsBench `weighted-gdp-calc`, **Azure `azure-foundry-openai/gpt-5.5`**: healthy trajectory, 11 tool calls all `completed` (0 dangling), agent iterates on outputs, usage extracted (provider_response)
- [x] Real e2e on SkillsBench `weighted-gdp-calc`, **Bedrock `aws-bedrock/us.anthropic.claude-opus-4-7`**: healthy trajectory, 9 tool calls all `completed` (0 dangling), usage extracted
- [x] Real e2e on SkillsBench `pdf-excel-diff`: reward 1.0 on both Azure gpt-5.5 and Bedrock opus-4.7
- [x] Regression — existing `claude-agent-acp` through the same pipeline unaffected

Note: `reward 0.0` on `weighted-gdp-calc` reflects model/task difficulty, not the integration — trajectory, ACP lifecycle, and token-usage extraction are all healthy.


bingran-you:

PR #576 adds `mini-swe-agent` as a first-class BenchFlow ACP agent. It registers a new `mini-swe` agent, installs mini-swe into an isolated venv, launches a Python ACP shim, routes BenchFlow provider env into LiteLLM, emits ACP tool-call lifecycle events, and adds routing/submit tests plus integration/conformance config.

The shape is reasonable, but I would not merge yet. There are experiment-health and maintainability issues that can make trajectories incomplete or the harness non-reproducible.

### Blocking findings

1. **ACP tool trajectories do not record the observation the agent actually saw.**

   In `src/benchflow/agents/mini_swe_acp_shim.py:217-220`, the shim emits ACP tool results using only `output.get("output", "")`. That drops `returncode`, `exception_info`, and mini-swe's own observation rendering. Then `src/benchflow/agents/mini_swe_acp_shim.py:233-237` separately feeds the model the formatted mini-swe observation.

   That means the BenchFlow ACP/tool trajectory is not the same evidence the agent used. A failed command is still marked `completed`, but the ACP trajectory may not show its return code. Long outputs are truncated to 2k chars here, while mini-swe's config exposes up to 10k head/tail to the model. For SkillsBench/experiment review, this is a trajectory-health blocker.

   **Handoff:** emit the same rendered observation mini-swe gives the model, or emit structured JSON text containing at least `returncode`, `output`/`output_head`/`output_tail`, `elided_chars`, and `exception_info`. Add non-skipped regressions for a nonzero command and a >10k output command, asserting the ACP `tool_call_update` includes return code and truncation metadata.

2. **The harness depends on an unpinned upstream package, while the critical contract test is skipped in normal CI.**

   `src/benchflow/agents/registry.py:609-610` installs `mini-swe-agent` with no version or commit pin. Meanwhile `tests/test_mini_swe_submit.py:16` skips the submit/lifecycle regression unless `minisweagent` happens to be installed locally.

   This is too fragile for a benchmark harness. The shim hand-copies part of `DefaultAgent.execute_actions`; if upstream changes action shape, exception class, or observation contract, BenchFlow can silently start producing unhealthy trajectories. Standard CI would not catch it because the boundary test is skipped.

   **Handoff:** pin `mini-swe-agent` to an exact version or source commit, and make the submit lifecycle test run in CI either by adding the package to dev dependencies or by testing the shim boundary with small fake `DefaultAgent`/`Submitted` modules. Also update regression test docstrings to name PR #576 or the guarding commit per `AGENTS.md`.

3. **The `openai-responses` non-Claude route is encoded as “best-effort” but appears unsupported.**

   `src/benchflow/agents/mini_swe_acp_shim.py:75-76` maps `openai-responses` non-Anthropic models to LiteLLM's `openai` provider, and `tests/test_mini_swe_routing.py:27-28` locks this in for `aws-bedrock/openai.gpt-oss-20b-1:0`.

   But mini-swe uses `litellm.completion`, and the PR description says it cannot speak OpenAI Responses. BenchFlow's Bedrock proxy exposes `/v1/messages` and `/v1/responses`, not a chat-completions surface (`src/benchflow/providers/bedrock_proxy.py:257-261`). This path is likely to become an infra error, not a healthy model failure.

   **Handoff:** for `openai-responses` + non-Anthropic models, fail fast with an explicit unsupported-provider error, or add/select an actual chat-completions endpoint before claiming support. The routing test should assert the chosen supported behavior, not “best effort”.

### Residual risks

- The Bedrock E2E claim should state the sandbox backend. On Daytona/modal, host-side Bedrock proxy routing is unreachable for agents without direct Bedrock support; currently direct support is only OpenHands. Add a Docker-vs-Daytona canary or document mini-swe Bedrock as Docker-only until direct routing exists.
- `src/benchflow/agents/registry.py` is now close to the 1k-line review threshold. The next agent/harness change should extract agent-specific setup out of the central registry.

PR #585, opened 2026-05-30
fix: extend trajectory JSONL redaction to cover Google, AWS, Daytona, and header patterns (#537)

### LINKED ISSUE #537: Trajectory JSONL redaction can miss Google AIza, AWS AKIA, x-api-key, Daytona dtn_ patterns
## Summary

`benchflow.trajectories.types.to_jsonl` only redacts two secret patterns when writing trajectories:

- `sk-ant-` (Anthropic short-form)
- `Bearer ` (HTTP auth header value)

Every other production credential pattern can survive into `trajectory/acp_trajectory.jsonl` and any other JSONL-serialized trajectory artifact if the value enters trajectory content:

- **Google AI: `AIzaSy*`** (Gemini API keys — the v0.5 e2e guide's default credential)
- **AWS access keys: `AKIA*`**, `ASIA*` (Bedrock)
- **Daytona: `dtn_*`** (Daytona SDK access tokens)
- **OpenAI: `sk-proj-*`, `sk-` outside `sk-ant-`**
- **Generic header forms: `x-api-key:`, `authorization: <not Bearer>`, `api-key:`**

The §14 secret-leak grep recipe in `docs/v05-e2e-testing-guide.md` checks for these patterns in job artifacts. The redaction code that should make those greps clean does not actually cover the patterns the guide tests for.

## Concrete file:line

[`src/benchflow/trajectories/types.py:152-178`](https://github.com/benchflow-ai/benchflow/blob/v0.5-integration/src/benchflow/trajectories/types.py#L152-L178) — `to_jsonl` redaction block.

Visible patterns in the regex:
```python
r\"(sk-ant-[a-zA-Z0-9_-]{10})[a-zA-Z0-9_-]+\"
r'(\"authorization\"\\s*:\\s*\"Bearer\\s+)[^\"]+(\")'
```

No regex for `AIzaSy*`, `AKIA*`, `dtn_*`, `x-api-key:`, generic `sk-*`.

## Repro

```bash
# Any v0.5-integration run with a real Gemini key:
export GEMINI_API_KEY=AIzaSy...
export TASKS=$(pwd)/.cache/datasets/benchflow-ai/skillsbench/tasks

uv run bench eval create --tasks-dir $TASKS --include weighted-gdp-calc \
  --agent gemini --model gemini-2.5-flash --sandbox daytona \
  --jobs-dir /tmp/redaction-test

# Force the agent to surface its env (e.g. inject a prompt: "print env" — or
# any task where the agent calls `env`/`set` in a tool). The trajectory will
# carry the unredacted key.
grep -rn 'AIzaSy' /tmp/redaction-test/  # finds matches if the key entered trajectory content
```

Today's e2e runs were clean because the gemini ACP layer does not surface the host env to the trajectory channel. But:

1. Any agent that does `bash 'env'` as a tool call will produce a trajectory entry containing `GEMINI_API_KEY=AIzaSy...`.
2. Any agent that writes env to a file under `artifacts/` would expose it.
3. Any future agent integration that prints request headers (`x-api-key: ...`) will leak.

The redaction is a defense-in-depth layer that should be conservative. Right now it covers only Anthropic + generic Bearer.

## Severity

P0.

- This is a release-blocking **secret-leak attack surface**, not just a hygiene issue. Production e2e runs that share trajectory zips for review (per the v0.5 testing guide's audit goal) can leak keys.
- The guide's own §14 grep recipe asserts none of these patterns appear in job artifacts — but the redaction layer does not enforce the same.
- Companion to [#522](https://github.com/benchflow-ai/benchflow/issues/522) (test-failure key leak): both are \"benchflow's own tooling can echo user keys\" classes of bug.

## Proposed fix

1. Extend `to_jsonl`'s redaction regex set to cover (at minimum) the canonical patterns:
   - `AIzaSy[A-Za-z0-9_-]{20,}` (Google)
   - `AKIA[A-Z0-9]{16,}` / `ASIA[A-Z0-9]{16,}` (AWS access keys)
   - `dtn_[A-Za-z0-9_]{10,}` (Daytona)
   - `sk-[a-zA-Z0-9_-]{10,}` (catch-all for OpenAI / generic; Anthropic's `sk-ant-` is already covered)
   - `x-api-key:\\s*[^\\s\"]+` (HTTP header value)
   - `api[_-]?key[\"']?\\s*[:=]\\s*[\"']?[A-Za-z0-9_-]{20,}` (loose key=value pattern)

2. Move the redaction patterns into a shared module (e.g. `_utils/redaction.py`) so they can be reused by:
   - `summary.json` writer
   - `result.json` writer
   - dashboard ingestion (cross-ref [#494](https://github.com/benchflow-ai/benchflow/issues/494))
   - any future logging path

3. Add a unit test that constructs a trajectory entry containing each pattern and asserts the JSONL output is redacted. CI will catch regressions on every PR.

4. Consider a `BENCHFLOW_REDACTION_EXTRA` env var or config file so operators can add private credential patterns without forking.

## Found via

v0.5 e2e validation arch-audit at `5a7a326`. Reproducible via static read of [`trajectories/types.py:152-178`](https://github.com/benchflow-ai/benchflow/blob/v0.5-integration/src/benchflow/trajectories/types.py#L152-L178).

ElegantLin:
## Summary

`Trajectory.to_jsonl` only redacted two secret patterns (`sk-ant-`, `Bearer`). Every other production credential pattern could leak into `acp_trajectory.jsonl` if it entered trajectory content (e.g. agent running `env` as a tool call).

Added redaction for:
- **Google AI / Gemini**: `AIzaSy*`
- **AWS access keys**: `AKIA*` / `ASIA*` (anchored to exact 20-char length to avoid matching `ASIAPACIFIC` etc.)
- **Daytona**: `dtn_*` (≥16 char suffix to avoid short identifiers)
- **OpenAI project keys**: `sk-proj-*`
- **HTTP headers**: `x-api-key`, `api-key` (Azure)

Extracted patterns into a compiled `_REDACTION_PATTERNS` list and a reusable `redact_trajectory_text()` function for downstream callers.

## Test plan

- [x] 10 parametrized tests covering each secret pattern
- [x] 7 parametrized false-positive guard tests (English words, hyphenated slugs, short identifiers)
- [x] End-to-end tests through `Trajectory.to_jsonl(redact_keys=True/False)`
- [x] 21/21 trajectory tests pass; ruff clean
- [x] Self-reviewed for regex correctness and false positives before submitting

Closes #537

---

<a href="https://app.devin.ai/review/benchflow-ai/benchflow/pull/585" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/docs/v05-e2e-testing-guide.md
+++ b/docs/v05-e2e-testing-guide.md
@@ -438,13 +438,15 @@ uv run python tests/integration/check_results.py "$JOB_DIR" \
 
 ## 14. Secret leak audit
 
-After any eval run, check that no API key **values** leaked into trajectories.
-Redaction (`Trajectory.to_jsonl(redact_keys=True)`, #537/#585) replaces secret
-*values* whole — prefix included — so grep for the live-key shapes, not the
-variable names (a bare `GEMINI_API_KEY=` with a redacted value is not a leak):
+After any eval run, check that no API key **values** leaked into trajectories
+or trainer JSONL artifacts. Redaction (`Trajectory.to_jsonl(redact_keys=True)`
+and Verifiers JSONL export, #537/#585) replaces secret *values* whole — prefix
+included — so grep for the live-key shapes, not the variable names (a bare
+`GEMINI_API_KEY=` with a redacted value is not a leak):
 
 ```bash
-grep -rn "AIzaSy[A-Za-z0-9_-]\{20,\}\|dtn_[A-Za-z0-9_]\{16,\}\|sk-ant-[a-zA-Z0-9_-]\{12,\}" /tmp/test-include/
+SECRET_SHAPES='AIzaSy[A-Za-z0-9_-]{20,}|dtn_[A-Za-z0-9_]{16,}|sk-ant-[A-Za-z0-9_-]{12,}|sk-proj-[A-Za-z0-9_-]{12,}|sk-[A-Za-z0-9]{12,}|(AKIA|ASIA)[A-Z0-9]{16}([^A-Z0-9]|$)|(authorization|x-api-key|x-goog-api-key|api-key|api_key)[[:space:]]*[:=][[:space:]]*"?[^"[:space:],}*]+'
+grep -rnE "$SECRET_SHAPES" /tmp/test-include/
 ```
 
 **Verify:** No matches. (Redacted placeholders read `***REDACTED***`; the
--- a/docs/v05-e2e-testing-guide.md
+++ b/docs/v05-e2e-testing-guide.md
@@ -438,13 +438,17 @@ uv run python tests/integration/check_results.py "$JOB_DIR" \
 
 ## 14. Secret leak audit
 
-After any eval run, check that no API keys leaked into trajectories:
+After any eval run, check that no API key **values** leaked into trajectories.
+Redaction (`Trajectory.to_jsonl(redact_keys=True)`, #537/#585) replaces secret
+*values* whole — prefix included — so grep for the live-key shapes, not the
+variable names (a bare `GEMINI_API_KEY=` with a redacted value is not a leak):
 
 ```bash
-grep -rn "AIzaSy\|dtn_\|GEMINI_API_KEY\|DAYTONA_API_KEY" /tmp/test-include/
+grep -rn "AIzaSy[A-Za-z0-9_-]\{20,\}\|dtn_[A-Za-z0-9_]\{16,\}\|sk-ant-[a-zA-Z0-9_-]\{12,\}" /tmp/test-include/
 ```
 
-**Verify:** No matches.
+**Verify:** No matches. (Redacted placeholders read `***REDACTED***`; the
+variable names themselves are not secrets and may appear.)
 
 ---
 
--- a/src/benchflow/hosted_env.py
+++ b/src/benchflow/hosted_env.py
@@ -33,6 +33,7 @@ from benchflow._utils.result_metadata import (
     trajectory_summary_from_events,
 )
 from benchflow.diagnostics import RolloutDiagnostics
+from benchflow.trajectories.types import redact_acp_trajectory_jsonl
 
 logger = logging.getLogger(__name__)
 
@@ -480,8 +481,7 @@ def _write_run_artifacts(
     }
 
     (result.run_dir / "trajectory" / "acp_trajectory.jsonl").write_text(
-        "\n".join(json.dumps(e, default=str) for e in trajectory)
-        + ("\n" if trajectory else "")
+        redact_acp_trajectory_jsonl(trajectory) + ("\n" if trajectory else "")
     )
 
     result_payload: dict[str, Any] = {
--- a/src/benchflow/rollout.py
+++ b/src/benchflow/rollout.py
@@ -100,6 +100,7 @@ from benchflow.trajectories._capture import (
 )
 from benchflow.trajectories.metrics import count_skill_invocations
 from benchflow.trajectories.tree import RolloutNode, RolloutTree, Step
+from benchflow.trajectories.types import redact_acp_trajectory_jsonl
 from benchflow.usage_tracking import UsageTrackingConfig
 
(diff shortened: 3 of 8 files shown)


bingran-you:

This PR is trying to close #537 by expanding `Trajectory.to_jsonl(redact_keys=True)` beyond the old Anthropic/OpenAI/Bearer-only redaction. The proposed solution moves regexes into a compiled `_REDACTION_PATTERNS` list, exposes `redact_trajectory_text()`, and adds focused tests for Google, AWS, Daytona, OpenAI project keys, and JSON-shaped `x-api-key` / `api-key` headers.

The direction is right, but I would not merge this yet. There are still high-confidence gaps against the issue this PR claims to close.

### Blocking findings

1. **`x-api-key:`, `api-key:`, and non-Bearer `authorization:` text still leak.**

   `src/benchflow/trajectories/types.py:160-173` only redacts headers when trajectory text contains JSON object keys like `"x-api-key": "..."` or `"api-key": "..."`. But #537 explicitly calls out generic header forms such as `x-api-key: ...`, `api-key: ...`, and `authorization: <not Bearer>`, which can appear when an agent prints env, curls with verbose headers, dumps request logs, or writes tool output into a trajectory. For Azure-style `api-key: abc123secret456value`, the value does not match any provider-prefix regex and survives.

   **Handoff:** add redaction for header/key-value text forms, not just JSON keys, for at least `x-api-key`, `api-key`, `x-goog-api-key`, and `authorization` regardless of scheme. Add tests that pass raw trajectory-content strings such as `x-api-key: abc123secret456value`, `api-key=abc123secret456value`, and `authorization: Token abc123secret456value` through `Trajectory.to_jsonl()`, not only through the helper.

2. **The documented secret-leak audit still fails on redacted Google/Daytona env output.**

   The v0.5 audit in `docs/v05-e2e-testing-guide.md:443` checks for `AIzaSy`, `dtn_`, `GEMINI_API_KEY`, and `DAYTONA_API_KEY` and expects no matches. This implementation preserves those prefixes in `src/benchflow/trajectories/types.py:149-159`: Google becomes `AIzaSy....***REDACTED***`, and Daytona becomes `dtn_....***REDACTED***`. If trajectory content contains `GEMINI_API_KEY=AIzaSy...`, the output still contains both `GEMINI_API_KEY` and `AIzaSy`, so the experiment-health audit still flags the artifact.

   **Handoff:** either redact the whole matched token/pair for audit-sensitive families, or update the audit and tests to distinguish redacted placeholders from live secrets. Include an end-to-end regression that constructs a trajectory body containing `GEMINI_API_KEY=AIzaSy...` and `DAYTONA_API_KEY=dtn_...`, writes JSONL, and proves the same leak-audit intent passes.

### Test and maintainability gaps

The new tests are useful, but they mostly test `redact_trajectory_text()` on idealized JSON strings. The risky path is benchmark artifact publication after messy agent/tool output, so coverage should include `Trajectory.to_jsonl()` with raw shell env output, raw HTTP headers, and key/value logs.

Also, per this repo's `AGENTS.md`, regression tests should name the PR/commit they guard in the docstring. The new test file should make that explicit.

Verdict: request changes. The abstraction is small and the file stays well under 1k lines, but the implementation still misses core cases from #537 and leaves the documented experiment-health audit ambiguous.

ElegantLin:
Follow-up (`589597a`) after a deeper self-review found this PR didn't actually achieve its goal:
- **Redaction wasn't wired to `acp_trajectory.jsonl`** (the named target) — only `llm_trajectory.jsonl` was redacted. Added `redact_acp_trajectory_jsonl()` and wired it into all three ACP-trajectory write sites (rollout local write, verifier-publish tmpfile, hosted_env).
- **Namespaced env keys leaked**: `GEMINI_API_KEY=...` / `{"openai_api_key": ...}` slipped through due to a lookbehind on the `api_key` rule. Removed it (the `\1` capture preserves the key name, so no over-redaction).
New tests assert the raw secret is absent from the written file.

PR #593, opened 2026-06-01
Add openai + us-openai provider routing

EYH0602:
## Summary
- Register first-party `openai` provider so `model: openai/<id>` routes correctly (was silently falling back to unauthenticated Anthropic).
- Add `us-openai` provider for the US data-residency endpoint (`https://us.api.openai.com/v1`), reusing `OPENAI_API_KEY`.
- Expose both `openai-completions` and `openai-responses` endpoint variants on each.
- TODO comment left in the registry for `eu-openai` when needed.

## Root cause
`openai/<model>` looked provider-prefixed, but `openai` was not in `PROVIDERS`. BenchFlow skipped `BENCHFLOW_PROVIDER_*` injection, and pi-acp (and any launcher mirroring its conditional) fell through to the Anthropic env path — producing zero-tool-call trials with no error surfaced.

## What changes for users
- `model: openai/gpt-5.4-mini` — routes to `api.openai.com` with `OPENAI_API_KEY`.
- `model: us-openai/gpt-5.4-mini` — routes to `us.api.openai.com` with the same key.

## Validation
- `uv run pytest tests/test_providers.py tests/test_resolve_env_helpers.py tests/test_pi_acp_launcher.py tests/test_registry_invariants.py tests/test_agent_registry.py -q`
- `uv run ruff check src/benchflow/agents/providers.py tests/test_providers.py tests/test_resolve_env_helpers.py`
- `uv run ty check src/benchflow/agents/providers.py src/benchflow/agents/env.py`
- Verified end-to-end against fork's `local` branch.

## Test plan
- [ ] CI passes
- [ ] Reviewer confirms provider registry placement / naming

### DIFF
--- a/src/benchflow/agents/env.py
+++ b/src/benchflow/agents/env.py
@@ -50,6 +50,7 @@ _CLAUDE_OAUTH_TOKEN_ENV = "CLAUDE_OAUTH_TOKEN"
 _CUSTOM_OPENAI_ENDPOINT_KEYS = frozenset(
     {"BENCHFLOW_PROVIDER_BASE_URL", "OPENAI_BASE_URL"}
 )
+_CANONICAL_OPENAI_URL = "https://api.openai.com/v1"
 _GENERIC_PROVIDER_OVERRIDE_KEYS = frozenset(
     {"BENCHFLOW_PROVIDER_BASE_URL", "BENCHFLOW_PROVIDER_API_KEY"}
 )
@@ -246,7 +247,14 @@ def _is_codex_native_openai_context(
     model: str | None,
     required_key: str | None,
 ) -> bool:
-    """True when Codex can use its own OpenAI auth mechanisms directly."""
+    """True when Codex can use its own OpenAI auth mechanisms directly.
+
+    Native auth covers bare model IDs (``gpt-*``) and the first-party
+    ``openai/`` provider prefix pointing at ``api.openai.com``. Custom or
+    proxy providers (``vllm/``, ``us-openai/``, etc.) must supply
+    ``OPENAI_API_KEY`` explicitly — subscription/access-token auth does not
+    apply to them.
+    """
     if agent != "codex-acp" or required_key != "OPENAI_API_KEY":
         return False
     if model is None:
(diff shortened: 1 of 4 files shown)


bingran-you:

This PR fixes provider-prefixed OpenAI model routing. Today, `openai/<model>` looks like a BenchFlow provider prefix, but `openai` is not registered in `PROVIDERS`, so provider env injection is skipped and some agents can fall through to the wrong provider-shaped path. The proposed solution registers first-party `openai` plus `us-openai`, exposes both Chat Completions and Responses protocol URLs, and adds focused provider/env tests.

### Blocking finding

1. **`openai/` registration regresses existing Codex native auth paths.**

   Registering `openai` makes `find_provider("openai/...")` return non-`None`, but the existing Codex auth helpers treat Codex native OpenAI auth as valid only when `find_provider(model) is None` (`src/benchflow/agents/env.py:255-257`). That means `codex-acp` with `model="openai/gpt-5.4-mini"` can reject supported Codex auth modes that still work for bare `gpt-*` models: `CODEX_API_KEY`, `CODEX_ACCESS_TOKEN`, `CODEX_AUTH_JSON`, and host `~/.codex/auth.json` subscription auth.

   The new tests exercise `resolve_provider_env`, but they do not cover the full `resolve_agent_env` validation/subscription-auth layer where the regression happens.

   **Handoff:** make the first-party `openai` provider still count as Codex native OpenAI for auth purposes while keeping custom/proxy providers protected. A small helper such as `_is_codex_first_party_openai_context(agent, model, required_key)` could allow either no provider or provider name `openai` with the canonical `https://api.openai.com/v1` endpoint. Add regression tests against `resolve_agent_env("codex-acp", "openai/gpt-5.4-mini", ...)` for `CODEX_API_KEY`, `CODEX_ACCESS_TOKEN`, `CODEX_AUTH_JSON`, and host subscription auth. Keep `vllm/` and other custom endpoints rejecting subscription/access-token auth as they do today.

### Non-blocking notes

- `us-openai` gets provider-level coverage, but there is no full `resolve_agent_env` coverage for either `pi-acp` or `codex-acp` using `us-openai/...`. Given this PR is about launch routing, add at least one full-pipeline test proving the regional base URL survives env mapping.
- `tests/test_resolve_env_helpers.py` is close to the 1k-line review threshold. The next additions should probably split provider/auth resolution cases into a narrower module.
- The `openai` and `us-openai` registry entries duplicate the same shape. Fine for two entries, but if more regional variants are added, use a tiny local factory/helper to avoid registry drift.

I would block merge on the Codex auth regression. The provider-registration direction is reasonable, but it needs to preserve the existing native Codex auth contract for first-party `openai/...` models.

EYH0602:
Confirmed true positive — pushed a fix in 2a14405.

**Context for the registration.** The `openai/` prefix is for non-Codex agents (e.g. pi-acp) — they need a registered provider so `resolve_provider_env` injects `BENCHFLOW_PROVIDER_*` and the agent-native env mapping. Codex itself is happy with bare `gpt-*`; the regression only kicks in when a user routes Codex through the new prefix.

**Fix.** Two small edits in `src/benchflow/agents/env.py`:

1. `_is_codex_native_openai_context` now also returns True when `find_provider(model)` matches the first-party `openai` provider pointing at the canonical `https://api.openai.com/v1`. `vllm/`, `us-openai/`, and other custom/proxy providers still fall through to "non-native" and keep requiring an explicit `OPENAI_API_KEY`.
2. `_has_custom_openai_endpoint` now ignores `BENCHFLOW_PROVIDER_BASE_URL` / `OPENAI_BASE_URL` when their value is the canonical OpenAI URL — needed because `resolve_provider_env` writes those keys for `openai/...` and the old predicate treated them as "custom".

**Tests.** `TestResolveAgentEnvCodexOpenAIPrefix` in `tests/test_resolve_env_helpers.py` exercises `resolve_agent_env("codex-acp", "openai/gpt-5.4-mini", ...)` for all four native auth modes (`CODEX_API_KEY` → bridged to `OPENAI_API_KEY`, `CODEX_ACCESS_TOKEN`, `CODEX_AUTH_JSON`, host `~/.codex/auth.json` subscription auth), plus negative cases asserting `us-openai/` and `vllm/` still require an explicit key.

712 tests in the env/auth/provider/codex/openai sweep pass.

PR #594, opened 2026-06-01
fix: upload absolute str skills_dir host dirs in scene skill activation

Yiminnn:
## Summary
`Rollout._activate_step_skills` skips uploading scene skills when `skills_dir` is an absolute **str** host path, so deployed SkillsBench task skills never reach the agent — the with-skills arm silently becomes a no-op.

## Root cause
`src/benchflow/rollout.py::_activate_step_skills` gated the upload on the argument **type**:
```python
if isinstance(source_value, os.PathLike) and local_source.is_dir():
    ...upload...
elif source.startswith("/"):
    remote_source = source            # treated as already-in-sandbox, NO upload
elif local_source.is_dir():           # unreachable for absolute paths
    ...upload...
```
The SkillsBench entrypoint always passes an absolute **str** host path (`EvaluationConfig.skills_dir` is typed `str`; the YAML loaders and `_resolve_skills_dir` all `str(...)`). So `isinstance(str, os.PathLike)` is `False`, the host path falls into the `startswith("/")` "already in the sandbox" branch (no upload), and `_link_skill_paths` then `rm -rf`s the discovery dir and `ln -sfn`s it to a host path that does not exist inside the container → **dangling symlink** → the task skill is never discovered.

OpenHands' built-in/public skills load independently, so the catalog still shows those — the deployed **task** skill is the only thing missing.

## Evidence (raw leaderboard trajectories)
- OpenHands with-skills skill delta: **PR2 +13.1pp, PR3 +11.1pp → PR4 −1.1pp**.
- PR4 raw with-skills trajectories: task skill **absent** from `<available_skills>` (only the 42 generic built-ins; byte-identical to the no-skills catalog); `invoke_skill` called **0/10**. PR2/PR3: task skill is the first `<skill>` entry; `invoke_skill` 7/14.
- Bisected to the commit that introduced the `isinstance(..., os.PathLike)` gate.

## Fix
Key the upload on the directory existing on the **host** (the pre-regression behaviour) rather than the argument's type:
```python
if local_source.is_dir():            # real host dir -> upload (str or PathLike)
    ...upload...
elif source.startswith("/"):         # absolute path not on host -> sandbox path from an earlier scene
    remote_source = source
else:
    raise FileNotFoundError(...)
```

## Note on prior intent (please sanity-check)
The gate was meant to treat absolute **str** `skills_dir` values as in-sandbox paths produced by an earlier scene (self-gen flow). That intent is preserved: such sandbox paths (e.g. `/app/generated-skills`) don't exist on the orchestrator host, so `local_source.is_dir()` is `False` and they're still linked as-is with no upload. The only behaviour change is that an absolute str path that **does** exist as a host directory is now uploaded — exactly the SkillsBench case. Existing self-gen tests still pass.

## Test plan
- New regression test `test_scene_skills_upload_absolute_str_host_dir`: an absolute **str** host `skills_dir` is uploaded to `/skills/<scene>` and linked into the agent discovery path.
- `tests/test_scene_outbox_trial.py` — 12/12 pass (incl. existing self-gen tests).
- `tests/test_scene.py tests/test_rollout_config_path_coercion.py tests/test_rollout_branch.py tests/test_agent_setup.py` — 49/49 pass.
- `ruff check` / `ruff format --check` / `ty check` clean.

---
Draft for maintainer review. Root cause + evidence derived from raw HF leaderboard trajectories (PR2/PR3/PR4) and a git bisect of the skill-deploy path.

### DIFF
--- a/src/benchflow/rollout.py
+++ b/src/benchflow/rollout.py
@@ -2128,19 +2128,24 @@ class Rollout:
 
         scene_name = str(step.data.get("scene") or "scene")
         role = scene_step_role(step)
-        source_value = skills_dir
-        source = str(source_value)
+        source = str(skills_dir)
         local_source = Path(source).expanduser()
-        if isinstance(source_value, os.PathLike) and local_source.is_dir():
+        # Upload whenever skills_dir resolves to a real directory on the
+        # orchestrator host, regardless of whether it arrived as a str or a
+        # PathLike. The SkillsBench entrypoint passes an absolute *str* host
+        # path (EvaluationConfig.skills_dir is typed str), so gating the upload
+        # on isinstance(..., os.PathLike) silently skipped it: the path then
+        # fell through to the "already inside the sandbox" branch below and
+        # _link_skill_paths produced a dangling symlink, so deployed task skills
+        # never reached the agent. An absolute path that does NOT exist on the
+        # host is a sandbox path produced by an earlier scene and is linked
+        # as-is.
+        if local_source.is_dir():
             remote_source = f"/skills/{_safe_skill_name(scene_name)}"
             await _ensure_sandbox_dir(self._env, Path(remote_source).parent)
             await self._env.upload_dir(local_source, remote_source)
         elif source.startswith("/"):
             remote_source = source
-        elif local_source.is_dir():
-            remote_source = f"/skills/{_safe_skill_name(scene_name)}"
-            await _ensure_sandbox_dir(self._env, Path(remote_source).parent)
-            await self._env.upload_dir(local_source, remote_source)
         else:
             raise FileNotFoundError(f"Scene skills_dir not found: {skills_dir}")
 
--- a/tests/test_scene_outbox_trial.py
+++ b/tests/test_scene_outbox_trial.py
@@ -359,3 +359,39 @@ async def test_scene_skills_prefers_remote_string_path_over_matching_host_path(
 
     assert trial._env._uploads == []
     assert not any(cmd.startswith("mkdir -p /skills") for cmd in trial._env._exec_log)
+
+
+async def test_scene_skills_upload_absolute_str_host_dir(tmp_path: Path) -> None:
+    """An absolute *str* skills_dir that exists on the host is uploaded.
+
+    Regression test guarding the OpenHands with-skills delivery fix: a prior
+    change gated the scene-skill upload on ``isinstance(skills_dir, os.PathLike)``,
+    but the SkillsBench entrypoint passes an absolute *str* host path
+    (``EvaluationConfig.skills_dir`` is typed ``str``). The gate skipped the
+    upload, the path fell through to the in-sandbox branch, and
+    ``_link_skill_paths`` produced a dangling symlink -- so deployed task skills
+    never reached the agent's available-skills catalog and the with-skills arm
+    became a no-op. Uploading must key on the directory existing on the host,
+    not on the argument's type.
+    """
+    skills_root = tmp_path / "environment" / "skills"
+    (skills_root / "mesh-analysis").mkdir(parents=True)
+    (skills_root / "mesh-analysis" / "SKILL.md").write_text(
+        "---\nname: mesh-analysis\ndescription: analyze meshes\n---\n# Mesh\n"
+    )
+    scene = Scene(
+        name="solve",
+        roles=[Role("solver", "claude-agent-acp", "haiku")],
+        turns=[Turn("solver")],
+        skills_dir=str(skills_root),  # absolute STR host path, as production passes
+    )
+    trial = _make_trial(scene)
+    trial._agent_cwd = "/app"
+
+    await trial._activate_step_skills(compile_scenes_to_steps([scene])[0])
+
+    assert trial._env._uploads == [(skills_root, "/skills/solve")]
+    assert any(
+        "ln -sfn /skills/solve /home/agent/.claude/skills" in cmd
+        for cmd in trial._env._exec_log
+    )


Yiminnn:
## Before/after proof (no Docker)

Drives the real `Rollout._activate_step_skills` with a fake sandbox on the **unchanged** code vs the **fix**, for two minimal cases:
- **A** — SkillsBench with-skills: `skills_dir` is an absolute **str** host path that exists (what production passes).
- **B** — self-gen: `skills_dir` is an absolute **str** sandbox path produced by an earlier scene (not on host).

### Result

| case | **UNCHANGED** (`isinstance` gate) | **FIXED** (host `is_dir` first) |
|---|---|---|
| **A** host str path | `uploaded=False` → `~/.claude/skills` → `/tmp/.../environment/skills` = **DANGLING → skill MISSING** ✗ | `uploaded=True` → `~/.claude/skills` → `/skills/solve` = **skill LOADS** ✓ |
| **B** sandbox str path | linked as-is ✓ | linked as-is ✓ (identical) |

Captured output:

```text
================ UNCHANGED (isinstance gate present) ================
  [A  skillsbench host str path]
      skills_dir = '/tmp/.../environment/skills'  (exists on host? True)
      uploaded   = False   uploads=[]
      agent link = ~/.claude/skills -> /tmp/.../environment/skills
      VERDICT    = ✗ BUG: host dir linked but NOT uploaded -> DANGLING in container -> skill MISSING
  [B  self-gen sandbox str path]
      skills_dir = '/app/generated-skills-not-on-host'  (exists on host? False)
      uploaded   = False   uploads=[]
      agent link = ~/.claude/skills -> /app/generated-skills-not-on-host
      VERDICT    = ✓ in-sandbox path linked as-is (from earlier scene; not a host dir)

================ FIXED (host is_dir first) ================
  [A  skillsbench host str path]
      skills_dir = '/tmp/.../environment/skills'  (exists on host? True)
      uploaded   = True   uploads=[(PosixPath('/tmp/.../environment/skills'), '/skills/solve')]
      agent link = ~/.claude/skills -> /skills/solve
      VERDICT    = ✓ skill UPLOADED into sandbox and linked -> agent LOADS it
  [B  self-gen sandbox str path]
      uploaded   = False   agent link = ~/.claude/skills -> /app/generated-skills-not-on-host
      VERDICT    = ✓ in-sandbox path linked as-is
```

Confirms: the unchanged code drops the host-sourced task skill (Case A → dangling symlink → absent from `<available_skills>`, matching PR4); the fix uploads + links it; the self-gen path (Case B) is unchanged. Same conclusion as the added unit test `test_scene_skills_upload_absolute_str_host_dir` (fails on unchanged, passes on fix).

<details><summary>repro script</summary>

```python
import asyncio, inspect, re, tempfile
from dataclasses import dataclass
from pathlib import Path
from benchflow.rollout import Role, Rollout, RolloutConfig, Scene, Turn
from benchflow.scenes import compile_scenes_to_steps


@dataclass
class FakeExecResult:
    stdout: str = ""
    stderr: str = ""
    return_code: int = 0


class FakeEnv:
    def __init__(self):
        self.exec_log, self.uploads = [], []

    async def exec(self, cmd, **kw):
        self.exec_log.append(cmd)
        return FakeExecResult()

    async def upload_dir(self, src, dst):
        self.uploads.append((Path(src), dst))


def make_trial(scene):
    cfg = RolloutConfig(task_path=Path("tasks/fake"), scenes=[scene], environment="docker")
    t = Rollout(cfg)
    t._env, t._resolved_prompts, t._agent_cwd = FakeEnv(), ["x"], "/app"
    return t


async def run_case(label, skills_dir):
    scene = Scene(name="solve", roles=[Role("solver", "claude-agent-acp", "haiku")],
                  turns=[Turn("solver")], skills_dir=skills_dir)
    t = make_trial(scene)
    await t._activate_step_skills(compile_scenes_to_steps([scene])[0])
    link = next((c for c in t._env.exec_log if "ln -sfn" in c), "")
    src = (re.search(r"ln -sfn (\S+) ", link) or [None, "?"])[1]
    host = Path(skills_dir).expanduser().is_dir()
    verdict = ("✓ UPLOADED -> agent loads" if src.startswith("/skills/")
               else "✗ BUG: host dir linked but NOT uploaded -> dangling" if host
               else "✓ in-sandbox path linked as-is")
    print(f"  [{label}] uploaded={bool(t._env.uploads)} link=~/.claude/skills->{src}  {verdict}")


async def main():
    gated = "isinstance(source_value, os.PathLike)" in inspect.getsource(Rollout._activate_step_skills)
    print("UNCHANGED" if gated else "FIXED")
    with tempfile.TemporaryDirectory() as d:
        s = Path(d) / "environment" / "skills" / "mesh-analysis"
        s.mkdir(parents=True)
        (s / "SKILL.md").write_text("---\nname: mesh-analysis\ndescription: x\n---\n")
        await run_case("A host str path", str(s.parent))
    await run_case("B sandbox str path", "/app/generated-skills-not-on-host")


asyncio.run(main())
```
Run: check out the buggy and fixed `rollout.py` in turn (editable install) and `python repro.py` each time.
</details>

bingran-you:
## Validation run: PR #594 skill delivery confirmed working ✅

I ran a **with-skills** SkillsBench task through this branch end-to-end to confirm the scene-skill upload fix actually delivers task-bundled skills to the agent, and cross-checked it against the Harbor harness as a reference.

### Setup
- **Branch:** `fix/openhands-scene-skill-upload` @ `2d95e2a`
- **Task:** `benchflow-ai/skillsbench/tasks/latex-formula-extraction` — bundles two skills (`marker`, `pdf`) under `environment/skills/`
- **Command:** `bench eval create --tasks-dir <skillsbench>/tasks --include latex-formula-extraction --agent openhands --model aws-bedrock/us.anthropic.claude-opus-4-6-v1 --sandbox daytona --skills-dir <task>/environment/skills --skill-mode default --concurrency 1`
- **Sandbox/model:** Daytona + AWS Bedrock, `us.anthropic.claude-opus-4-6-v1` (us-west-2), 1 trial

### Result for this PR: skills are delivered AND used ✅
benchflow run log:
```
Skills injected into Dockerfile: 2 items
Skills already injected via Dockerfile
Skills distributed to 2 paths for openhands
```
The agent then **actually used** the `marker` skill — the ACP trajectory shows it running the skill's scripts (`marker_single` / `marker_to_markdown.py`) to convert the PDF. So with `EvaluationConfig.skills_dir` arriving as an absolute **str** host path, the skills now reach the agent's catalog. This is exactly the path the PR fixes; before it, the upload was gated on `isinstance(..., os.PathLike)` and silently skipped (the dangling-symlink no-op described in the PR body).

> Minor FYI (not a blocker): `result.json → agent_result.n_skill_invocations` was `0`, even though the agent demonstrably used `marker`. The agent invokes the skill as plain shell calls to the bundled scripts rather than a tracked skill-protocol call, so that counter under-reports real usage here. Might be worth a follow-up if that metric is meant to capture script-based skill use.

### Run outcome
- **reward `0.0`**, but a **near-pass** — clean run, no errors/timeouts. 33 tool calls, ~1.16M tokens (1.09M cache-read), ~7 min agent execution.
- Produced exactly 5 formulas; **4/5 byte-identical to the oracle**, including correctly keeping the buggy formula 3 verbatim and appending its corrected version (the task's "fix problematic formulas" step).
- The only miss is **cosmetic LaTeX on formula 1**, which the task's strict MathJax-rendered set-diff rejects:
```
oracle    : …^N  \frac{1}{\left(x_i-x_j\right)^2} \cdot \text{sgn}\left(x_i-x_j\right)
benchflow : …^{N} \frac{1}{(x_i-x_j)^2}           \cdot \operatorname{sgn}(x_i-x_j)
```
(`\text{sgn}`→`\operatorname{sgn}`, `\left(\right)`→`()`, `^N`→`^{N}` — visually ~identical, semantically equal.)

### Cross-check vs Harbor (same task / model / skills)
Ran the identical task on Harbor 0.3.0 with `openhands-sdk` for parity:

| | benchflow @ this PR (`openhands` ACP) | Harbor (`openhands-sdk`) |
|---|---|---|
| Reward | **0.0** (near-pass) | **1.0** (pass) |
| Skills delivered + used | ✅ | ✅ (`Loaded 2 skills`: marker, pdf) |
| Formulas written | 5 (1 cosmetic miss) | 5 (exact match) |
| Tokens | ~1.16M | ~858K |

**Both harnesses delivered the skills and the agent used `marker` in both.** The reward gap is entirely the formula-1 cosmetic LaTeX difference (agent-frontend formatting × brittle verifier), **not** a skill-delivery difference — i.e. this PR's fix behaves correctly and on par with the reference harness.

### Environment notes (FYI, unrelated to this PR)
- benchflow on Daytona+Bedrock needs the `sandbox-daytona` extra and routes `openhands` to **direct** Bedrock (the host-side Bedrock proxy is unreachable from a remote Daytona sandbox); usage telemetry went through the Daytona sandbox-local proxy. Worked as expected.
- I originally targeted Opus **4.8**, but it is currently **unrunnable through the OpenHands + litellm(1.86) Bedrock path**: Opus 4.8 rejects the deprecated extended-thinking format the stack emits — `BedrockException: "thinking.type.enabled" is not supported for this model. Use "thinking.type.adaptive" and "output_config.effort"` → surfaces as `ACP error -32603`. It failed identically on both harnesses, so I fell back to Opus 4.6. Flagging since it will affect any Opus-4.8 OpenHands run, independent of this PR.

src/benchflow/rollout.py
```diff
@@ -2128,19 +2128,24 @@ async def _activate_step_skills(self, step: Step) -> None:
...
+        # orchestrator host, regardless of whether it arrived as a str or a
+        # PathLike. The SkillsBench entrypoint passes an absolute *str* host
+        # path (EvaluationConfig.skills_dir is typed str), so gating the upload
+        # on isinstance(..., os.PathLike) silently skipped it: the path then
+        # fell through to the "already inside the sandbox" branch below and
+        # _link_skill_paths produced a dangling symlink, so deployed task skills
+        # never reached the agent. An absolute path that does NOT exist on the
+        # host is a sandbox path produced by an earlier scene and is linked
+        # as-is.
+        if local_source.is_dir():
```
bingran-you, line 2143: This fixes the SkillsBench host-path case, but it also makes sandbox-vs-host semantics depend on whatever happens to exist on the orchestrator filesystem. A self-gen solver scene uses `skills_dir="/app/generated-skills"`; if BenchFlow is run from a container/dev environment where `/app/generated-skills` exists on the host side, this branch uploads that host directory to `/skills/self-gen-solver` and links the agent there, instead of linking the skills the creator produced inside the sandbox at `/app/generated-skills`. That is a real boundary regression from the old type-based split.

Can we keep the boundary explicit instead? For example, normalize the legacy/Evaluation `skills_dir` host path to a `Path` before it becomes `Scene.skills_dir` (so the SkillsBench case uploads), while preserving absolute `str` scene paths as in-sandbox paths. Also, the nearby `test_scene_skills_prefers_remote_string_path_over_matching_host_path` does not actually create `/app/generated-skills` on the host, so it does not cover this collision; please add/adjust a test that simulates that exact case.

tests/test_scene_outbox_trial.py
```diff
@@ -359,3 +359,39 @@ async def test_scene_skills_prefers_remote_string_path_over_matching_host_path(
 
     assert trial._env._uploads == []
     assert not any(cmd.startswith("mkdir -p /skills") for cmd in trial._env._exec_log)
+
+
+async def test_scene_skills_upload_absolute_str_host_dir(tmp_path: Path) -> None:
+    """An absolute *str* skills_dir that exists on the host is uploaded.
```
bingran-you, line 365: Repo convention says regression-test docstrings must name the PR/commit they guard. This explains the incident, but it does not name PR #594 or the commit/PR that introduced the `PathLike` gate. Can we make that explicit here so future readers know exactly which regression this test is protecting?

bingran-you (changes requested):
I think the root-cause analysis is right and the happy-path fix restores SkillsBench absolute-str host-path uploads. I also verified the relevant tests, the full test suite, ruff, and ty locally.

I am requesting changes because the implementation now makes the host-vs-sandbox source boundary depend on the orchestrator filesystem. That can regress self-gen when the sandbox path happens to exist on the host/container, and one nearby test still claims to guard the old remote-string behavior without actually covering that collision. Please tighten that boundary before merge.

bingran-you (approved):

PR #598, opened 2026-06-02
Enable openhands SkillsBench runs for Bedrock Opus 4.8 (MAX thinking) + Gemini usage tracking + apples-to-apples token accounting

bingran-you:
## TL;DR

This PR unblocks the **openhands** agent for two SkillsBench model combos that were previously impossible to run end-to-end, and makes token/cost telemetry **consistent and comparable across providers**. It is **5 functional fixes + 1 test alignment**, each independently scoped so that **every other model, provider, and harness keeps byte-identical behavior**.

Both target combos now produce a **clean `reward 1.0`** on Daytona (`latex-formula-extraction`, with skills), with full skill loading, complete token/meta telemetry, and 0 leakage — and both pass `/benchflow-experiment-review`.

| # | Unblocks | One-line problem | One-line fix | Gate |
|---|----------|------------------|--------------|------|
| 1 | Opus 4.8 MAX + openhands + Bedrock — **Daytona** | litellm sends legacy `thinking.type=enabled`; Opus 4.8 rejects → `ACP -32603` | pin `litellm==1.88.0rc1` + a regex-gated litellm shim in the sandbox | model-id regex |
| 2 | Opus 4.8 MAX + openhands + Bedrock — **Docker** | proxy path bypasses the sandbox shim → 0 reasoning | inject adaptive+max thinking in benchflow's Bedrock proxy | model-id regex |
| 3 | Gemini 3.5 Flash + openhands — **usage tracking** | litellm builds a malformed cache URL → 404 → telemetry forced off | rewrite the Gemini upstream path in the usage proxy | Google host |
| 4 | Heavy tasks on tight sandboxes — **verifier ENOSPC** | verifier dep-install overflows Daytona's 10 GB cap → infra loss, not a real score | free only re-downloadable caches before the verifier | `main` + path-scoped |
| 5 | **Apples-to-apples** cost/usage across providers | `n_input_tokens` means different things per provider; Gemini cache double-charged | normalize input to *total incl. cache*; price cache as a subset | provider-shape, not model |

---

## Why this PR exists

Two things were blocking SkillsBench coverage and trustworthy cost reporting:

1. **The openhands agent could not run Opus 4.8 (MAX thinking) on Bedrock, nor Gemini with usage tracking on.** These are first-class combos for the leaderboard; both failed for *infrastructure* reasons, not model quality.
2. **Even when runs succeeded, the token numbers were not comparable across providers** — and one provider (Gemini) was being **double-charged** in the cost estimate. Any cross-model cost comparison built on these numbers would be wrong.

Every fix below is deliberately narrow: gated by a model-id regex, a target host, a filesystem path, or a provider response-shape — never a broad behavior change.

---

## Fix 1 — Opus 4.8 MAX + openhands + Bedrock on **Daytona**

> **Problem.** Launching this combo fails immediately with `ACP -32603`; the agent never produces a turn.

> **Root cause.** openhands talks to Bedrock through **litellm**. litellm's `AnthropicConfig._is_adaptive_thinking_model` only recognizes the *direct-Anthropic* id `claude-opus-4-8` as an adaptive-thinking model — **not** the Bedrock inference-profile ids (e.g. `us.anthropic.claude-opus-4-8-...`). So litellm's Converse transform emits the **legacy** `thinking.type=enabled` block. Opus 4.8 on Bedrock only accepts the **new** `thinking.type=adaptive` + `output_config.effort=max` shape, so Bedrock rejects the request and the ACP bridge surfaces `-32603`.

> **Fix.**
> - `uv tool install` for openhands now pins **`--with 'litellm==1.88.0rc1'`** — stable `1.86.x` lacks the `max`-effort + `output_config` code path entirely.
> - A small, **regex-gated** litellm monkeypatch shim, [`oh_bedrock_opus_patch.py`](src/benchflow/agents/oh_bedrock_opus_patch.py), is base64-shipped into the openhands tool-venv and auto-loaded via a `zz_oh_bedrock_opus_patch.pth` file. For models matching `claude-(opus|sonnet|haiku)-4-(8|9|1\d)` it:
>   - forces `AnthropicConfig._is_adaptive_thinking_model` → `True`,
>   - coerces `AmazonConverseConfig._handle_reasoning_effort_parameter` → `"max"`,
>   - flips the litellm cost-map `supports_adaptive_thinking` flag.
> - For every non-matching model the shim is a **no-op**.

> **Scope / safety.** Regex-gated to `4-(8|9|1\d)`; any other model's payload and code path is unchanged. The `litellm==1.88.0rc1` pin lives only inside the openhands sandbox venv.

> **Evidence.** ✅ Daytona `latex-formula-extraction` (with skills): **`reward 1.0`**, **107 reasoning references (MAX thinking demonstrably active)** where the same run previously died at `-32603`.

**Files:** [`registry.py`](src/benchflow/agents/registry.py), [`oh_bedrock_opus_patch.py`](src/benchflow/agents/oh_bedrock_opus_patch.py)

---

## Fix 2 — Opus 4.8 MAX + openhands + Bedrock on **Docker**

> **Problem.** The same combo on the VM's Docker sandbox runs, but produces **0 reasoning tokens** — MAX thinking silently never engages.

> **Root cause.** On Docker, Bedrock traffic is routed through **benchflow's host-side proxy**, not the in-sandbox litellm. So Fix 1's sandbox shim never executes on this path — the legacy thinking block goes out unmodified.

> **Fix.** `_force_bedrock_adaptive_thinking_for_opus_4_8(payload)` in [`bedrock_runtime.py`](src/benchflow/providers/bedrock_runtime.py) injects
> ```
> additionalModelRequestFields = { thinking: {type: "adaptive"}, output_config: {effort: "max"} }
> ```
> for matching models, in **both** converse builders (`anthropic_request_to_bedrock_converse` and `openai_responses_request_to_bedrock_converse`), gated by `_BEDROCK_ADAPTIVE_THINKING_RE` (same `claude-(opus|sonnet|haiku)-4-(8|9|1\d)` regex as Fix 1).

> **Scope / safety.** Regex-gated; for non-4.8 models the converse payload is **byte-identical** to before (covered by a unit test).

> **Evidence.** ✅ Proxy emits the adaptive+max block for 4.8 and is byte-identical for other models; the Docker path produces reasoning where it produced none before.

**Files:** [`bedrock_runtime.py`](src/benchflow/providers/bedrock_runtime.py)

---

## Fix 3 — Gemini 3.5 Flash + openhands **usage tracking**

> **Problem.** Enabling usage tracking with Gemini returns **404**, which kills the run. The only workaround was `--usage-tracking off` — i.e. **no token telemetry at all** for Gemini.

> **Root cause.** When litellm is pointed at a custom `api_base` (benchflow's sandbox usage proxy), its URL builder constructs a **malformed context-cache URL**: it emits `/models/{model}:cachedContents` — which (a) **omits the required `/v1beta` version prefix** and (b) invents a bogus *per-model* `:cachedContents` action instead of Google's real **top-level** `/v1beta/cachedContents` resource. Google returns 404.

> **Fix.** [`sandbox_usage_proxy.js`](src/benchflow/providers/assets/sandbox_usage_proxy.js) adds an `isGeminiTarget` gate (host `generativelanguage.googleapis.com`) and `normalizeGeminiUpstreamPath`, which rewrites `^/models/{m}:cachedContents` → `/cachedContents` and prepends `/v1beta` (→ `/v1beta/cachedContents`), leaving already-versioned `/v1beta/…` and `/v1/…` paths untouched.

> **Scope / safety.** Gated on the Google host only — **openai / anthropic / bedrock forwarding is untouched**. Crucially, this **keeps prompt caching ON**, so cache tokens are still measured (turning caching off would have *silently corrupted* cache-token accounting).

> **Evidence.** ✅ Gemini run with **usage tracking enabled** (impossible before): `reward 1.0` (5/5 exact), **1.49M tokens incl. 455K cache-read** — real caching preserved and now correctly measured.

**Files:** [`sandbox_usage_proxy.js`](src/benchflow/providers/assets/sandbox_usage_proxy.js)

---

## Fix 4 — Verifier `ENOSPC` on disk-constrained sandboxes

> **Problem.** Heavy tasks (playwright + marker + ~5 HF models + verifier deps) fail at **verifier dependency install** with `No space left on device`. The run is lost to *infrastructure*, not scored on merit.

> **Root cause.** These tasks overflow Daytona's **hard 10 GB-per-sandbox cap** during verifier setup. The cap cannot be raised — a 20 GB request is rejected by Daytona outright.

> **Fix.** In [`lockdown.py`](src/benchflow/sandbox/lockdown.py)'s `harden_before_verify` (i.e. **before** the verifier runs), best-effort free **only re-downloadable caches**: `~/.cache/uv`, `~/.cache/pip`, `~/.cache/uv_build`, `/tmp/uv-*`, and apt `.deb` archives. Wrapped in `try/except` with a `debug` log so it can never break a run.

> **Scope / safety.** **Result-neutral**: the verifier still installs and runs identically (it just re-fetches if needed). It runs **on `main` only**, matching the existing hardening policy, and **never touches** workspace, outputs, tools, assets, or the agent's deliverable (e.g. `/root/*.md`). A unit test pins the cache-only scope.

> **Evidence.** ✅ A Gemini verifier run that previously failed `No space left on device` now installs + runs cleanly — **0 disk errors, real comparison table produced**; the reward-1 run confirms the agent deliverable survives the cleanup.

**Files:** [`lockdown.py`](src/benchflow/sandbox/lockdown.py)

---

## Fix 5 — Apples-to-apples token accounting (+ Gemini cost double-charge)

> **Problem.** `n_input_tokens` **meant different things for different providers**, so cross-provider usage/cost was not comparable. A cache-heavy agent run surfaced the nonsensical
> ```
> in 124 · out 41,715 · cache_read 3,118,030 · total 3,581,987
> ```
> — input understated by ~25,000×. Separately, **Gemini was being double-charged** in the cost estimate.

> **Root cause.**
> - **Additive providers** — Anthropic Messages (`cache_read_input_tokens`) and Bedrock Converse (`cacheReadInputTokens`) — report `input_tokens`/`inputTokens` as the **uncached delta**; cache reads/writes are **separate fields you must add on**.
> - **Inclusive providers** — OpenAI (`*_tokens_details.cached_tokens`) and Gemini (`cachedContentTokenCount`) — report the **cache-inclusive total**, with cache as a **subset** of it.
> - benchflow did not distinguish the two, so additive-provider input looked tiny.
> - **Gemini double-charge:** the old `_cache_tokens_are_input_breakdown` probe only inspected OpenAI's `*_tokens_details`; it never looked at Gemini's `cachedContentTokenCount`. So Gemini's cache was billed at **both** the full input rate **and** the cache-read rate.

> **Fix.** [`_exchange_token_usage`](src/benchflow/trajectories/types.py) is rewritten so `input_tokens` **uniformly means "total input the model processed, cache included"**:
> - fold the **additive** cache (Anthropic/Bedrock) into input;
> - leave the **inclusive** cache (OpenAI/Gemini) alone — it is already counted;
> - keep `cache_read` / `cache_creation` **broken out as subsets of input** for pricing;
> - fold Gemini `thoughtsTokenCount` (reasoning, billed as output) into **output**, matching Anthropic/OpenAI which already include reasoning in output;
> - `total_tokens` = `input + output` (cache no longer re-added — it now lives *inside* input), preserving any provider-reported total when present.
>
> [`extract_usage`](src/benchflow/providers/usage_proxy_runtime.py) now always prices with `cache_tokens_included_in_input=True` (full rate on `input − cache`, cache at cache rates), which **fixes the Gemini double-charge**. The now-dead probe is removed.

> **Scope / safety.** This is **provider-shape driven, not model-gated**. `total_tokens` is **unchanged for every shape** (it equalled `input + output + cache` before and `input + output` now — identical, since cache moved *inside* input). The cache breakdown's meaning is unchanged. A new test pins the invariant across **all five provider shapes**.

> **Evidence.** ✅ The flagged aggregate normalizes correctly, input now ≫ output (as expected for an agent run), total preserved:
> ```
> BEFORE:  in 124       · out 41,715 · cache_read 3,118,030                          · total 3,581,987
> AFTER:   in 3,540,272 · out 41,715 · cache_read 3,118,030 · cache_creation 422,118 · total 3,581,987
> ```

**Files:** [`trajectories/types.py`](src/benchflow/trajectories/types.py), [`usage_proxy_runtime.py`](src/benchflow/providers/usage_proxy_runtime.py)

---

## Test alignment (not a behavior change)

When Fix 1 added the `litellm` pin to the openhands install command, `tests/test_agent_registry.py` was updated but the **sibling assertion** in [`tests/test_agent_setup.py`](tests/test_agent_setup.py) was missed, leaving it red on the branch. This commit aligns that one assertion with the actual command. No production code changes.

---

## How each fix is scoped to be a no-op everywhere else

| Fix | Activation gate | What is guaranteed unchanged |
|-----|-----------------|------------------------------|
| 1 | model id matches `claude-(opus\|sonnet\|haiku)-4-(8\|9\|1\d)` | every other model's litellm path; pin is sandbox-local |
| 2 | same model-id regex, in the Bedrock proxy | non-4.8 converse payloads are byte-identical (unit-tested) |
| 3 | target host `generativelanguage.googleapis.com` | openai/anthropic/bedrock forwarding; prompt caching stays ON |
| 4 | runs on `main`, deletes only re-downloadable cache paths | workspace / outputs / tools / assets / agent deliverables |
| 5 | provider response-shape (additive vs inclusive cache) | `total_tokens` for every shape; cache breakdown semantics |

---

## End-to-end validation

Both combos: Daytona, `latex-formula-extraction`, **with task skills loaded**, audited via `/benchflow-experiment-review`.

| Combo | Reward | Thinking / usage proof | Tokens (cache-read) | Skills | Leakage |
|-------|--------|------------------------|---------------------|--------|---------|
| **Opus 4.8 MAX + openhands + Bedrock** | ✅ `1.0` | 107 reasoning refs (MAX active); `usage_source=provider_response` | 3.58M (3.1M) | `task_skills_loading=1` | 0 |
| **Gemini 3.5 Flash + openhands + Gemini key** | ✅ `1.0` (5/5 exact) | usage tracking **enabled** (impossible before Fix 3) | 1.49M (455K) | `task_skills_loading=1` | 0 |

Additional proof:
- **Fix 4:** a verifier run that previously hit `No space left on device` now installs + runs cleanly (0 disk errors, real comparison table).
- **Fix 5:** flagged Opus aggregate `in 124 → 3,540,272` (input ≫ output, total preserved).
- **Both** runs pass `/benchflow-experiment-review` as clean reward-1 runs.

**Static + unit:** `ruff check`, `ruff format`, and `ty check` all clean. Unit coverage added/updated: shim no-op for non-4.8; proxy converse byte-identical for non-4.8; Gemini path-rewrite; disk-reclaim cache-scope; cross-provider token-normalization invariant across Anthropic / Bedrock / OpenAI-Responses / OpenAI-Chat / Gemini. **Full suite: 2598 passed, 0 failures.**

---

## Files changed (14)

```
src/benchflow/agents/oh_bedrock_opus_patch.py       +58   (new)  Fix 1 — litellm shim
src/benchflow/agents/registry.py                    +21          Fix 1 — pin + shim deploy
src/benchflow/providers/bedrock_runtime.py          +25          Fix 2 — proxy adaptive+max
src/benchflow/providers/assets/sandbox_usage_proxy.js +28        Fix 3 — Gemini path rewrite
src/benchflow/sandbox/lockdown.py                   +21          Fix 4 — pre-verify cache reclaim
src/benchflow/trajectories/types.py                 +/-87        Fix 5 — token normalization
src/benchflow/providers/usage_proxy_runtime.py      +/-16        Fix 5 — cost: cache as subset
tests/test_agent_registry.py                        +10          Fix 1 tests
tests/test_usage_proxy_gemini_rewrite.py            +85   (new)  Fix 3 tests
tests/test_sandbox_hardening.py                     +26          Fix 4 tests
tests/test_token_usage_normalization.py             +140  (new)  Fix 5 invariant (5 shapes)
tests/test_usage_proxy.py                           +/-6         Fix 5 assertion (150→163)
tests/test_daytona_usage_runtime.py                 +/-5         Fix 5 assertion (34→334)
tests/test_agent_setup.py                           +1           test alignment
```

## Commit map

| Commit | Contents |
|--------|----------|
| `33c1c98` | Fixes 1 + 2 + 3 (Bedrock Opus 4.8 thinking, Daytona & Docker; Gemini usage) |
| `08630a3` | Fix 4 (free re-downloadable caches before the verifier) |
| `6e455cb` | Fix 5 (token normalization + Gemini cost double-charge) |
| `938fc1d` | Test alignment (install-command assertion) |

---

## Caveats & follow-ups

- **`litellm==1.88.0rc1` is a pre-release.** Stable `1.86.x` lacks the `max`-effort + `output_config` path. This self-resolves once litellm ships the Bedrock `supports_adaptive_thinking` metadata upstream — at which point the pin and shim can be dropped. (Sandbox-local only; verified Gemini still serves on it.)
- The shim deploy into the openhands tool-venv is **best-effort** (logged, never fatal); the proxy path (Fix 2) is the Docker safety net for the same goal.

## Suggested review order

1. **Fix 5** ([`trajectories/types.py`](src/benchflow/trajectories/types.py), [`usage_proxy_runtime.py`](src/benchflow/providers/usage_proxy_runtime.py)) — the only change touching shared accounting; [`test_token_usage_normalization.py`](tests/test_token_usage_normalization.py) documents the exact intended semantics per provider.
2. **Fixes 1–2** (the regex gate is identical in both; confirm non-4.8 is untouched).
3. **Fix 3** (host gate + path rewrite) and **Fix 4** (path-scoped cache reclaim) — both narrowly gated and unit-pinned.

### DIFF
--- a/src/benchflow/agents/oh_bedrock_opus_patch.py
+++ b/src/benchflow/agents/oh_bedrock_opus_patch.py
@@ -1,0 +1,58 @@
+# Deployed into the openhands sandbox's litellm (via registry.py install_cmd) to make
+# Bedrock Claude Opus/Sonnet/Haiku 4.8+ use ADAPTIVE thinking at MAX effort.
+#
+# litellm flags only the direct-anthropic id (`claude-opus-4-8`) as adaptive-thinking,
+# NOT the Bedrock inference-profile ids (`us./eu./global./bare anthropic.claude-opus-4-8`).
+# So on Daytona (direct-bedrock routing) the converse transform emits the legacy
+# `thinking.type=enabled` that Bedrock 4.8 rejects (ACP -32603 / 400 ValidationException).
+#
+# This shim is regex-gated to 4.8+ ids => a strict no-op for every other model
+# (gemini, azure, opus-4.6, etc. import-and-call straight through to the originals).
+# It is never imported by benchflow itself; it is base64-shipped and auto-loaded in the
+# sandbox via a .pth file. Requires litellm>=1.88.0rc1 (the `max` effort + output_config path).
+import re
+
+_NEW = re.compile(r"claude-(opus|sonnet|haiku)-4-(8|9|1\d)")
+
+# (1) gate: treat 4.8+ as adaptive-thinking models
+try:
+    from litellm.llms.anthropic.chat.transformation import AnthropicConfig
+
+    _orig_gate = AnthropicConfig._is_adaptive_thinking_model
+
+    def _gate(model):
+        try:
+            if model and _NEW.search(model.lower()):
+                return True
+        except Exception:
+            pass
+        return _orig_gate(model)
+
+    AnthropicConfig._is_adaptive_thinking_model = staticmethod(_gate)
+except Exception:
+    pass
+
+# (2) force MAX effort for 4.8+ (openhands caps its own effort at xhigh)
+try:
+    from litellm.llms.bedrock.chat.converse_transformation import AmazonConverseConfig
+
+    _orig_handle = AmazonConverseConfig._handle_reasoning_effort_parameter
+
+    def _handle(self, model, reasoning_effort, optional_params):
+        if model and _NEW.search(model.lower()):
+            reasoning_effort = "max"
+        return _orig_handle(self, model, reasoning_effort, optional_params)
+
+    AmazonConverseConfig._handle_reasoning_effort_parameter = _handle
+except Exception:
+    pass
+
+# (3) belt-and-suspenders: flip the cost-map flag too
+try:
+    import litellm
+
+    for _k in list(litellm.model_cost):
+        if _NEW.search(_k.lower()):
+            litellm.model_cost[_k]["supports_adaptive_thinking"] = True
+except Exception:
+    pass
--- a/src/benchflow/agents/oh_bedrock_opus_patch.py
+++ b/src/benchflow/agents/oh_bedrock_opus_patch.py
@@ -1,5 +1,5 @@
 # Deployed into the openhands sandbox's litellm (via registry.py install_cmd) to make
-# Bedrock Claude Opus/Sonnet/Haiku 4.8+ use ADAPTIVE thinking at MAX effort.
+# Bedrock Claude Opus/Sonnet/Haiku 4.8+ use the ADAPTIVE thinking contract.
 #
 # litellm flags only the direct-anthropic id (`claude-opus-4-8`) as adaptive-thinking,
 # NOT the Bedrock inference-profile ids (`us./eu./global./bare anthropic.claude-opus-4-8`).
@@ -9,12 +9,22 @@
 # This shim is regex-gated to 4.8+ ids => a strict no-op for every other model
 # (gemini, azure, opus-4.6, etc. import-and-call straight through to the originals).
 # It is never imported by benchflow itself; it is base64-shipped and auto-loaded in the
-# sandbox via a .pth file. Requires litellm>=1.88.0rc1 (the `max` effort + output_config path).
+# sandbox via a .pth file. Requires litellm>=1.88.0rc1 (the effort + output_config path).
+#
+# Effort is NOT hardcoded: it is taken from the BENCHFLOW_BEDROCK_THINKING_EFFORT env
+# (MAX mode sets it to `max`); unset => the agent's requested effort passes through.
+# The model matcher mirrors providers/bedrock_runtime.BEDROCK_ADAPTIVE_THINKING_RE
+# (kept identical; tests/test_bedrock_thinking.py pins parity).
+import os
 import re
 
-_NEW = re.compile(r"claude-(opus|sonnet|haiku)-4-(8|9|1\d)")
+_NEW = re.compile(r"claude-(?:opus|sonnet|haiku)-4-(?:8|9|1\d)(?!\d)")
 
-# (1) gate: treat 4.8+ as adaptive-thinking models
+_EFFORT_ENV = "BENCHFLOW_BEDROCK_THINKING_EFFORT"
+_VALID_EFFORTS = {"minimal", "low", "medium", "high", "xhigh", "max"}
+
+# (1) gate: treat 4.8+ as adaptive-thinking models (the load-bearing fix — Bedrock
+# regional ids lack the registry capability flag, so the legacy shape is rejected)
 try:
     from litellm.llms.anthropic.chat.transformation import AnthropicConfig
 
@@ -32,7 +42,8 @@ try:
 except Exception:
     pass
 
-# (2) force MAX effort for 4.8+ (openhands caps its own effort at xhigh)
+# (2) config-gated effort for 4.8+: override ONLY when BENCHFLOW_BEDROCK_THINKING_EFFORT
+# is set (e.g. MAX mode => `max`); otherwise the agent's requested effort is honored.
 try:
     from litellm.llms.bedrock.chat.converse_transformation import AmazonConverseConfig
 
@@ -40,7 +51,9 @@ try:
 
     def _handle(self, model, reasoning_effort, optional_params):
         if model and _NEW.search(model.lower()):
-            reasoning_effort = "max"
+            _override = os.environ.get(_EFFORT_ENV, "").strip().lower()
+            if _override in _VALID_EFFORTS:
+                reasoning_effort = _override
         return _orig_handle(self, model, reasoning_effort, optional_params)
 
     AmazonConverseConfig._handle_reasoning_effort_parameter = _handle
--- a/src/benchflow/agents/registry.py
+++ b/src/benchflow/agents/registry.py
@@ -551,8 +551,14 @@ AGENTS: dict[str, AgentConfig] = {
             "  fi && "
             "uv tool install --force --refresh "
             "--with 'boto3>=1.40' "
-            # Bedrock Opus 4.8+ needs litellm's adaptive-thinking `max` effort + output_config
-            # path (absent from stable 1.86.x). Inert for other models. See oh_bedrock_opus_patch.py.
+            # Pin litellm to the prerelease carrying the Bedrock adaptive-thinking
+            # effort + output_config path (absent from stable 1.86.x), required for
+            # Claude Opus 4.8+. NOTE: this is a DELIBERATE, OpenHands-wide litellm
+            # bump — it changes the litellm version for every provider/model this
+            # OpenHands install talks to, not just Bedrock 4.8. The bundled shim is
+            # regex-gated and inert elsewhere, but the version itself is global.
+            # Revisit/drop once the fix ships in a stable litellm. See
+            # oh_bedrock_opus_patch.py.
             "--with 'litellm==1.88.0rc1' "
             "--from 'git+https://github.com/OpenHands/OpenHands-CLI.git@main' "
             "openhands --python 3.12 && "
@@ -564,17 +570,32 @@ AGENTS: dict[str, AgentConfig] = {
             "mkdir -p ~/.openhands && "
             'echo \'{"llm":{"model":"placeholder","api_key":"placeholder"}}\' '
             "> ~/.openhands/agent_settings.json && "
-            # Deploy the Bedrock Opus 4.8+ adaptive-thinking shim into the sandbox litellm.
-            # Best-effort (the trailing `true` keeps it from ever failing the install); the
-            # shim is regex-gated internally => a strict no-op for every non-4.8 model.
-            "( SP=\"$(ls -d \"$(uv tool dir)\"/openhands/lib/python*/site-packages "
-            "2>/dev/null | head -1)\"; "
-            "[ -n \"$SP\" ] && "
+            # Deploy the Bedrock Opus 4.8+ adaptive-thinking shim into the sandbox
+            # litellm, then BEHAVIORALLY self-test it. On Daytona (direct Bedrock,
+            # no host proxy) this shim is the ONLY mechanism, so a silent deploy
+            # failure would regress Claude 4.8 thinking to the rejected legacy
+            # shape. The install stays non-fatal (other providers must still
+            # install) but emits a distinct ACTIVE / NOT-active marker so a failure
+            # is visible instead of swallowed. Regex-gated => a no-op for non-4.8.
+            '( SP="$(ls -d "$(uv tool dir)"/openhands/lib/python*/site-packages '
+            '2>/dev/null | head -1)"; '
+            '[ -n "$SP" ] && '
             f"echo '{_OH_BEDROCK_OPUS_SHIM_B64}' | base64 -d "
-            "> \"$SP/oh_bedrock_opus_patch.py\" && "
+            '> "$SP/oh_bedrock_opus_patch.py" && '
             "printf 'import oh_bedrock_opus_patch\\n' "
-            "> \"$SP/zz_oh_bedrock_opus_patch.pth\" "
+            '> "$SP/zz_oh_bedrock_opus_patch.pth" '
             "|| echo 'benchflow: opus-4.8 bedrock thinking shim NOT deployed' >&2; "
+            # The .pth auto-imports the shim at interpreter startup, so the venv
+            # python must now classify a regional 4.8 id as adaptive-thinking.
+            'PY="$(uv tool dir)/openhands/bin/python"; '
+            '[ -x "$PY" ] && "$PY" -c '
+            "'import sys; from litellm.llms.anthropic.chat.transformation "
+            "import AnthropicConfig as C; "
+            "sys.exit(0 if C._is_adaptive_thinking_model"
+            '("us.anthropic.claude-opus-4-8") else 1)\' 2>/dev/null '
+            "&& echo 'benchflow: opus-4.8 bedrock thinking shim ACTIVE' >&2 "
+            "|| echo 'benchflow: opus-4.8 bedrock thinking shim NOT active"
+            " (Daytona Bedrock 4.8 thinking will regress)' >&2; "
             "true ) && "
             "command -v openhands >/dev/null 2>&1"
         ),
--- a/src/benchflow/agents/registry.py
+++ b/src/benchflow/agents/registry.py
@@ -194,6 +194,12 @@ _PI_LAUNCHER = (Path(__file__).parent / "pi_acp_launcher.py").read_text()
 # Path to the Harvey LAB ACP shim (runs Harvey LAB harness as an ACP agent)
 _HARVEY_LAB_SHIM = (Path(__file__).parent / "harvey_lab_acp_shim.py").read_text()
 
+# Bedrock Opus 4.8+ adaptive-thinking shim, base64-shipped into the openhands sandbox's
+# litellm (source: oh_bedrock_opus_patch.py). Regex-gated => a no-op for every other model.
+_OH_BEDROCK_OPUS_SHIM_B64 = base64.b64encode(
+    (Path(__file__).parent / "oh_bedrock_opus_patch.py").read_text().encode()
+).decode()
+
 
 def _json_settings_merge(path: str, mutator: str) -> str:
     """Idempotent JSON-settings merge as a one-line bash snippet."""
@@ -545,6 +551,9 @@ AGENTS: dict[str, AgentConfig] = {
             "  fi && "
             "uv tool install --force --refresh "
             "--with 'boto3>=1.40' "
+            # Bedrock Opus 4.8+ needs litellm's adaptive-thinking `max` effort + output_config
+            # path (absent from stable 1.86.x). Inert for other models. See oh_bedrock_opus_patch.py.
+            "--with 'litellm==1.88.0rc1' "
             "--from 'git+https://github.com/OpenHands/OpenHands-CLI.git@main' "
             "openhands --python 3.12 && "
             "  uv tool list | grep -q '^openhands\\b' ) && "
@@ -555,6 +564,18 @@ AGENTS: dict[str, AgentConfig] = {
             "mkdir -p ~/.openhands && "
             'echo \'{"llm":{"model":"placeholder","api_key":"placeholder"}}\' '
             "> ~/.openhands/agent_settings.json && "
+            # Deploy the Bedrock Opus 4.8+ adaptive-thinking shim into the sandbox litellm.
+            # Best-effort (the trailing `true` keeps it from ever failing the install); the
+            # shim is regex-gated internally => a strict no-op for every non-4.8 model.
+            "( SP=\"$(ls -d \"$(uv tool dir)\"/openhands/lib/python*/site-packages "
+            "2>/dev/null | head -1)\"; "
+            "[ -n \"$SP\" ] && "
+            f"echo '{_OH_BEDROCK_OPUS_SHIM_B64}' | base64 -d "
+            "> \"$SP/oh_bedrock_opus_patch.py\" && "
+            "printf 'import oh_bedrock_opus_patch\\n' "
+            "> \"$SP/zz_oh_bedrock_opus_patch.pth\" "
+            "|| echo 'benchflow: opus-4.8 bedrock thinking shim NOT deployed' >&2; "
+            "true ) && "
             "command -v openhands >/dev/null 2>&1"
         ),
         launch_cmd=(
--- a/src/benchflow/providers/assets/sandbox_usage_proxy.js
+++ b/src/benchflow/providers/assets/sandbox_usage_proxy.js
@@ -15,6 +15,14 @@ function getConfig(name, argName) {
 }
 
 const target = new URL(getConfig("TARGET", "target").replace(/\/+$/, ""));
+// Google AI Studio (Gemini) needs upstream path normalization. With a custom api_base
+// (this proxy), litellm builds gemini URLs as `{api_base}/models/{model}:{action}` — it
+// omits the required `/v1beta` version prefix, and for context caching it emits the bogus
+// `:cachedContents` model-action instead of Google's real top-level `/v1beta/cachedContents`
+// collection (litellm vertex_llm_base._check_custom_proxy, still present in 1.88.x). Without
+// this rewrite every gemini call — especially the prompt-cache probe — 404s through the proxy.
+// Gated on the gemini host, so all other providers (openai/anthropic/bedrock) are untouched.
+const isGeminiTarget = /(^|\.)generativelanguage\.googleapis\.com$/i.test(target.hostname);
 const statePath = getConfig("STATE_PATH", "state");
 const logPath = getConfig("LOG_PATH", "log");
 const pidPath = getConfig("PID_PATH", "pid");
@@ -89,10 +97,26 @@ function bodyB64(chunks) {
   return Buffer.concat(chunks).toString("base64");
 }
 
+function normalizeGeminiUpstreamPath(pathWithQuery) {
+  const qIdx = pathWithQuery.indexOf("?");
+  let path = qIdx === -1 ? pathWithQuery : pathWithQuery.slice(0, qIdx);
+  const query = qIdx === -1 ? "" : pathWithQuery.slice(qIdx);
+  // already version-prefixed -> leave as-is
+  if (path.startsWith("/v1beta/") || path.startsWith("/v1/")) return pathWithQuery;
+  // litellm's bogus per-model cache action -> Google's real top-level collection
+  path = path.replace(/^\/models\/[^/]+:cachedContents\b/, "/cachedContents");
+  // all Google AI Studio resources live under /v1beta
+  if (!path.startsWith("/v1beta/")) path = `/v1beta${path}`;
+  return `${path}${query}`;
+}
+
 function upstreamPath(requestUrl) {
+  const requestPath = isGeminiTarget
+    ? normalizeGeminiUpstreamPath(requestUrl)
+    : requestUrl;
   const basePath = target.pathname.replace(/\/+$/, "");
-  if (!basePath) return requestUrl;
-  return `${basePath}${requestUrl.startsWith("/") ? requestUrl : `/${requestUrl}`}`;
+  if (!basePath) return requestPath;
+  return `${basePath}${requestPath.startsWith("/") ? requestPath : `/${requestPath}`}`;
 }
 
 function maybeApplyPromptCacheRetention(requestUrl, headers, body) {
--- a/src/benchflow/providers/assets/sandbox_usage_proxy.js
+++ b/src/benchflow/providers/assets/sandbox_usage_proxy.js
@@ -97,6 +97,9 @@ function bodyB64(chunks) {
   return Buffer.concat(chunks).toString("base64");
 }
 
+// Keep in sync with the Python source of truth:
+// benchflow/trajectories/gemini_paths.py:normalize_gemini_upstream_path
+// (tests/test_gemini_path_normalization.py pins parity across both runtimes).
 function normalizeGeminiUpstreamPath(pathWithQuery) {
   const qIdx = pathWithQuery.indexOf("?");
   let path = qIdx === -1 ? pathWithQuery : pathWithQuery.slice(0, qIdx);
--- a/src/benchflow/providers/bedrock_runtime.py
+++ b/src/benchflow/providers/bedrock_runtime.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 
 import base64
 import json
+import re
 import time
 from typing import Any
 
@@ -300,6 +301,28 @@ def _tools_to_bedrock_tool_config(
     return config
 
 
+# Bedrock Claude Opus/Sonnet/Haiku 4.8+ require the adaptive-thinking contract
+# (thinking.type=adaptive + output_config.effort) and reject the legacy enabled form.
+_BEDROCK_ADAPTIVE_THINKING_RE = re.compile(r"claude-(opus|sonnet|haiku)-4-(8|9|1\d)")
+
+
+def _force_bedrock_adaptive_thinking_for_opus_4_8(payload: dict[str, Any]) -> None:
+    """Force Bedrock Claude 4.8+ to adaptive thinking at MAX effort.
+
+    On Docker, openhands→litellm reaches Bedrock through this host proxy via the
+    Anthropic-Messages / OpenAI-Responses path, which otherwise drops the incoming
+    thinking block (=> 0 extended thinking) — and Bedrock 4.8 rejects the legacy
+    ``thinking.type=enabled`` regardless. Regex-gated on the resolved model id, so
+    every other model's payload is byte-identical to before.
+    """
+    model = str(payload.get("modelId", "")).lower()
+    if not _BEDROCK_ADAPTIVE_THINKING_RE.search(model):
+        return
+    fields = payload.setdefault("additionalModelRequestFields", {})
+    fields["thinking"] = {"type": "adaptive"}
+    fields["output_config"] = {"effort": "max"}
+
+
 def anthropic_request_to_bedrock_converse(body: dict[str, Any]) -> dict[str, Any]:
     """Translate an Anthropic Messages request to Converse kwargs."""
     payload: dict[str, Any] = {
@@ -325,6 +348,7 @@ def anthropic_request_to_bedrock_converse(body: dict[str, Any]) -> dict[str, Any
     )
     if tool_config:
         payload["toolConfig"] = tool_config
+    _force_bedrock_adaptive_thinking_for_opus_4_8(payload)
     return payload
 
 
@@ -421,6 +445,7 @@ def openai_responses_request_to_bedrock_converse(
     )
     if tool_config:
         payload["toolConfig"] = tool_config
+    _force_bedrock_adaptive_thinking_for_opus_4_8(payload)
     return payload
 
 
--- a/src/benchflow/providers/bedrock_runtime.py
+++ b/src/benchflow/providers/bedrock_runtime.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 
 import base64
 import json
+import os
 import re
 import time
 from typing import Any
@@ -303,24 +304,65 @@ def _tools_to_bedrock_tool_config(
 
 # Bedrock Claude Opus/Sonnet/Haiku 4.8+ require the adaptive-thinking contract
 # (thinking.type=adaptive + output_config.effort) and reject the legacy enabled form.
-_BEDROCK_ADAPTIVE_THINKING_RE = re.compile(r"claude-(opus|sonnet|haiku)-4-(8|9|1\d)")
+# Canonical matcher — keep in sync with agents/oh_bedrock_opus_patch.py (the Daytona
+# sandbox shim, which cannot import benchflow); tests/test_bedrock_thinking.py pins
+# parity. The trailing ``(?!\d)`` anchors the minor version so ``claude-opus-4-80``
+# (a different model) does not match ``4-8``.
+BEDROCK_ADAPTIVE_THINKING_RE = re.compile(
+    r"claude-(?:opus|sonnet|haiku)-4-(?:8|9|1\d)(?!\d)"
+)
+_BEDROCK_ADAPTIVE_THINKING_RE = BEDROCK_ADAPTIVE_THINKING_RE  # backwards-compat alias
+
+# Explicit BenchFlow override for the adaptive-thinking effort level. Unset =>
+# honor the caller's requested effort (never silently force MAX). Set this (e.g.
+# to ``max``) to run Claude 4.8+ at a fixed effort regardless of the agent's cap.
+BEDROCK_THINKING_EFFORT_ENV = "BENCHFLOW_BEDROCK_THINKING_EFFORT"
+_VALID_BEDROCK_EFFORTS = {"minimal", "low", "medium", "high", "xhigh", "max"}
+_DEFAULT_BEDROCK_EFFORT = "high"
 
 
-def _force_bedrock_adaptive_thinking_for_opus_4_8(payload: dict[str, Any]) -> None:
-    """Force Bedrock Claude 4.8+ to adaptive thinking at MAX effort.
+def _resolve_bedrock_thinking_effort(requested_effort: str | None) -> str:
+    """Effort precedence: explicit BenchFlow override > caller request > default.
+
+    Crucially this does NOT default to ``max`` — forcing MAX on every 4.8+ request
+    was the over-broad behavior; MAX is now opt-in via the override env (which the
+    MAX-mode run config sets) or an explicit caller request.
+    """
+    override = os.environ.get(BEDROCK_THINKING_EFFORT_ENV, "").strip().lower()
+    if override in _VALID_BEDROCK_EFFORTS:
+        return override
+    if requested_effort and requested_effort.strip().lower() in _VALID_BEDROCK_EFFORTS:
+        return requested_effort.strip().lower()
+    return _DEFAULT_BEDROCK_EFFORT
+
+
+def _normalize_bedrock_thinking_for_opus_4_8(
+    payload: dict[str, Any],
+    *,
+    thinking_requested: bool,
+    requested_effort: str | None,
+) -> None:
+    """Convert Bedrock Claude 4.8+ thinking to the adaptive contract it requires.
 
     On Docker, openhands→litellm reaches Bedrock through this host proxy via the
-    Anthropic-Messages / OpenAI-Responses path, which otherwise drops the incoming
-    thinking block (=> 0 extended thinking) — and Bedrock 4.8 rejects the legacy
-    ``thinking.type=enabled`` regardless. Regex-gated on the resolved model id, so
-    every other model's payload is byte-identical to before.
+    Anthropic-Messages / OpenAI-Responses path. Bedrock 4.8+ rejects the legacy
+    ``thinking.type=enabled`` shape, so when the caller actually requests thinking
+    we rewrite it to ``thinking.type=adaptive`` + ``output_config.effort``. The
+    effort is resolved from the request / BenchFlow override — NOT hardcoded to
+    ``max`` — so a non-MAX 4.8 caller is not silently upgraded. No-op when thinking
+    was not requested or the model is not 4.8+; regex-gated => byte-identical
+    otherwise.
     """
     model = str(payload.get("modelId", "")).lower()
-    if not _BEDROCK_ADAPTIVE_THINKING_RE.search(model):
+    if not BEDROCK_ADAPTIVE_THINKING_RE.search(model):
+        return
+    if not thinking_requested:
         return
     fields = payload.setdefault("additionalModelRequestFields", {})
     fields["thinking"] = {"type": "adaptive"}
-    fields["output_config"] = {"effort": "max"}
+    fields["output_config"] = {
+        "effort": _resolve_bedrock_thinking_effort(requested_effort)
+    }
 
 
 def anthropic_request_to_bedrock_converse(body: dict[str, Any]) -> dict[str, Any]:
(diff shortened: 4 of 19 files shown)


bingran-you:
Reviewing this with the thermo-nuclear code-quality bar plus the BenchFlow experiment-review health criteria: I don't think this PR is ready as-is.

My read of the PR intent: unblock OpenHands SkillsBench runs for Bedrock Claude Opus 4.8 MAX thinking on Daytona/Docker, keep Gemini usage tracking on by normalizing LiteLLM's malformed cache path, avoid verifier ENOSPC by reclaiming cache space, and make token/cost accounting comparable across providers. Those are the right problems, but the implementation misses several of the stated scoping and parity guarantees.

## Blockers

1. **Gemini usage tracking is fixed only for one proxy path.**

   `src/benchflow/providers/assets/sandbox_usage_proxy.js:25` gates the rewrite on `generativelanguage.googleapis.com`, and the rewrite exists only in the sandbox-local Node proxy (`src/benchflow/providers/assets/sandbox_usage_proxy.js:100`). The host-side Docker `TrajectoryProxy` still forwards `target + path` directly at `src/benchflow/trajectories/proxy.py:175`, so Docker can still send `/models/{model}:cachedContents` upstream unchanged.

   I also hit this in a real Daytona OpenHands/Gemini smoke attempt from this branch: the captured request stayed `/models/gemini-3.5-flash:cachedContents`, usage stayed unavailable, and the rollout ended as `ACP error -32603` with a partial trajectory. In that environment the upstream target was configured through the AllHands LLM proxy, so the Google-host gate did not fire. That run is not healthy BenchFlow data, but it is exactly the kind of real base-url/proxy composition this PR needs to survive if usage tracking is advertised as unblocked.

   Please move Gemini path normalization to a canonical helper used by both proxy implementations, and test sandbox-local, host-side Docker, versioned base paths, and non-Google proxy hosts that front Gemini.

2. **The pre-verifier disk reclaim can delete agent-visible workspace data.**

   `src/benchflow/sandbox/lockdown.py:880` blindly removes `/root/.cache/{uv,pip,uv_build}` and `/home/*/.cache/{uv,pip,uv_build}`. BenchFlow rollouts can use `/root` as the task workspace; my Daytona smoke also reported `workspace=/root`. In that case this code deletes paths inside the workspace, despite the PR claiming it never touches workspace/output/assets/deliverables.

   This is a verifier-fidelity issue. A task can legitimately create or depend on workspace files under `.cache`, and verifier hardening should not silently mutate completed agent state outside the established workspace-restore policy. Please make the reclaim workspace-aware, or move redownloadable package caches to a dedicated non-workspace location. Add regression coverage for `workspace="/root"` and `workspace="/home/agent"`.

3. **Gemini token normalization still drops input-side tool-use tokens.**

   `src/benchflow/trajectories/types.py:133` normalizes Gemini input from `usageMetadata.promptTokenCount`, but does not include `toolUsePromptTokenCount`. `_USAGE_METADATA_KEYS` at `src/benchflow/trajectories/types.py:28` also omits it, so some Gemini responses can fail provider-usage detection or underreport normalized input/cost even when `totalTokenCount` is preserved. Please include this field in usage detection and input normalization, with a fixture covering prompt + cache + tool-use prompt + candidates + thoughts.

4. **The OpenHands/LiteLLM change is broader and more brittle than described.**

   `src/benchflow/agents/registry.py:552` globally pins every OpenHands install to `litellm==1.88.0rc1`, not just Bedrock Claude 4.8+ runs. That is not byte-identical behavior for every other OpenHands provider/model. Then the load-bearing shim deploy at `src/benchflow/agents/registry.py:570` is best-effort and forced successful by the trailing `true`; if the `.pth` write fails, the install still passes and Daytona Bedrock can regress back to ACP `-32603`.

   Please isolate the dependency/patch to the Bedrock 4.8 path, or make the broader prerelease LiteLLM upgrade an explicit compatibility decision for all OpenHands providers. Also make shim deployment fail closed for the affected model path or assert behaviorally that the installed tool venv actually transforms Bedrock 4.8 to adaptive/max.

5. **Bedrock MAX thinking is forced by model-id regex, not by requested run mode.**

   `src/benchflow/providers/bedrock_runtime.py:321` overwrites `additionalModelRequestFields.thinking` and `output_config.effort=max` for any model id matching `claude-(opus|sonnet|haiku)-4-(8|9|1\d)`. The OpenHands shim does the same at `src/benchflow/agents/oh_bedrock_opus_patch.py:41`. That makes every Claude 4.8+ Bedrock request through these paths pay MAX-thinking behavior even if the caller did not request MAX. Please gate this on the incoming reasoning/thinking request or explicit BenchFlow effort config, and centralize the model-family matcher so the shim/proxy do not drift. The current unanchored regex also matches unintended ids like `claude-opus-4-80`.

## Validation run

- `uv sync --extra dev --locked` passed.
- `uv run python -m pytest tests/` passed: `2561 passed, 49 skipped, 1 deselected`.
- `uv run ty check src/` passed.
- `uv run ruff check .` passed.
- `uv sync --extra dev --extra sandbox-daytona --extra bedrock --locked` passed.
- Focused PR-adjacent suites from the parallel subreviews also passed (`test_agent_registry`, `test_agent_setup`, Bedrock translator/runtime tests, Gemini rewrite tests, sandbox hardening cache test, token normalization tests).
- Real SkillsBench/OpenHands smoke did not produce healthy publishable E2E evidence from my attempt: the Daytona Gemini attempt ended partial with `ACP error -32603` and `usage_source=unavailable`; the `/benchflow-experiment-review` extractor found task skills in the catalog but still marked the run as requiring manual review because the launch shape did not serialize with-skill metadata cleanly. The VM Docker smoke was started against PR head on the GCP Docker host, but I stopped that initial run before counting it because it had not emitted artifacts and was launched from the same initial canary shape rather than the stricter YAML config.

Given the blockers above, I would not merge this until Docker/Daytona proxy parity, workspace-aware cleanup, Gemini accounting coverage, and the OpenHands/Bedrock gating story are tightened and then revalidated with healthy Docker + Daytona SkillsBench canaries.

bingran-you:
Follow-up E2E evidence after the initial review.

I reran the PR-targeted real SkillsBench canaries with the stricter `/benchflow-experiment-review` launch shape: `include_task_skills: true`, `max_retries: 0`, `usage_tracking: required`, OpenHands, task `latex-formula-extraction`, source ref `d1fd798d4d271586267c14abd408c6ec371ba9a4` when using `--source-repo`.

| Model | Sandbox | Outcome | Usage tracking | Skill audit / artifact health |
| --- | --- | --- | --- | --- |
| `gemini-3.5-flash` | Daytona | Healthy model fail: `reward=0`, `error=null`, `verifier_error=null`, `partial_trajectory=false`, 70 tool calls | `usage_source=provider_response`, `telemetry_coverage=1`, input `3,354,486`, output `29,662`, cache read `839,358` | `marker` + `pdf` skills loaded; leakage scan clean. `check_results.py` warns because this local run used `tasks_dir`, so source provenance is missing from result/config/summary. |
| `aws-bedrock/us.anthropic.claude-opus-4-8` | Daytona | Healthy model fail: `reward=0`, `error=null`, `verifier_error=null`, `partial_trajectory=false`, 60 tool calls | `usage_source=provider_response`, input `3,924,927`, output `53,818`, cache read `3,508,414`, cache creation `416,389` | `marker` + `pdf` skills loaded; leakage scan clean. Same local `tasks_dir` provenance warning. |
| `aws-bedrock/us.anthropic.claude-opus-4-8` | Docker on `daytona-orchestrator` VM | PASS: `reward=1.0`, `error=null`, `partial_trajectory=false`, 57 tool calls | `usage_source=provider_response`, `usage_tracking.environment=docker`, `endpoint_kind=host`, `telemetry_coverage=1`, input `3,097,466`, output `35,031` | `marker` + `pdf` skills loaded; leakage scan clean. Source metadata is present, but `check_results.py` still warns `source.resolved_sha is not reachable from source.repo/requested_ref`. |
| `gemini-3.5-flash` | Docker on `daytona-orchestrator` VM | **Unhealthy infra/runtime failure**. Host usage proxy started, OpenHands then died with `ACP error -32603: Internal error`; no `result.json`, no `llm_trajectory.jsonl`, no summary | No usable usage artifact emitted | `check_results.py`: no result files. I killed only the stuck PR test process/container after it failed to emit artifacts. |

This changes the E2E picture: the Bedrock paths are runnable, but Docker + Gemini remains a hard E2E blocker for this PR. Daytona + Gemini can be made healthy only with a clean provider env; the shared `.env` has generic provider variables that can poison OpenHands/Gemini unless stripped. That reinforces the earlier proxy/env-path concern: the PR fixes Gemini path rewriting in the sandbox-local Node proxy, but the Docker host-side `TrajectoryProxy` path still needs equivalent coverage/behavior before this is safe to merge.

bingran-you:
Thanks for both rounds of review — all five blockers are addressed in five focused commits on the branch (`938fc1d..9093d93`). Full suite **2650 passed** (+52 new tests); `ruff` and `ty` clean.

| # | Blocker | Fix | Commit |
|---|---|---|---|
| 1 | Gemini rewrite lived only in the sandbox Node proxy; the Docker host `TrajectoryProxy` forwarded the malformed `:cachedContents` unchanged | New `trajectories/gemini_paths.py` as the Python source of truth, applied in `TrajectoryProxy._upstream_url`; a node↔python parity test pins the two runtimes so they can't drift | `7751455` |
| 2 | Pre-verifier reclaim deleted `/root/.cache` when `workspace=/root` (and `restore_workspace` defaults False, so nothing restores it) | Reclaim is workspace-aware — skips `$workspace/.cache` in both overlap directions; tests for `/root`, `/home/agent`, `/app` | `e1ade53` |
| 3 | Gemini `toolUsePromptTokenCount` dropped from input | Added to usage detection + folded into input (it's additive), restoring `total == input + output` | `3f88d8a` |
| 5 | MAX forced on every 4.8 request; unanchored regex matched `claude-opus-4-80` | Convert the thinking *shape* (enabled→adaptive) only when thinking is requested; resolve effort from the request / `BENCHFLOW_BEDROCK_THINKING_EFFORT` override (default `high`, never forced MAX); anchored regex shared between proxy + shim (parity-pinned); override forwarded into the Daytona sandbox | `4c0112f` |
| 4 | Global litellm pin framed as scoped; shim deploy fail-open | Own the pin as the deliberate OpenHands-wide bump it is; add a behavioral self-test that runs the venv python against the patched classifier and emits a distinct `ACTIVE` / `NOT active` marker (kept non-fatal so other providers still install) | `9093d93` |

On the E2E follow-up: the Docker + Gemini `ACP -32603` was the `cachedContents` 404 going through the host `TrajectoryProxy` unrewritten — **`7751455` closes exactly that path**. Those canaries ran on `938fc1d`, before the fix, so it's worth a re-run on the new head. (Bedrock Docker already PASSed `reward=1.0`.)

Two honest notes for re-validation:
- **MAX mode is now opt-in** (blocker 5): set `BENCHFLOW_BEDROCK_THINKING_EFFORT=max` — it's auto-forwarded into the Daytona sandbox, and the Docker host proxy reads it directly. Without it, 4.8 runs at the requested / `high` effort instead of silently MAX.
- Blocker 4 is fail-**loud**, not fail-**closed**: the install can't know the model and must still succeed for non-Bedrock providers, so the self-test surfaces a shim-deploy failure rather than aborting.
- The Daytona+Gemini `.env`-poisoning observation stands as a runner-invocation requirement (strip the generic provider vars), not a code change here.

Still owed before merge: healthy Docker + Daytona Gemini canaries on the new head.

bingran-you:

I reviewed this PR with the `thermo-nuclear-code-quality-review` bar and the BenchFlow experiment-health checklist.

This PR is trying to unblock OpenHands SkillsBench runs for Bedrock Claude Opus 4.8 and Gemini, preserve usage telemetry through proxy paths, reclaim disk before verifier install on tight sandboxes, and normalize cross-provider token/cost accounting. The proposed implementation is a set of narrowly gated shims/rewrites: a LiteLLM/OpenHands Bedrock patch, host-side Bedrock thinking normalization, Gemini path normalization in both proxies, pre-verifier cache cleanup, and provider-shape-driven usage normalization.

I do not think this is ready to merge yet.

### Blocking findings

1. **CI is currently red.**

   The required `test` workflow is failing at `uv run ruff format --check src tests`, before tests even run. CI reports two unformatted files: `tests/test_sandbox_hardening.py` and `tests/test_usage_proxy_gemini_rewrite.py`. That contradicts the PR body's current static-validation claim and needs to be fixed before merge.

2. **Gemini path normalization can double-prefix `/v1beta` when the target already includes a version path.**

   `TrajectoryProxy._upstream_url()` appends the normalized request path directly to `self._target` at `src/benchflow/trajectories/proxy.py:92-95`. The normalizer prepends `/v1beta` for unversioned Gemini paths at `src/benchflow/trajectories/gemini_paths.py:52-60`. So a target like `https://generativelanguage.googleapis.com/v1beta` plus `/models/g:cachedContents` becomes `https://generativelanguage.googleapis.com/v1beta/v1beta/cachedContents`.

   The sandbox JS mirror has the same shape: it normalizes to `/v1beta/...` and then prepends `target.pathname` at `src/benchflow/providers/assets/sandbox_usage_proxy.js:116-122`. The tests assert that `https://generativelanguage.googleapis.com/v1beta` is a Gemini target, but they do not exercise `_upstream_url()` or JS `upstreamPath()` with that versioned target.

   **Handoff:** make the rewrite base-path aware, or canonicalize Gemini targets to the origin before path normalization. Add Python and JS parity tests for versioned targets.

3. **The Bedrock thinking-effort boundary still mixes provider vocabularies.**

   `src/benchflow/providers/bedrock_runtime.py:320-336` treats `xhigh` as a valid Bedrock `output_config.effort`, and `src/benchflow/providers/bedrock_runtime.py:361-365` writes the resolved value straight into the Bedrock payload. `tests/test_bedrock_thinking.py:71-72` pins that behavior by asserting requested `xhigh` is honored.

   This leaks an agent/OpenAI-facing reasoning level into a Bedrock wire field. The repo guidance distinguishes `xhigh` for GPT-style calls from `MAX` for Opus-style calls, and this PR is specifically about Opus 4.8 MAX.

   **Handoff:** split request vocabulary from Bedrock `output_config.effort` vocabulary. Either map `xhigh` to a valid Bedrock effort or reject it before constructing the payload. Add a test proving an OpenAI-style `reasoning.effort="xhigh"` never emits `{"effort": "xhigh"}` to Bedrock.

4. **The current PR head needs fresh experiment-health evidence.**

   The PR body still claims clean Daytona reward-1 runs, full skill loading, no leakage, full suite success, and a four-commit/file-count summary, but the branch now has more commits and review-fix changes that alter exact Bedrock effort behavior and Docker/Gemini proxy path behavior. Under `/benchflow-experiment-review`, the old canaries are stale once the proxy/shim/accounting behavior changes.

   **Handoff:** rerun current-head Docker and Daytona canaries for Gemini and Bedrock with usage tracking required, with-skill loading audited, no-skill leakage scan clean, and complete result/trajectory/token/timing metadata.

### Maintainability notes

`src/benchflow/agents/registry.py` is now close to the 1k-line smell threshold, and the OpenHands install command is doing dependency pinning, shim deployment, `.pth` installation, and a behavioral self-test inside one shell string. The next related change should extract this into a focused setup asset/helper rather than growing the registry further.

`src/benchflow/sandbox/lockdown.py` is also very close to 1k lines. I would not block solely on that here, but the verifier-cache cleanup coverage would be easier to maintain in a focused test module rather than further expanding an already-large hardening test file.

### Verification note

Global local checks on the current repo worktree passed: `uv run python -m pytest tests/` (`2552 passed, 49 skipped, 1 deselected`), `uv run ty check src/`, and `uv run ruff check .`. I am running the live Docker/Daytona SkillsBench matrix separately and will report those results separately.

bingran-you:
## Thermo-Nuclear Code Quality Review + Benchflow Experiment Review

Verdict: **not merge-ready yet**. The Bedrock Opus 4.8 path has encouraging Docker/Daytona evidence, but I found several benchmark-fidelity blockers and one schema-compatibility issue that should be fixed before this is used for production SkillsBench data.

### Blockers

1. **Docker ignores run-level `BENCHFLOW_BEDROCK_THINKING_EFFORT`, breaking Docker/Daytona parity.**

   `bedrock_runtime._resolve_bedrock_thinking_effort()` reads only host `os.environ` (`src/benchflow/providers/bedrock_runtime.py:331`), but the Docker Bedrock proxy is constructed with per-run `runtime_env` (`src/benchflow/providers/runtime.py:340`) and calls the translators without passing that env (`src/benchflow/providers/bedrock_proxy.py:299`, `src/benchflow/providers/bedrock_proxy.py:334`). A normal run config / CLI `--agent-env BENCHFLOW_BEDROCK_THINKING_EFFORT=max` therefore works for Daytona direct Bedrock but is ignored by Docker unless the BenchFlow host process also exported the variable. That can silently run Docker at `high` while Daytona runs `max`.

   Repro evidence:
   ```bash
   env -u BENCHFLOW_BEDROCK_THINKING_EFFORT uv run python - <<'PY'
   from benchflow.providers.bedrock_runtime import anthropic_request_to_bedrock_converse
   body = {
       "model": "us.anthropic.claude-opus-4-8",
       "messages": [{"role": "user", "content": "hi"}],
       "thinking": {"type": "enabled", "budget_tokens": 1024},
       "max_tokens": 16,
   }
   print(anthropic_request_to_bedrock_converse(body)["additionalModelRequestFields"])
   PY
   # {'thinking': {'type': 'adaptive'}, 'output_config': {'effort': 'high'}}
   ```

   Suggested fix: make the effort resolver explicit, e.g. pass the proxy runtime env into the Bedrock translators or store the resolved effort on the `BedrockProxyServer` instance. Avoid hidden global env reads for per-run semantics.

2. **The Gemini path rewrite double-prefixes versioned Gemini base URLs.**

   Both `TrajectoryProxy._upstream_url()` (`src/benchflow/trajectories/proxy.py:93`) and the sandbox JS proxy (`src/benchflow/providers/assets/sandbox_usage_proxy.js:117`) normalize the request path before composing it with `target.pathname`. If a user has `GOOGLE_GEMINI_BASE_URL=https://generativelanguage.googleapis.com/v1beta`, then `/models/gemini-...:generateContent` becomes `/v1beta/v1beta/models/...`, which will 404 normal calls and cache probes.

   Repro evidence:
   ```bash
   uv run python - <<'PY'
   from benchflow.trajectories.proxy import TrajectoryProxy
   for target in ["https://generativelanguage.googleapis.com/v1beta", "https://generativelanguage.googleapis.com/v1"]:
       p = TrajectoryProxy(target=target)
       print(target, "=>", p._upstream_url("/models/gemini-3.5-flash:generateContent"))
   PY
   # https://generativelanguage.googleapis.com/v1beta => https://generativelanguage.googleapis.com/v1beta/v1beta/models/gemini-3.5-flash:generateContent
   # https://generativelanguage.googleapis.com/v1 => https://generativelanguage.googleapis.com/v1/v1beta/models/gemini-3.5-flash:generateContent
   ```

   Suggested fix: normalize relative to the configured base path, or strip an existing `/v1` / `/v1beta` base before adding a default version. Add Python/JS parity cases for versioned targets.

3. **Pre-verifier cache reclaim can delete agent-visible non-cache state through symlinks.**

   `harden_before_verify()` runs `rm -rf "$u/.cache/uv" ...` as root (`src/benchflow/sandbox/lockdown.py:892`) after the agent phase. The workspace-overlap guard checks the textual `$u` path, but it does not protect against an agent-controlled symlink such as `/home/agent/.cache -> /app` or `/logs/artifacts`. That can make the verifier delete `/app/uv`, `/app/pip`, or output directories before scoring. Since `restore_workspace` defaults to `False`, this can corrupt benchmark data and turn a real attempt into an unhealthy trial.

   The same block also deletes `/tmp/uv-*`, `/tmp/.uv-*`, and apt archives unconditionally (`src/benchflow/sandbox/lockdown.py:894`), without applying the workspace/path-overlap guard. A task workspace or verifier-visible output under those paths can be removed before scoring.

   Suggested fix: fail closed to known cache directories only after resolving symlinks with `realpath -e` / `find -P`, and run the same overlap guard for every deletion target, including `/tmp` and apt paths. Tests should cover symlinked `.cache` and workspaces under `/tmp/uv-*`.

4. **Daytona Bedrock 4.8 shim deployment is non-fatal even though it is required for that path.**

   The OpenHands install command logs `NOT active` but always exits successfully because the shim deploy/self-test is wrapped with `|| echo ...` and trailing `true` (`src/benchflow/agents/registry.py:580`-`599`). For Daytona OpenHands + Bedrock 4.8, that shim is the only mechanism preventing litellm from sending the rejected legacy thinking shape. If the `.pth` write or import fails, the run can later show up as `ACP -32603` / provider error rather than a clear infra/config failure.

   Suggested fix: add a target-gated preflight/fail-closed path before launching Daytona OpenHands Bedrock 4.8 runs. Non-Bedrock providers can remain non-fatal, but the exact target combo this PR unblocks should not proceed with a known-inactive shim.

### Medium Priority

5. **`n_input_tokens` changed semantics without a schema marker.**

   `agent_result.n_input_tokens` is still serialized under the same field name (`src/benchflow/providers/usage_proxy_runtime.py:710`), but this PR changes its meaning to cache-inclusive normalized input. Old cached Anthropic/Bedrock results and new results are indistinguishable in downstream summaries even though they aggregate differently. Please add an explicit usage-normalization/schema marker, or preserve raw vs normalized input fields so mixed historical comparisons do not silently drift.

6. **The OpenHands `litellm==1.88.0rc1` pin is global for OpenHands, not model-gated.**

   The shim is regex-gated, but the dependency pin in `src/benchflow/agents/registry.py:552` affects every OpenHands provider/model. That is broader than the PR description's byte-identical/no-op framing. Either narrow the rollout claim or add non-target OpenHands canaries proving no request/telemetry regression across representative non-Bedrock providers.

### Verification Run

Static/unit checks on PR head `9093d9386d11fcbf372341416eded98787b0cf78`:

- `uv sync --extra dev --locked` passed.
- `uv run python -m pytest tests/` passed: **2613 passed, 49 skipped, 1 deselected**.
- `uv run ty check src/` passed.
- `uv run ruff check .` passed.
- Targeted subagent checks also passed for Bedrock thinking, Gemini path normalization, token normalization, and sandbox hardening tests.

Key probes:

- Gemini key probe passed: model list returned 50 models.
- Bedrock probe passed in `us-west-2`: `us.anthropic.claude-opus-4-8` Converse returned `ok`; bare `anthropic.claude-opus-4-8` correctly required an inference profile.

Real SkillsBench e2e canaries:

| Case | Sandbox | Task / mode | Result |
|---|---|---|---|
| OpenHands + Bedrock Opus 4.8 + `BENCHFLOW_BEDROCK_THINKING_EFFORT=max` | Docker VM (`daytona-orchestrator`) | `latex-formula-extraction`, task skills | **pass**, reward 1.0, 44 tool calls, usage enabled, `total_tokens=1,917,681` |
| OpenHands + Bedrock Opus 4.8 + `BENCHFLOW_BEDROCK_THINKING_EFFORT=max` | Daytona | `latex-formula-extraction`, task skills | **pass**, reward 1.0, 59 tool calls, usage enabled, `total_tokens=3,917,065` |
| OpenHands + Gemini 3.5 Flash, clean direct Google env | Docker VM | `weighted-gdp-calc`, no task skills | complete healthy fail, reward 0.0, 30 tool calls, usage enabled, 33 cache-probe requests all 200 |
| OpenHands + Gemini 3.5 Flash, clean direct Google env | Daytona | `weighted-gdp-calc`, no task skills | complete healthy fail, reward 0.0, 31 tool calls, usage enabled, 34 cache-probe requests all 200 |
| OpenHands + Gemini 3.5 Flash with inherited external `LLM_BASE_URL`/`OPENAI_BASE_URL` | Daytona | `latex-formula-extraction`, task skills | unhealthy infra/config failure: `ACP -32603`, 0 tool calls, usage unavailable; upstream rejected `/models/gemini-3.5-flash:cachedContents` |

I did **not** mark the full SkillsBench matrix publishable. The targeted canaries show the Bedrock path can succeed on Docker and Daytona, and the clean direct-Google Gemini telemetry path can produce complete trajectories, but the blockers above are enough to invalidate release-readiness until fixed.

bingran-you:
## Supplemental E2E report + blocker root-cause table

AI-assisted supplemental review on PR head `9093d9386d11fcbf372341416eded98787b0cf78`. I separated **countable clean canaries** from **diagnostic/quarantined runs**: the dirty Gemini failures below are useful root-cause evidence, but not healthy benchmark data.

### Real task-run report

| Model | Sandbox | Outcome | Usage tracking | Skill audit / artifact health |
| --- | --- | --- | --- | --- |
| `aws-bedrock/us.anthropic.claude-opus-4-8` | Daytona | **PASS**: `reward=1`, `error=null`, `verifier_error=null`, `partial_trajectory=false`, 59 tool calls | `usage_source=provider_response`, `usage_tracking.environment=daytona`, `endpoint_kind=sandbox`, `telemetry_coverage=1`, input `3,869,658`, output `47,407`, cache read `3,380,754`, cache creation `488,782` | `latex-formula-extraction`, task skills. `marker` + `pdf` present in source hashes and recovered from OpenHands startup catalog (`task_skills_loading=1`). `check_results.py` OK. |
| `aws-bedrock/us.anthropic.claude-opus-4-8` | Docker on `daytona-orchestrator` VM | **PASS**: `reward=1.0`, `error=null`, `verifier_error=null`, `partial_trajectory=false`, 44 tool calls | `usage_source=provider_response`, `usage_tracking.environment=docker`, `endpoint_kind=host`, `telemetry_coverage=1`, input `1,906,188`, output `11,493`, cache read `0`, cache creation `0` | `latex-formula-extraction`, task skills. `marker` + `pdf` recovered from startup catalog (`task_skills_loading=1`). `check_results.py` OK. |
| `gemini-3.5-flash` | Daytona, clean direct-Google env | Healthy model fail: `reward=0`, `error=null`, `verifier_error=null`, `partial_trajectory=false`, 31 tool calls | `usage_source=provider_response`, `usage_tracking.environment=daytona`, `endpoint_kind=sandbox`, `telemetry_coverage=1`, input `1,245,763`, output `29,277`, cache read `368,683`, cache creation `0`; 66 LLM exchanges all `200`, including 34 `cachedContents` cache probes | `weighted-gdp-calc`, no task skills. Expected `xlsx` task skill was **not** loaded (`task_skills_loading=0`), as intended for no-skill mode; leakage scan clean. `check_results.py` OK. |
| `gemini-3.5-flash` | Docker on `daytona-orchestrator` VM, clean direct-Google env | Healthy model fail: `reward=0.0`, `error=null`, `verifier_error=null`, `partial_trajectory=false`, 30 tool calls | `usage_source=provider_response`, `usage_tracking.environment=docker`, `endpoint_kind=host`, `telemetry_coverage=1`, input `1,299,662`, output `31,619`, cache read `357,295`, cache creation `0`; 64 LLM exchanges all `200`, including 33 `cachedContents` cache probes | `weighted-gdp-calc`, no task skills. Expected `xlsx` task skill was **not** loaded (`task_skills_loading=0`), as intended for no-skill mode; leakage scan clean. `check_results.py` OK. |
| `gemini-3.5-flash` | Daytona with inherited shared `.env` provider base URLs | **Unhealthy infra/config failure**: `ACP error -32603`, `partial_trajectory=true`, 0 tool calls, no reward | usage unavailable; local dirty no-skill run captured `401` on `/models/gemini-3.5-flash:cachedContents`; dirty with-skill run captured `403/405` from the external LiteLLM proxy route policy | Root cause pinned: broad env inheritance copied `OPENAI_BASE_URL` / `LLM_BASE_URL` / `BENCHFLOW_PROVIDER_BASE_URL`, and usage proxy target selection chose those agent-level base URLs before the Gemini default. Clean env rerun proves this is config-routing, not model capability. |
| `gemini-3.5-flash` | Docker on `daytona-orchestrator` VM with inherited shared `.env` provider base URLs | **Unhealthy diagnostic run, quarantined**: wrong external-proxy route; process/container was killed after it failed to emit artifacts | no usable usage artifact; only `config.json` exists for the two dirty Docker diagnostics | Same root cause as Daytona dirty env: config shows inherited `OPENAI_BASE_URL` or `BENCHFLOW_PROVIDER_BASE_URL`/`LLM_BASE_URL` for a Gemini model. Not counted as benchmark data. |

### Root-cause table for the 4 High blockers

| High blocker | Root root cause | Impact | How to fix |
| --- | --- | --- | --- |
| Docker ignores run-level `BENCHFLOW_BEDROCK_THINKING_EFFORT` | The Docker Bedrock host proxy is constructed with per-run `runtime_env`, but the Bedrock translators resolve effort from process-global `os.environ` (`bedrock_runtime.py:331`) and `BedrockProxyServer` calls them without passing the per-run env (`bedrock_proxy.py:299`, `:334`). | Docker and Daytona can silently run different thinking efforts. A run config / `--agent-env BENCHFLOW_BEDROCK_THINKING_EFFORT=max` works for Daytona direct Bedrock, but Docker can fall back to `high` unless the BenchFlow host process exported the same variable. This invalidates apples-to-apples Docker/Daytona evidence. | Make effort an explicit runtime parameter: resolve it from `runtime_env` when constructing `BedrockProxyServer`, store it on the server, and pass it into the Bedrock request translators. Add a regression test where host env is unset but proxy `runtime_env` contains `BENCHFLOW_BEDROCK_THINKING_EFFORT=max`. |
| Gemini path rewrite double-prefixes versioned Google base URLs | Both proxy implementations normalize the request path to `/v1beta/...` and then concatenate it with the configured target path. A target like `https://generativelanguage.googleapis.com/v1beta` becomes `/v1beta/v1beta/...`. | Users who set `GOOGLE_GEMINI_BASE_URL` / target to the documented versioned base URL will get 404s for normal Gemini calls and cache probes. Current tests prove versioned targets are recognized as Gemini, but do not test final upstream URL composition. | Canonicalize Gemini targets to origin before rewrite, or make normalization base-path aware. Add Python `TrajectoryProxy` and JS `upstreamPath()` parity tests for `/v1`, `/v1beta`, unversioned targets, and already-versioned incoming paths. |
| Pre-verifier cache reclaim can delete agent-visible state through symlinks / unguarded tmp paths | The workspace-overlap guard is textual and applies to `/root`/`/home/*` user dirs, but `rm -rf "$u/.cache/uv" ...` runs as root after the agent phase and follows agent-created symlinks. `/tmp/uv-*`, `/tmp/.uv-*`, and apt archive deletes are unconditional. | Verifier hardening can mutate completed agent state before scoring, especially with workspaces or outputs under `/tmp` or symlinked `.cache` directories. Since `restore_workspace` defaults false, this can turn real attempts into unhealthy or falsely failed trials. | Resolve deletion candidates with `realpath -e` / `find -P`, reject symlinks, and apply the same workspace/output overlap guard to every path, including `/tmp` and apt paths. Add tests for `.cache -> /app`, `.cache -> /logs/artifacts`, and workspace under `/tmp/uv-*`. |
| Daytona Bedrock 4.8 shim deploy is fail-open although required for that path | The OpenHands install command prints `NOT active` on shim deploy/self-test failure but forces success with trailing `true` (`registry.py:580-599`). For Daytona OpenHands + Bedrock 4.8, that shim is the only protection against LiteLLM's legacy thinking shape. | A shim failure is discovered later as `ACP -32603` / provider error instead of a clear setup failure, wasting long real-task runs and producing unhealthy partial artifacts. It also weakens the PR claim that Daytona Bedrock 4.8 is unblocked. | Keep non-Bedrock providers non-fatal, but add a target-gated preflight before launching Daytona OpenHands Bedrock 4.8: if the shim self-test is not active for `us.anthropic.claude-opus-4-8`, fail the setup clearly before the task starts. |

### Extra runtime root cause pinned during runs

The inherited shared `.env` issue is a separate runner/config-routing bug, not a healthy model failure. `auto_inherit_env()` copies broad base-url variables such as `OPENAI_BASE_URL`, `LLM_BASE_URL`, and `BENCHFLOW_PROVIDER_BASE_URL`; then usage proxy target resolution checks `BENCHFLOW_PROVIDER_BASE_URL` and `_agent_base_url_envs("openhands")` before inferring the Gemini default. For a bare `gemini-3.5-flash` model this can route `cachedContents` to an external LiteLLM proxy, where it is rejected as `401/403/405`. Clean reruns that unset those generic provider base URLs route to `https://generativelanguage.googleapis.com` and produce complete usage artifacts in both Docker and Daytona.

Suggested fix for that separate issue: choose the usage-proxy upstream by resolved provider/model first, not by broad agent-level base-url variables. For Gemini models, prefer `GOOGLE_GEMINI_BASE_URL` / `GEMINI_API_BASE_URL` or the Google default, and ignore unrelated `OPENAI_BASE_URL` / generic LiteLLM proxy URLs unless the model/provider explicitly requests a LiteLLM route.

bingran-you:
## Live SkillsBench validation update

I ran the requested live SkillsBench validation on the GCP `daytona-orchestrator` VM after correcting the environment so Gemini routes directly to Google rather than the AllHands proxy. The initial run was discarded because sourcing the full `.env` injected `OPENAI_BASE_URL` / `LLM_BASE_URL` and routed Gemini through `https://llm-proxy.eval.all-hands.dev`, which hit a proxy budget cap. The clean rerun used a minimal environment: `GEMINI_API_KEY`, `GOOGLE_API_KEY`, `DAYTONA_API_KEY`, `PATH`, `HOME`, and `USER` only.

### Commands / scope

Branch/worktree on VM: `/home/bingran_you/benchflow-pr598` at PR head `9093d9386d11fcbf372341416eded98787b0cf78`.

Task source: `benchflow-ai/skillsbench@main`, resolved to `d1fd798d4d271586267c14abd408c6ec371ba9a4`.

Matrix:

- Docker, no task skills: `jobs/pr-review-e2e-20260602-clean/full-docker-noskill/2026-06-02__23-07-50`
- Docker, with task skills (`--skills-dir auto`): `jobs/pr-review-e2e-20260602-clean/full-docker-skills/2026-06-02__23-07-50`
- Daytona, no task skills: `jobs/pr-review-e2e-20260602-clean/full-daytona-noskill/2026-06-02__23-07-50`
- Daytona, with task skills (`--skills-dir auto`): `jobs/pr-review-e2e-20260602-clean/full-daytona-skills/2026-06-02__23-07-50`

Each run used the 9-task release subset: `jax-computing-basics`, `python-scala-translation`, `jpg-ocr-stat`, `grid-dispatch-operator`, `threejs-to-obj`, `data-to-d3`, `lake-warming-attribution`, `weighted-gdp-calc`, `shock-analysis-supply`; agent/model `gemini` / `gemini-3.1-flash-lite-preview`; `--usage-tracking required`; concurrency 3.

### Results

- **Docker no-skill:** complete and healthy as failed model attempts. Summary: `0/9`, `errors=0`, `verifier_errored=0`, `telemetry_coverage=1.0`, `timing_coverage=1.0`, `total_skill_invocations=0`. Official `tests/integration/check_results.py` status: OK.
- **Docker with-skills:** complete and healthy as failed model attempts. Summary: `0/9`, `errors=0`, `verifier_errored=0`, `telemetry_coverage=1.0`, `timing_coverage=1.0`. Official checker status: OK.
- **Daytona no-skill:** summary exists and final task set is complete: `1/9`, `errors=0`, `verifier_errored=0`, `telemetry_coverage=1.0`, `timing_coverage=1.0`, `total_skill_invocations=0`. However, official checker invalidates the lane because one stale retry artifact for `weighted-gdp-calc` hit idle timeout. Under `/benchflow-experiment-review`, this lane is **not clean publishable without quarantining the stale retry artifact**.
- **Daytona with-skills:** not clean/publishable. The run produced 11 result artifacts for 9 unique tasks but no `summary.json`; `weighted-gdp-calc` had an input-token-limit error on one attempt, and `shock-analysis-supply` produced two `partial_acp` idle-timeout artifacts. The driver then hung after result files were written and had to be terminated. This is an experiment-fidelity failure, not a healthy model failure.

### Artifact-health notes

For latest unique task results in the completed lanes, ACP trajectories were parseable, timing was present, usage came from `provider_response`, and no-skill lanes had `n_skill_invocations=0`. The invalid Daytona artifacts above should be rerun/quarantined before claiming Daytona parity or publishable with-skills coverage.

This does not clear the review blockers I left above; it adds current-head live evidence and confirms the Docker paths are usable while Daytona with-skills still has unhealthy artifacts.

PR #605, opened 2026-06-03
Fix daytona sandbox cleanup/list for SDK >=0.18 iterator API

bingran-you:
## Problem

`bench cleanup`, `bench environment cleanup`, and `bench environment list` all crash against daytona SDK >=0.18 (current resolved version is 0.183):

```
AttributeError: 'generator' object has no attribute 'items'
  at src/benchflow/cli/main.py  ->  if not result.items:
```

This means there is currently **no working way to reap leaked Daytona sandboxes from the CLI** — they keep running and cost money.

## Root cause

The daytona SDK replaced its sandbox-listing API:

| | old (<0.18) | new (>=0.18) |
|---|---|---|
| call | `d.list(page=N, limit=100)` | `d.list(query=None)` |
| return | page object with `.items` (list) | **auto-paginating `Iterator[Sandbox]`** |

Both call sites in `cli/main.py` still drove the old paged shape (`result.items`, `page += 1`, `len(result.items) < 100`), so the first `if not result.items` blows up on the generator.

Why CI didn't catch it: the test fake in `test_cli_daytona.py` also returned the **old** `SimpleNamespace(items=...)` shape, so the suite stayed green while the real CLI was broken — a fixture mirroring an obsolete API.

## Fix

- **`cli/main.py`** — iterate `d.list()` directly at both call sites (`_cleanup_daytona_sandboxes` + `environment list`); drop the manual paging. The SDK paginates internally via cursor. `Sandbox.id/.state/.created_at` and `d.delete(sb)` are unchanged.
- **`test_cli_daytona.py`** — fake `list()` now returns `Iterator[Sandbox]` (the real >=0.18 shape), so the suite would have caught this (and guards against a relapse).
- **`pyproject.toml` / `uv.lock`** — bump `sandbox-daytona` floor `0.153 -> 0.183`. The code now requires the iterator API, and this also moves off the broken `0.176` (its top-level `Daytona` export was missing → `bench` crashed on `from daytona import Daytona`).

## Test

```
uv run pytest tests/test_cli_daytona.py -q
3 passed
```
(daytona 0.183.0, click 8.3.1, typer 0.24.1). Verified live: `bench cleanup --dry-run` was 100% reproducing the AttributeError before this change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,10 @@ dev = [
     "ty>=0.0.1a1",
 ]
 sandbox-daytona = [
-    "daytona>=0.153.0",
+    # >=0.183: list() returns an auto-paginating Iterator[Sandbox] (the older
+    # paged list(page=, limit=) -> .items API was removed), and 0.176's broken
+    # top-level `Daytona` export is fixed.
+    "daytona>=0.183.0",
     "tenacity>=8.0",
 ]
 sandbox-modal = [
--- a/src/benchflow/cli/main.py
+++ b/src/benchflow/cli/main.py
@@ -170,41 +170,36 @@ def _cleanup_daytona_sandboxes(dry_run: bool, max_age_minutes: int) -> None:
 
     d = _daytona_client_or_exit()
     now = datetime.now(UTC)
-    page = 1
     total_deleted = 0
     total_found = 0
     total_skipped = 0
 
-    while True:
-        result = d.list(page=page, limit=100)
-        if not result.items:
-            break
-        total_found += len(result.items)
-        for sb in result.items:
-            if not sb.created_at:
-                continue
-            created_at = datetime.fromisoformat(sb.created_at.replace("Z", "+00:00"))
-            age_minutes = (now - created_at).total_seconds() / 60
-            if age_minutes < max_age_minutes:
-                total_skipped += 1
-                if dry_run:
-                    console.print(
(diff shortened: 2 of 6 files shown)


bingran-you:
Pushed 2 follow-up commits to address the red `test` check + review nit:

- **`2be8b65` — unblock CI**: `ruff format --check src tests` (whole-tree, required check) was red on `tests/test_sandbox_hardening.py` and `tests/test_usage_proxy_gemini_rewrite.py`. Those are **not** part of this PR's logic diff — they landed format-drifted on `main` (last touched by the PR #598 ENOSPC work, e1ade53/08630a3), so the check is red for any PR. Commit is reformat-only (long-tuple line wrapping), no logic change.
- **`66a8f04` — review nit**: per AGENTS.md ('Regression tests must name the PR/commit they guard'), added `Guards PR #605` docstrings to the three daytona list/cleanup tests.

Local gate now fully green:
```
uv run ruff format --check src tests   -> 337 files already formatted
uv run ruff check .                    -> All checks passed!
uv run pytest tests/test_cli_daytona.py -> 3 passed
```

bingran-you:
Note: opened #606 to isolate the `ruff format` drift fix (the two non-daytona test files) so it can land independently and unblock **all** open PRs, not just this one. The format commit (`2be8b65`) stays here too so #605 remains green on its own — whichever merges first resolves it harmlessly.

PR #611, opened 2026-06-04
dashboard: add live Daytona sandboxes panel

bingran-you:
## What

Adds a **Daytona** tab to the v0.5 dashboard showing **active sandbox count, state breakdown, and per-sandbox created-time / age**. This makes orphaned or wedged (`DESTROYING`) sandboxes obvious during large Daytona runs — exactly the case where a stuck teardown or a killed orchestrator leaks sandboxes that keep billing.

## How it's organized (follows existing conventions)

The dashboard already separates concerns: `generate.py` (data), `serve.py` (thin HTTP), `jobs_visibility.py` / `roadmap.py` (domain logic). This PR keeps that split:

- **`dashboard/daytona_status.py`** (new) — `snapshot()` domain logic, mirroring `jobs_visibility.py`. SDK key is passed **explicitly via `DaytonaConfig(api_key=...)`** (no `os.environ` mutation). Sandbox fields are probed across SDK-version-variant names. Returns `{count, by_state, rows, as_of, error?}` — an `error` field instead of raising, so the panel renders failures inline.
- **`dashboard/serve.py`** — a thin `/daytona.json` endpoint that reads the key from an **`X-Daytona-Key` header** (kept out of URLs/request logs; `log_message` is silenced), falling back to the `DAYTONA_API_KEY` env var. Also forwards `--allow-stale-evidence` through to `generate.py` (previously only `--allow-missing-linear` was forwarded, so the server refused to start without a fresh `junit.xml`).
- **`dashboard/index.html`** — a `Daytona` tab + render function + API-key input (stored in `localStorage`) + 15s auto-refresh, following the existing `render*()` pattern.

## Verification

Run locally against real Daytona; panel renders count + state pills + a sorted table of `id · state · created (UTC) · age · target`, auto-refreshing every 15s. `/daytona.json` returns clean JSON (and a friendly `error` string when the SDK or key is missing). No behavior change to any existing tab.

## Usage

`python dashboard/serve.py --allow-missing-linear --allow-stale-evidence` then open the **Daytona** tab; leave the key field blank to use the server's `DAYTONA_API_KEY`, or paste a key to target another account.


bingran-you:
## Thermo-nuclear review fixes (pushed in 25c7991)

All blockers from the review addressed; verified in the **canonical locked env** (the earlier "verified live" had used a manually-upgraded SDK — the real bug was env-specific).

**P0 — panel was broken in the canonical env.** Lock pinned `daytona==0.183.0`, but the sync `Daytona` export only exists from `0.184`, and the SDK's sync client imports `anyio.AsyncContextManagerMixin` at import time (the repo's compat shim handles this; the panel bypassed it).
- Bumped `sandbox-daytona` → `daytona>=0.184.0` (relocked).
- Added canonical `benchflow.sandbox.daytona.build_sync_client()` (anyio compat + explicit-key sync client; no `os.environ` mutation). `daytona_status.snapshot()` **and** `cli/_daytona_client_or_exit()` now both reuse it — and the **duplicate** anyio shim in `cli/main.py` is deleted.

**P1 — static deploy showed a dead tab.** Added a live-server-only `/daytona/available` probe; the frontend renders the Daytona tab only when it returns 200, so the static Vercel build (404) hides it. Probe is gated to http(s) so non-browser/file:// contexts skip it.

**P1 — credential hygiene.** Key is no longer persisted in `localStorage` — held in session memory only (cleared on reload). Server-side `DAYTONA_API_KEY` is the primary path.

**Readiness items.**
- `ruff check` — clean.
- New `tests/test_daytona_status.py` covers `snapshot()` (rows, state-enum stripping, ordering, missing-time, error-returning) via a mocked `build_sync_client` — the SDK path the existing 49 tests didn't reach.
- `ty` — `cli/main.py` passes; `daytona.py` is in the existing `[tool.ty.src] exclude`.
- pip-audit — unrelated/pre-existing (this branch's only dep change is the daytona floor bump; the alert predates it).

Verified live in the locked env: `/daytona/available` → `{"available": true}`, `/daytona.json` → real sandbox counts, no import error.

PR #616, opened 2026-06-04
Make ACP model selection robust across the @agentclientprotocol family

bingran-you:
## Summary

Generalizes the Claude ACP `session/set_model` → `session/set_config_option` fix from [#614](https://github.com/benchflow-ai/benchflow/pull/614) so the **whole `@agentclientprotocol` family** survives the protocol's `set_model` → config-option migration — with **no per-agent registry churn** and **no behavior change for agents that still implement `set_model` today**.

#614 fixed `claude-agent-acp` specifically (it advertises a `model` config option and the registry declares `acp_model_config_id="model"`). But `codex-acp` is the same `@agentclientprotocol` family, is installed **unpinned**, and still routes through `session/set_model`. Its current release (`codex-acp@0.0.45`) only works because it pins the ACP SDK to `^0.22.1` (which still has `session/set_model`) — and it already bundles `configOptions` handling, i.e. the same migration is in progress. When codex-acp ships a release on ACP SDK ≥0.23, an unpinned install would float to it and break with the identical `-32601 Method not found`.

This PR closes that gap at the root.

## Root cause (confirmed at the protocol level)

`claude-agent-acp@0.40.0` depends on `@agentclientprotocol/sdk@0.24.0`. That SDK registers these `session/*` methods:

```
cancel, close, delete, fork, list, load, new, prompt,
request_permission, resume, set_config_option, set_mode, update
```

`session/set_model` → **0 occurrences (removed)**; `session/set_config_option` → its replacement. So any benchflow `set_model` call against an SDK-0.24 agent necessarily returns `-32601`.

## Changes

- **`acp/runtime.py` — capability fallback (the root fix).** On a `set_model` failure that is JSON-RPC `-32601` ("method not found"), automatically re-route the model through the **session-advertised** model config option instead of failing the rollout. Option discovery prefers the registry-declared `acp_model_config_id`, else the conventional `"model"` id, and **only triggers when `set_model` is provably gone**. Any other failure (timeout, non-`-32601` error, no advertised option) still **fails closed**.
  - Net effect: `claude-agent-acp` is unchanged (still the explicit config-option path from #614). `codex-acp` is **byte-for-byte unchanged today** (its `set_model` still works) and is **auto-defused** the moment it migrates — no registry edit required, ever, for the next family member.
- **`agents/registry.py` — pin `codex-acp@0.0.45`** (same hardening as the `claude-agent-acp@0.40.0` pin) so a floating `latest` can't silently break activation. Both pins now carry a comment documenting that the protocol version and the `model`/`effort` config-option ids are coupled — re-verify when bumping.
- **Tests.** `-32601` → config-option fallback; three fail-closed guards (no advertised option, non-`-32601` error, `reasoning_effort` requested with no effort config id); codex pin assertion; and a **gated** pin↔protocol integration guard (`tests/test_acp_pinned_protocol_guard.py`, `RUN_ACP_DEP_GUARD=1`) that codifies the manual SDK-method check.

## Verification

**Unit + static** (off `origin/main` with #614):
- `uv run --extra dev python -m pytest tests/` → **2672 passed, 50 skipped, 0 failed** (extra skips vs #614 are docker-gated tests; no docker on this host).
- `ruff check` ✓ · `ty check src/` ✓

**Live E2E — real Daytona sandbox, local Claude Code subscription (`CLAUDE_OAUTH_TOKEN`; `ANTHROPIC_API_KEY` filtered out), `claude-opus-4-8`, `reasoning_effort=max`:**

```
ACP agent: @agentclientprotocol/claude-agent-acp        # pinned 0.40.0
Session: 4c0ad199-...                                    # session/new OK — no -32601
ACP model config option 'model' set to: claude-opus-4-8 # via set_config_option
ACP reasoning effort config option 'effort' set to: max # via set_config_option
→ end_turn, 8 total tool calls
Running verifier... Rewards: {'reward': 0.667}
Job complete: 0/1, errors=0, idle_timeouts=0, time=10.8min
```

- `result.json`: `error = null`, `model = claude-opus-4-8` · `config.json`: records `model` + `reasoning_effort=max` · trajectory: 24 events, valid JSON, `user_message` → `agent_message`, 8 tool calls. Healthy `fail` (reward 0.667 is opus's task performance, not an infra failure).

**Scope note:** no separate live `codex-acp` run — this change only *adds* a `try/except` around codex's existing `set_model` call (success path identical), the pin is codex's current `latest`, and codex is covered by units (`test_codex_uses_session_advertised_model_id` + the new fallback tests). Happy to run one if you'd like.


bingran-you:
## Review response — `9bab945`

Thanks for the review. Addressed all three findings.

### Finding 1 — capability-first dispatch (the missing "one cut")
`_configure_acp_session` is now genuinely capability-first: it resolves the model config option from the agent's **advertised `session/new` config options** (the registry `acp_model_config_id` is only an override/hint), and dispatches `set_config_option` when present, `set_model` when absent. The "hit the deprecated RPC, then recover on `-32601`" control flow (and the `-32601`/`ACPError` helpers) is gone.

I de-risked the codex regression by **spawning the real agents** and reading what they advertise at `session/new`:
- `claude-agent-acp@0.40.0` → `['mode', 'model', 'effort']`
- `codex-acp@0.0.45` → **`['fast-mode']` only — no `model`**

So capability-first keeps `codex-acp` on `session/set_model` exactly as today (no regression), and a future codex that adds a `model` option is handled automatically with no registry change. `_session_config_option_ids` is hardened against a non-list `config_options`.

### Finding 2 — live guard
`tests/test_acp_pinned_protocol_guard.py` is upgraded from an SDK method-name grep to a **live `session/new` check**: it installs the pinned `claude-agent-acp@0.40.0`, starts it over ACP stdio, and asserts `{"model", "effort"} ⊆` advertised config option ids — so an id *rename* (same method, different id) now fails the guard. `session/new` advertises the options without auth, so it needs no credentials (still gated behind `RUN_ACP_DEP_GUARD=1`). Verified passing.

### Finding 3 — test file split
Dispatch tests moved out of `tests/test_acp.py` (reverted that bloat) into the focused `tests/test_acp_model_config_dispatch.py`.

### Verification
- `uv run --extra dev python -m pytest tests/` → **2710 passed, 13 skipped, 0 failed**
- `ruff check` ✓ · `ty check src/` ✓
- Live guard (`RUN_ACP_DEP_GUARD=1`) → passes
- Real Daytona E2E on the refactored code (claude-opus-4-8 / `reasoning_effort=max` / subscription) re-running to reconfirm the config-option path end-to-end; will post the artifact check.

On the `source.local_path` provenance note from your E2E: agreed, that's an artifact-copyback path issue (run root under the VM's `/tmp`), unrelated to this runtime change.

PR #622, opened 2026-06-04
dashboard: add Experiments tab for SkillsBench fill monitoring

bingran-you:
## What

Adds a new **Experiments** tab to the dashboard for monitoring a SkillsBench experiment fill, with four progressive, non-overlapping panels:

1. **Queue** — cells waiting to run
2. **Running** — in-flight cells (sandbox id + start time)
3. **Completed** — ran but awaiting / failed review (incl. infra `run_failed`, `review_fail`, `quarantined`)
4. **Reviewed & passed** — results: task · model · mode · reward · health · skills-loaded · tokens · time · HF link

A summary line shows `reviewed-healthy / target` with per-state chips, 15s auto-refresh.

## How

Mirrors the existing **live Daytona panel** pattern exactly:
- `dashboard/experiments_status.py` — `snapshot()` reads a generated `experiments_ledger.json` and returns `{as_of, target, summary:{by_state, by_bucket, total}, rows}`, with an `error` field instead of raising.
- `dashboard/serve.py` — adds `/experiments.json` + `/experiments/available` endpoints (same thin-HTTP + capability-probe pattern as `/daytona.json`).
- `dashboard/index.html` — `Experiments` tab registered in `TABS`/`RENDER`, gated by the `/experiments/available` probe (hidden on static Vercel deploys), timer cleared on tab switch.
- `dashboard/.gitignore` — ignores the generated `experiments_ledger.json`.

The ledger itself is produced by an **out-of-band fill orchestrator** (not in this PR) and is gitignored; this PR only renders it.

## Test

- `tests/test_experiments_status.py` — 5 cases (state counts, four-bucket rollup, target fallback, missing/corrupt ledger → error contract). All pass alongside the existing `test_daytona_status.py`.
- Verified live: server serves both endpoints, the tab renders all four panels with sample + real data, no console errors.

If this looks good it can stay as a long-term dashboard module.


bingran-you:
Added a second commit: **persist the Daytona API key server-side**. The panel previously kept the key in memory only, so it was lost on every reload (→ "Authentication credentials not found" unless the server env had `DAYTONA_API_KEY`).

- `POST /daytona/key` (Save button) writes to a gitignored, mode-600 `dashboard/.daytona_key`
- `/daytona.json` resolves: `X-Daytona-Key` header → persisted file → server env
- `do_GET` 404s any dotfile path, so the key file is never served as static content (no localStorage either — avoids exfil via rendered job/artifact content)

Verified: Save persists across reload + server restart; `GET /.daytona_key` → 404; live panel lists sandboxes with no re-entry.

bingran-you:
Split out the Daytona-key persistence into its own PR (#625) and force-pushed this branch back to just the Experiments-tab commit, so the two changes review independently. This PR is now solely the new Experiments tab.

PR #734, opened 2026-06-13
fix: fully detach manifest service starts so Daytona evals don't wedge (#676)

### LINKED ISSUE #676: ManifestEnvironment service start lacks fd detachment — manifest evals wedge on Daytona (same class as downstream service-hook wedge)
## Gap

`ManifestEnvironment` (Environment-plane, `[[environment.services]]`) starts declared services in the sandbox **without full fd redirection**. This is the same failure class as the downstream service-hook wedge root-caused in https://github.com/benchflow-ai/smolclaws/pull/85: on Daytona, `exec` runs as a *session* command and only reports `exit_code` once nothing holds the session's stdout/stderr stream — a backgrounded service that inherits those fds keeps the command "running". With the `timeout_sec=None` hard cap (#670's BF-6 commit) this is now a bounded ~1h hang instead of forever — but a manifest-driven eval on Daytona still wedges instead of running.

Docker is unaffected (`communicate()` returns when the foreground shell exits), so the bug is invisible in local runs and CI.

## Fix

Start manifest services with full detachment — `nohup <cmd> </dev/null >/tmp/<svc>.log 2>&1 &` (all three fds redirected; no `disown`, which is bash/zsh-only and breaks under dash) — plus a bounded exec timeout on the start/readiness calls, keeping per-service logs retrievable. Mirrors the fix that resolved the identical wedge downstream. A fix is being prepared on the clawsbench-compat fork (follow-up to #670).

ElegantLin:
## Summary

`ManifestEnvironment.provision()` started declared services with `{cmd} > {log} 2>&1 &` — stdout/stderr redirected, but **stdin left attached and no `nohup`**.

On Daytona, `exec` runs as a *session* command that only reports an exit code once nothing holds the session's streams (the failure class root-caused in smolclaws#85). A backgrounded service that inherits the session's stdin/stdout/stderr keeps the call "running" until the hard timeout cap (#670) → a manifest-driven eval **wedges (~1h) instead of starting**.

Docker is unaffected — `communicate()` returns when the foreground shell exits — so the bug is invisible in local runs and CI, and only bites on Daytona.

## Fix

Start services with full detachment:

```sh
nohup {cmd} </dev/null >/tmp/benchflow-env-<svc>.log 2>&1 &
```

- all three fds redirected — **the missing `</dev/null` (stdin) is the operative change**; the old form only covered stdout/stderr
- `nohup` to detach from the session's SIGHUP
- **no `disown`** — it is a bash/zsh builtin and the Daytona DinD shell is `sh`/dash
- per-service logs stay retrievable at `/tmp/benchflow-env-<svc>.log`

Mirrors the fix that resolved the identical wedge downstream.

## Test plan

- [x] `test_provision_fully_detaches_service_fds` — asserts `nohup` + all three fd redirections + backgrounded + **no `disown`**; **verified to fail on the old `> {log} 2>&1 &` form** (fail-then-pass). The pre-existing test only checked the command ended with `&`, which is exactly why the missing stdin redirect slipped through.
- [x] `tests/environment/` + manifest inbound suite: 75 passed
- [x] ruff check + format, ty clean

Closes #676

### DIFF
--- a/src/benchflow/environment/manifest_env.py
+++ b/src/benchflow/environment/manifest_env.py
@@ -109,8 +109,19 @@ class ManifestEnvironment:
                 if probe.return_code != 0:
                     continue  # binary missing or its package absent — skip
                 log = f"/tmp/benchflow-env-{svc.name}.log"
+                # Full fd detachment (#676). On Daytona, ``exec`` is a *session*
+                # command that only reports an exit code once nothing holds the
+                # session's streams. A backgrounded service that inherits the
+                # session's stdin/stdout/stderr keeps the call "running" until
+                # the hard timeout cap (#670), so a manifest eval wedges instead
+                # of starting. Redirect ALL THREE fds — ``</dev/null`` for stdin
+                # was the missing one; ``>log 2>&1`` only covered stdout/stderr —
+                # and ``nohup`` to detach from the session's SIGHUP. No
+                # ``disown``: it is a bash/zsh builtin and the Daytona DinD shell
+                # is ``sh``/dash. Docker is unaffected (``communicate()`` returns
+                # when the foreground shell exits), so this only bites on Daytona.
                 await self._sandbox.exec(
-                    f"{svc.command} > {log} 2>&1 &",
+                    f"nohup {svc.command} </dev/null >{shlex.quote(log)} 2>&1 &",
                     timeout_sec=15,
                 )
                 self._started.append(svc)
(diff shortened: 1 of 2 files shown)


bingran-you:
Closing as superseded by the fix already merged through #703 and then into main via #665. Current main contains commit 654febfb (`fix(environment): detach manifest service starts for Daytona session execs`), which moved the service-start shape into `service_start_command()` and already includes regression coverage for full fd detachment, bounded start timeout, and reset restart shape. This PR is now dirty because it patches the old inline implementation; merging it would duplicate/stale the fix rather than improve main. Issue #676 is being closed separately with the same evidence.

PR #787, opened 2026-06-15
feat(cli)!: flatten `bench eval adopt`; drop Pier; prune CONVERT.md /…

xdotli:
… dev-docs / labs

Rebuilt off current main (the prior branch had diverged from main's in-flight 0.6.2 doc work, which is kept).

CLI:
- `bench eval adopt` is now a single multi-mode command: `bench eval adopt <source>` (scaffold-if-missing + drive the conversion), `--scaffold-only`, `--verify`. The three action bodies are extracted into shared functions; the deprecated `bench agent create|run|verify` and `bench adopt …` aliases still work (hidden) and point their hints at the new shapes.
- Remove the `pier` export target entirely: `bench tasks export --target` is harbor-only; `CompatibilityTarget = Literal["harbor"]`.

Adoption guide:
- Remove the benchmarks/CONVERT.md file; the conversion guide now lives in code (agent_router_guide.CONVERSION_GUIDE), rewritten to the current adoption flow and embedded into the convert prompt.

Cleanup (conflict-free with main):
- Remove .claude/dev-docs/, docs/labs/, and docs/benchmark-adoption.md; sync the Mintlify nav (docs.json + docs/docs.json) and fix the resulting doc references (cli.md, running-any-benchmark, running-benchmarks, sandbox-hardening, concepts).

CI-exact gates green (ruff/format/ty over src tests tools) + full suite 4189 passed.


xdotli:
Addressed the bot review in 795cc93b:

- **@chatgpt-codex-connector** (`--benchmarks-dir` dropped in convert): now honored end-to-end — `benchmarks_dir` is threaded through `run_convert_action` → `run_agent_adoption`/`prepare_adoption_launch` → the codex prompt's target path, so the scaffold location and the conversion target stay consistent (default `benchmarks/<name>/` unchanged). Added a regression test.
- **@greptile-apps**: dropped the stale `init` subcommand reference ("stubbed by `init`" → "by the scaffold") and fixed the tense ("scaffolded" → "scaffolds") in the embedded conversion guide.

Full suite green (4190 passed).

PR #818, opened 2026-06-20
feat(agents): core manifest consumer loader (decoupling contract, gated off)

Yiminnn:
## What

The **consumer half** of the agent-decoupling contract (ADR-0001): core reads an
agent's data-only `manifest.toml` and turns it into the `AgentConfig` the registry
already speaks. A new agent can then ship as **data** (a manifest in
`benchflow-ai/agents`) instead of a hand-edited entry in `registry.AGENTS`
(decision #4), and a directory of manifests merges into the registry on **explicit
opt-in** via `$BENCHFLOW_AGENTS_DIR` (decision #7).

This is the core-side counterpart to the author-side schema + reference loader
already in `benchflow-ai/agents` (`contract/manifest_schema.json`, PR #14).

## Gated OFF by default — zero behaviour change

`registry.py` calls the merge at end-of-module, but it is a **no-op when
`BENCHFLOW_AGENTS_DIR` is unset**, so a default `import benchflow.agents.registry`
is byte-for-byte unchanged (`AGENTS` count unchanged). The dual-source registry
only activates when a developer opts in.

## `src/benchflow/agents/manifest.py`

- **`load_agent_manifest(path) -> LoadedManifest{config, aliases}`** — the
  *consumer* contract (deliberately more lenient than the author-side JSON Schema):
  enforces the 4 required keys, the **contract-major gate** (`SUPPORTED_CONTRACT_MAJOR=1`;
  a v1.x manifest carrying a field added in a later minor still loads — SemVer
  forward-compat), and an **acp-only protocol gate** (the consumer re-enforces it
  because `$BENCHFLOW_AGENTS_DIR` bypasses the author-side jsonschema and
  `register_manifest_agents` writes `AGENTS` directly, skipping `VALID_PROTOCOLS` —
  so this loader is the last protocol gate). Maps the 14 contract data fields onto
  `AgentConfig`; ignores unknown keys.
- **`_FIELD_MAP` / `_SHIM_ONLY` partition `AgentConfig` exactly** — a test asserts
  it, so a field added to `AgentConfig` can never be **silently dropped** by the
  loader (it must be consciously routed to one set or the other). This guard already
  earned its keep: it caught `session_factory` (the non-ACP `module:callable` seam,
  re-added in #782) and pinned it as shim-only.
- **`discover_manifests`** — recursive, skips the root-level manifest and
  vendor/VCS dirs, so nested families (`ai-sdk/<variant>/manifest.toml` alongside
  `mimo-acp/manifest.toml`) are all found.
- **`register_manifest_agents` / `register_env_manifest_agents`** — fail-loud on
  name/alias collision (ambiguous source of truth, not a silent shadow),
  all-or-nothing, and write **every** name-keyed registry map
  (`AGENTS` / `AGENT_INSTALLERS` / `AGENT_LAUNCH` / `AGENT_ALIASES`) — `install.py`
  and `rollout_planes.py` read the installer/launch projections by name, so
  AGENTS-only would leave a manifest agent uninstallable.

## Test plan

- [x] `pytest tests/agents/test_manifest_loader.py test_manifest_dirscan.py test_manifest_wiring.py` — **27 passed**
  (loader 15 / dir-scan 10 / import-time wiring 2).
- [x] Partition guard green against current `AgentConfig` (22 fields = 14 declarable + 8 shim-only).
- [x] Default import unchanged when `BENCHFLOW_AGENTS_DIR` unset (wiring tests spawn a fresh interpreter to prove it).
- [x] `ruff check` + `ruff format --check` clean.
- [ ] CI on this PR.

## Not in this PR

Authoring residents (`ai-sdk/*`, omnigent ACP-stdio server) and the real
standalone-platform parity fixture are follow-ups tracked in ADR-0003. This PR is
just the core seam, dark by default.

### DIFF
--- a/src/benchflow/agents/manifest.py
+++ b/src/benchflow/agents/manifest.py
@@ -1,0 +1,296 @@
+"""Thin agent-manifest loader — the consumer half of the decoupling contract.
+
+An agent declares itself in a ``manifest.toml`` (see the agents repo's
+``contract/manifest_schema.json`` for the authoritative, strict schema). Core
+*consumes* that declaration here, turning it into the ``AgentConfig`` the rest
+of BenchFlow already understands — so a new agent can ship as data (a manifest)
+instead of a hand-edited entry in ``registry.AGENTS`` (design decision #4), and
+a directory of manifests can be merged into the registry at opt-in (#7).
+
+Authoring vs. consuming, on purpose:
+
+* The agents-repo JSON Schema is the *author's* contract — ``additionalProperties:
+  false``, every field type-checked. It fails an author loudly for a typo.
+* This loader is the *consumer's* contract — it enforces only the major-version
+  gate and the four required keys, then maps the declared fields onto
+  AgentConfig and **ignores unknown keys**. A v1.x manifest that carries a field
+  added in a later minor still loads on this v1 loader (SemVer: a minor bump is a
+  backward-compatible addition). The AgentConfig fields outside the contract are
+  the shim/credential set (session_factory, credential_files, subscription_auth,
+  acp_model_config_id, acp_effort_config_id, disallow_web_tools_*); they keep
+  their AgentConfig defaults because they are logic the shim owns, not data
+  (see _SHIM_ONLY and the partition test).
+"""
+
+from __future__ import annotations
+
+import os
+import re
+import tomllib
+from collections.abc import Mapping
+from dataclasses import dataclass
+from pathlib import Path
+
+from benchflow.agents.registry import AgentConfig
+
+# This loader speaks contract major version 1. A manifest declaring a different
+# major is rejected (its shape may be incompatible); a different minor/patch is
+# accepted (additive-only within a major).
+SUPPORTED_CONTRACT_MAJOR = 1
+
+# The contract is acp-only ("one protocol", ADR-0001 §1; author schema enum:["acp"]).
+# The consumer re-enforces it because the BENCHFLOW_AGENTS_DIR dev override bypasses
+# the author-side jsonschema, and register_manifest_agents writes AGENTS directly,
+# skipping registry.VALID_PROTOCOLS — so this loader is the last protocol gate.
+_SUPPORTED_PROTOCOLS = frozenset({"acp"})
+
+# Directory env override for opt-in filesystem discovery (decision #7).
+MANIFEST_DIR_ENV = "BENCHFLOW_AGENTS_DIR"
+
+_REQUIRED = ("contract_version", "name", "install_cmd", "launch_cmd")
+_VERSION_RE = re.compile(r"^\d+\.\d+(\.\d+)?$")
+
+# manifest key -> AgentConfig field: the contract's data fields (the PR #14 schema
+# minus the meta keys contract_version/aliases). The consumer must cover EVERY
+# schema data field with an AgentConfig home; _SHIM_ONLY below holds the rest.
+# A test asserts _FIELD_MAP.values() | _SHIM_ONLY partitions AgentConfig exactly,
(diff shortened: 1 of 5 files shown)


bingran-you:
Automation triage: reproduced the failing CI as a local ty narrowing error in register_env_manifest_agents, and confirmed Greptile's same-batch alias collision finding is a real fail-loud contract gap. I pushed follow-up commit ab631123 on this PR branch that:

- uses explicit None checks so ty narrows the registry maps correctly
- preflights same-batch alias/name and alias/alias collisions before mutating any registry map
- adds regression coverage for existing-alias, same-batch alias/alias, and same-batch alias/name collisions

Local validation after the patch:

- uv sync --extra dev --extra sandbox-daytona --extra sandbox-modal --locked
- uv run python -m pytest tests/agents/test_manifest_loader.py tests/agents/test_manifest_dirscan.py tests/agents/test_manifest_wiring.py -q (30 passed)
- uv run python -m pytest tests/agents -q (46 passed)
- uv run ty check src
- uv run ruff check src tests tools
- uv run ruff format --check src tests tools
- uv run python -m pytest tests/ -q (4485 passed, 4 skipped, 7 deselected)

CI should be the remaining gate before human review/merge.

bingran-you:
Automation triage (2026-06-21): moved `review:pending` → `review:changes-requested` because `test` is currently red on a `ty` type error (4 diagnostics), so the author needs to push before this can merge.

**Root cause** — `register_env_manifest_agents` (`src/benchflow/agents/manifest.py`) types `agents/aliases/installers/launch` as `dict[...] | None`, then only reassigns the `None` ones *inside* the `if None in (agents, aliases, installers, launch):` block. `ty` cannot narrow tuple-membership, so on the branch where the condition is False the four names keep their `... | None` type, and the subsequent `register_manifest_agents(loaded, agents=agents, …)` call fails `invalid-argument-type` (`None` not assignable to `dict[str, str]`) — once per param, hence the 4 diagnostics.

**Fix** — narrow explicitly after the defaulting block, e.g.:
```python
    agents = AGENTS if agents is None else agents
    aliases = AGENT_ALIASES if aliases is None else aliases
    installers = AGENT_INSTALLERS if installers is None else installers
    launch = AGENT_LAUNCH if launch is None else launch
assert agents is not None and aliases is not None
assert installers is not None and launch is not None
```
(or lift the four `X = GLOBAL if X is None else X` lines out of the `if` so they run unconditionally — that also lets `ty` see them as non-`None` at the call site). `uv run ty check` should then pass.

bingran-you:
CI follow-up: GitHub checks are green on the patched head ab631123 (, , , detect-scope; matrix/final-review lanes skipped by scope). Non-author human review is the remaining merge gate.

bingran-you:
Correction to the previous CI note: GitHub checks are green on patched head ab631123. Passing checks are test, pip-audit, rollout-smoke, and detect-scope; matrix/final-review lanes were skipped by scope. Non-author human review remains the merge gate.

Yiminnn:
Closing for now per request — reverting the outward push; will revisit before re-opening for review. Branch retained.

Yiminnn:
Superseded by #825 — the combined all-paths decouple core consolidates this loader plus additive merge-mode, the codex self-writing launcher, and the omnigent session-factory seam, end-to-end-proven on docker+daytona. Recommend closing this in favor of #825.

bingran-you:
Automation triage (2026-06-23): closing this as superseded by #825.

I verified `origin/feat/agent-manifest-loader` is an ancestor of `origin/feat/acp-decouple-all`, and #825's PR body explicitly consolidates this consumer-loader PR plus the additive merge mode, codex self-write launcher, and session-factory seam. Keeping the branch untouched; #825 is now the active review path.

PR #825, opened 2026-06-23
feat(agents): all-paths decouple core — manifest loader (8 ACP) + omnigent session-factory seam (additive, gated)

Yiminnn:
## What

Additive, **compatible** decoupling of per-agent integration data out of benchflow core into the standalone `benchflow-ai/agents` manifest contract — plus the host seam for the one non-ACP agent family (omnigent). **Gated off by default:** `AGENTS` is byte-identical unless `BENCHFLOW_AGENTS_DIR` is set, and no core agent is deleted (strangler-fig).

This consolidates and **supersedes #818** (the consumer loader alone): it adds the additive merge-mode, the codex self-writing launcher, and the omnigent session-factory connect/drive seam, into one deployable, end-to-end-proven all-paths core. #818 can be closed in favor of this.

## Commits
- `6e4699d9` — codex-acp **self-writes** `~/.codex/auth.json` in its launcher (drops the core-side `credential_files` writer; byte-identical template; subscription mode untouched).
- `82f35f55` — core **manifest consumer loader** (`load_agent_manifest`; MAJOR-gated `SUPPORTED_CONTRACT_MAJOR=1`; closed `_FIELD_MAP` + named `_SHIM_ONLY` partition guard; protocol re-gate for the `BENCHFLOW_AGENTS_DIR` bypass). *(= #818's content.)*
- `52a1cca6` — **additive merge-mode** (`merge_shim_only`): a manifest reproducing a core agent overrides its 14 data fields but **keeps** core's host-side shim-only fields (`subscription_auth`, `credential_files`, …) → merged == orig.
- `ae26646c` — **session-factory connect + drive seam** to host non-ACP agents (omnigent's one-shot `omnigent run -p`); capture stays protocol-agnostic via the LiteLLM proxy.
- `685a7990` — `execute()` connected-guard accepts a session-factory session (no ACP client).
- `2ce0e0d8` — plumb kernel-resolved `agent_cwd` into the session-factory connect (omnigent runs in the resolved workspace, not a hardcoded `/app`).
- `03d1ed0c` — **codex ChatGPT-subscription fix**: gate `_configure_codex_custom_provider` on `uses_native_subscription_auth` so subscription runs keep codex on its built-in `openai` provider (chatgpt auth) instead of an injected `benchflow-openai` custom provider that forced API-key mode. Pre-existing core bug (in `release/v0.6.0` too) — not a decouple regression; surfaced here because the decouple exercises the subscription path end-to-end. + regression test.

## Compatibility
[...]

PR #838, opened 2026-06-27
test(parity): assert manifest-loaded agent defs byte-match in-core (gate for the agents-repo flip)

Yiminnn:
## What

Adds `tests/agents/test_manifest_parity.py` -- the byte-identical parity gate for the agent-decoupling source-of-truth flip. For every core agent that is also declared by a `manifest.toml` in [benchflow-ai/agents](https://github.com/benchflow-ai/agents), it asserts the manifest-loaded `AgentConfig` matches the in-core `registry.AGENTS` entry field-for-field on the 14 manifest-owned data fields.

This change is additive and test-only: it does not change the default agent source (core `registry.AGENTS` stays authoritative) and it deletes no core definitions.

## Why -- this gates the future "step 6"

Decision #4 of the decouple moves the data for each agent out of `registry.AGENTS` into a `manifest.toml`. The future flip ("step 6") makes the manifests the default source and then deletes the in-core defs. That flip is only safe if every manifest that reproduces an existing core agent is byte-identical to the core entry -- otherwise deleting the core def would silently change behaviour.

This PR is that gate. **Step 6 (flip default + delete core defs) must not proceed unless this test is green for all manifest-backed core agents.** On drift it fails loudly naming the exact agent + field to reconcile.

## What it checks

- Parametrized over the 10 manifest-backed core agents (each its own test case): claude-agent-acp, codex-acp, deepagents, gemini, harvey-lab-harness, mimo, openclaw, opencode, openhands, pi-acp.
- Compares the 14 data fields owned by the manifest contract (equal to `manifest._FIELD_MAP`): name, description, install_cmd, launch_cmd, protocol, api_protocol, acp_model_format, supports_acp_set_model, requires_env, install_timeout, env_mapping, default_model, skill_paths, home_dirs. The remaining `AgentConfig` fields are `_SHIM_ONLY` host/credential concerns a data manifest never carries (out of scope).
- `test_parity_set_is_exactly_the_core_manifest_agents`: the core-vs-manifest intersection is exactly those 10, keeping `omnigent-pi` (a session-factory package agent wired via #825, not a manifest) the sole core agent without a manifest, and forcing a deliberate update when a manifest-backed core agent is added or removed.
- `test_parity_fields_track_the_loader_contract`: pins the compared field set to `_FIELD_MAP` so a newly-added contract data field cannot silently escape the gate.
- `test_parity_assertion_catches_synthetic_drift`: hermetic load-bearing check that the comparison rejects a perturbed field, so the gate is never a silent no-op.

## CI-robust design

A session-scoped fixture clones benchflow-ai/agents at a pinned commit ([eb6c60da](https://github.com/benchflow-ai/agents/commit/eb6c60da6de3cad331e79424d4048048d5399a2e), agents [PR #24](https://github.com/benchflow-ai/agents/pull/24) "restore chmod 600 on auth.json in launch_cmd (parity with core #825)") into a throwaway dir, then loads it with the production `load_agents_from_dir`. Pinning the SHA makes the gate deterministic, offline-after-clone, hermetic (no `$BENCHFLOW_AGENTS_DIR` setup -- the pin lives in source), and auditable. When the agents repo advances, bump `AGENTS_REPO_PIN` and re-run; drift surfaces immediately. If the clone is impossible (offline or no git) the live checks skip with a clear reason while the hermetic synthetic-drift test still runs.

## Local verification

- `pytest tests/agents/test_manifest_parity.py` -> 13 passed (10/10 manifest core agents byte-match at the pin).
- Load-bearing: temporarily drifting one in-core field (`gemini.description`) makes the `gemini` case fail naming `[description]` with manifest-vs-core values; then restored.
- Regression `tests/agents/ tests/test_agent_registry.py` -> 87 passed. `ruff check` + `ruff format` clean.

### DIFF
--- a/tests/agents/test_manifest_parity.py
+++ b/tests/agents/test_manifest_parity.py
@@ -1,0 +1,240 @@
+"""Byte-identical parity gate: every core agent reproduced by a benchflow-ai/agents
+manifest must match its in-core ``registry.AGENTS`` entry, field for field.
+
+The agent-decoupling plan (decision #4) makes each agent ship as *data* — a
+``manifest.toml`` in ``benchflow-ai/agents`` — instead of a hand-authored entry in
+``registry.AGENTS``. The future "step 6" flips the source of truth (default to the
+manifests, then delete the in-core defs). That flip is only safe if every manifest
+that reproduces an existing core agent is **byte-identical** to the core entry —
+otherwise deleting the core def would silently change behaviour. This file is that
+gate: it must be green for all manifest-backed core agents before step 6 proceeds.
+
+Why a pinned-ref clone (the most CI-robust shape)
+-------------------------------------------------
+A session-scoped fixture clones ``benchflow-ai/agents`` at a single pinned commit
+(``AGENTS_REPO_PIN``) into a throwaway dir, then loads the manifests with the
+production ``load_agents_from_dir``. Pinning the exact SHA makes the gate:
+
+* deterministic — the exact commit yields the exact manifests, reproducible run
+  to run (a moving branch tip could flip the result between identical pushes);
+* offline after the one clone — no network mid-test, no flaky live API;
+* hermetic — no ``$BENCHFLOW_AGENTS_DIR`` / env setup; the pin lives in source;
+* auditable + updatable — the commit is visible here; when the agents repo
+  advances you bump ``AGENTS_REPO_PIN`` and re-run, and any drift surfaces at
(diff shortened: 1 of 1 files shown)


Yiminnn:
Superseded by #837 (already merged to main — same path tests/agents/test_manifest_parity.py, more CI-robust env-var/child-interpreter design). Closing as redundant.

PR #839, opened 2026-06-27
fix(eval): emit llm_trajectory.jsonl for streaming claude-agent-acp rollouts

### LINKED ISSUE #833: claude-agent-acp does not emit trajectory/llm_trajectory.jsonl for successful rollouts (pi does)
## Summary

The `claude-agent-acp` agent runner does **not** emit `trajectory/llm_trajectory.jsonl` for **successful** rollouts. It only writes that file on the error/partial path. The `pi` agent, by contrast, writes both `trajectory/acp_trajectory.jsonl` **and** `trajectory/llm_trajectory.jsonl` for every successful rollout.

Net effect: for a clean `claude` eval run you can recover the agent-side ACP event stream but **not** the per-LLM-call request/response trajectory, so the two trajectory artifacts are never both populated for the same rollout. This breaks parity with `pi` and makes successful claude rollouts unusable for any consumer that expects the LLM-level trajectory (e.g. trajectory export / RL training).

## Environment

- benchflow **v0.6.3** (installed via `uv tool install "git+https://github.com/benchflow-ai/benchflow"`)
- sandbox: `docker`
- agent: `claude` (registered as `claude-agent-acp`)
- model: self-hosted via the `vllm/` provider (OpenAI-compatible endpoint), routed through BenchFlow's litellm anthropic→openai proxy
- benchmark: 50 single-question SimpleQA task dirs, `--concurrency 3`

## Expected

Every rollout produces both, regardless of agent or outcome:
- `trajectory/acp_trajectory.jsonl` — agent-side ACP event stream
- `trajectory/llm_trajectory.jsonl` — per-call `/v1/messages` request/response capture (`{request, response, duration_ms}` records)

## Actual

Artifact presence is mutually exclusive by outcome. Counts over the 70 task dirs (50 scored + 20 errored retries) of one run:

| count | `trajectory_source` | `agent/acp_trajectory.jsonl` | `trajectory/acp_trajectory.jsonl` | `trajectory/llm_trajectory.jsonl` |
|------:|---------------------|------------------------------|-----------------------------------|-----------------------------------|
| 50 | `acp` (success)     | populated | populated | **missing** |
| 11 | `partial_acp` (errored) | missing | stub (382 B, 0 events) | populated |
|  6 | `partial_acp` (errored) | missing | stub | missing |

So on the happy path (`trajectory_source: acp`) the LLM trajectory is never written; it only appears when the rollout errors out (`partial_acp`), and even then inconsistently (11 of 17).

### Cross-agent contrast (same BenchFlow version, same `vllm/` provider)

| agent | successful rollouts with `trajectory/llm_trajectory.jsonl` |
|-------|-----------------------------------------------------------|
| `pi`               | 44 / 44 |
| `claude-agent-acp` | 0 / 50 |

The `pi` runner emits `llm_trajectory.jsonl` for every non-errored rollout; `claude-agent-acp` emits it for none. This points to the LLM-trajectory capture being wired into the pi runner but missing from the claude-acp runner's success path (it appears to only get flushed from the proxy log on the error path).

## Repro

1. `bench eval run --agent claude --model vllm/<model> --tasks-dir <simpleqa tasks> --sandbox docker`, with the engine env injected for the `vllm/` provider (`BENCHFLOW_PROVIDER_BASE_URL`, `OPENAI_API_KEY`, `BENCHFLOW_PROVIDER_MODELS`).
2. Inspect any successful task dir under the jobs dir: `trajectory/acp_trajectory.jsonl` is present and populated, `trajectory/llm_trajectory.jsonl` is absent.
3. Run the same benchmark with `--agent pi`: both files are present on successful tasks.

## Notes

- This particular run did hit self-hosted-engine instability (the inference endpoint dropped mid-run, producing the `partial_acp` retries with `ACP error -32603 … Cannot connect to host …`). That explains *why there are errored rollouts*, but is independent of the artifact bug: the missing `llm_trajectory.jsonl` is on the **successful** (`acp`) rollouts, where there is no error.
- For the errored `partial_acp` rollouts, `trajectory/acp_trajectory.jsonl` exists but is a 382-byte stub containing no real ACP events, while `agent/acp_trajectory.jsonl` is absent entirely — possibly worth aligning too, but the primary ask is: **emit `trajectory/llm_trajectory.jsonl` for successful `claude-agent-acp` rollouts, matching the `pi` runner.**

Yiminnn:
## Summary

`claude-agent-acp` did not write `trajectory/llm_trajectory.jsonl` for **successful** rollouts (0/50), while `pi` did (44/44) and claude-acp's *error* path did (11/17). This breaks trajectory parity and makes successful claude rollouts unusable for any LLM-trajectory consumer (export / RL training).

## Root cause

The per-rollout LiteLLM proxy captures the LLM trajectory from its `litellm_settings.callbacks` success/failure events. BenchFlow's `vllm/` provider maps the upstream to `openai/<model>`. For the anthropic `/v1/messages` route against an `openai/`-prefixed upstream, **LiteLLM 1.89 bridges through the OpenAI Responses API (`/responses`), and its Responses *streaming* adapter never invokes the success callback** — so streaming `/v1/messages` success produces zero callback records → empty `server.trajectory.exchanges` → `_write_llm_trajectory` early-returns and writes nothing.

- `claude-agent-acp` (Claude Code) **streams** `/v1/messages` → hits the gap (0/50 on success).
- `pi` calls `/v1/chat/completions` directly → streaming success logs fine (44/44).
- claude-acp **errors** go through `post_call_failure_hook` (failures *do* log) → that's the 11/17 on the error path.

## Fix

Set `litellm.use_chat_completions_url_for_anthropic_messages` via the proxy's `litellm_settings`, so the anthropic `/v1/messages` bridge uses `/chat/completions` — the universally-supported endpoint whose streaming path logs success correctly. Native `anthropic/` providers don't use the Responses bridge (unaffected); the proxy is BenchFlow-owned and per-rollout, so there is no external blast radius.

## Verification

Reproduced with a live-proxy harness (the real proxy started via the runtime + a mock OpenAI-compatible upstream), driving the exact `openai/` route the `vllm/` provider produces:

| case (openai/ route = #833 config) | upstream hit | success rec | exchanges | file |
|---|---|---|---|---|
| messages-stream (claude-acp) — **before** | `/responses` | 0 | 0 | ✗ |
| messages-stream (claude-acp) — **after** | `/chat/completions` | 1 | 1 | ✓ |
| chat-stream (pi) — after | `/chat/completions` | 1 | 1 | ✓ (unchanged) |

A regression test (`test_proxy_config_forces_chat_completions_for_anthropic_messages`) asserts the flag is present in the generated proxy config. `tests/test_litellm_config.py` + `tests/test_litellm_hardening.py` pass (57 passed).

Closes #833

### DIFF
--- a/src/benchflow/providers/litellm_config.py
+++ b/src/benchflow/providers/litellm_config.py
@@ -397,5 +397,12 @@ def litellm_proxy_config(
             "callbacks": [f"{callback_module}.proxy_handler_instance"],
             "drop_params": True,
             "set_verbose": False,
+            # Force the anthropic /v1/messages bridge onto /chat/completions.
+            # LiteLLM routes openai/-prefixed upstreams (e.g. the vllm
+            # provider) through its Responses-API adapter, whose *streaming*
+            # path never fires the success callback -- so streaming
+            # claude-agent-acp rollouts produced no llm_trajectory.jsonl
+            # (#833). /chat/completions logs success correctly.
+            "use_chat_completions_url_for_anthropic_messages": True,
         },
     }
--- a/tests/test_litellm_config.py
+++ b/tests/test_litellm_config.py
@@ -125,6 +125,22 @@ def test_proxy_config_registers_plain_and_openai_aliases():
     }
 
 
+def test_proxy_config_forces_chat_completions_for_anthropic_messages():
+    """Streaming /v1/messages must bridge via /chat/completions so the
+    LiteLLM success callback fires and llm_trajectory.jsonl is written for
+    claude-agent-acp. LiteLLM's Responses-API streaming adapter (used for
+    openai/-prefixed upstreams such as the vllm provider) skips the success
+    callback; this flag opts out of it (#833).
+    """
+    route = resolve_litellm_route("openai/gpt-5.4-mini", {"OPENAI_API_KEY": "key"})
+    config = litellm_proxy_config(route, master_key="sk-local")
+
+    assert (
+        config["litellm_settings"]["use_chat_completions_url_for_anthropic_messages"]
+        is True
+    )
+
+
 def test_proxy_config_registers_requested_bare_model_name():
     """Codex ACP sends the bare selected model name to the proxy."""
     route = resolve_litellm_route("openai/gpt-5.4-mini", {"OPENAI_API_KEY": "key"})
--- a/tests/test_litellm_config.py
+++ b/tests/test_litellm_config.py
@@ -141,6 +141,20 @@ def test_proxy_config_forces_chat_completions_for_anthropic_messages():
     )
 
 
+def test_litellm_exposes_anthropic_messages_chat_completions_flag():
+    """Guard against a LiteLLM upgrade silently dropping the flag.
+
+    litellm_proxy_config sets use_chat_completions_url_for_anthropic_messages
+    via LiteLLM's generic litellm_settings -> setattr(litellm, ...) path, which
+    does NOT raise on an unknown key. If a future LiteLLM renames or removes
+    this attribute the fix would become a silent no-op and regress #833, so
+    assert the attribute still exists.
+    """
+    import litellm
+
+    assert hasattr(litellm, "use_chat_completions_url_for_anthropic_messages")
+
+
 def test_proxy_config_registers_requested_bare_model_name():
     """Codex ACP sends the bare selected model name to the proxy."""
     route = resolve_litellm_route("openai/gpt-5.4-mini", {"OPENAI_API_KEY": "key"})


bingran-you:
Users Simulation automation review (2026-06-27T12:21Z): **ready for this simulation scope**.

I verified the PR through cost-light CLI/SDK and focused code paths rather than a live expensive model rollout. The LiteLLM proxy flag is wired in the canonical config builder, focused runtime/hardening checks passed, and the trajectory validator still fails closed when `trajectory/llm_trajectory.jsonl` is missing.

Commands/evidence:
- `uv run pytest tests/test_litellm_config.py -q`
- `uv run pytest tests/test_litellm_runtime.py -q`
- `uv run pytest tests/test_litellm_hardening.py -q`
- `uv run ruff check src/benchflow/providers/litellm_config.py tests/test_litellm_config.py`
- `uv run ty check src/benchflow/providers/litellm_config.py`
- CLI/SDK dry paths: `benchflow --help`, `benchflow eval run --help`, `benchflow eval list --help`, `benchflow skills --help`, `benchflow hub check --help`, `benchflow sandbox --help`, Python SDK import smoke.
- `validate_run_artifacts.py` passed on a healthy synthetic rollout and failed closed after removing `llm_trajectory.jsonl`.

Residual risk: the regression test is static config coverage, not a real streamed `claude-agent-acp` provider canary, so this does not prove live Claude streaming end to end.

PR #879, opened 2026-07-02
fix(acp): derive model-via-env from registration data, not agent names

Yiminnn:
Found by the 3-path clean-remote census (all agents × deepseek-v4-flash × citation-check): every registry-wired agent that **advertises an ACP `model` config option** failed identically —

```
ACP session/set_config_option failed agent=qwen-code config=model value=benchflow-deepseek-deepseek-v4-flash: -32603
```

(`qwen-code`, `kilo`, `dimcode`; their manifests even document the gap.)

## Root cause

`_model_selection_owned_by_env` only fires when `BENCHFLOW_LITELLM_MODEL_VIA_ENV=1` — and that flag was set **only in hardcoded per-agent branches** (`codex-acp`, `openhands`, `claude-agent-acp`) in `_wire_litellm_agent_env`. Generic manifest agents that deliver the model via env (`supports_acp_set_model = false` + a `BENCHFLOW_PROVIDER_MODEL` env mapping) never got it, so the capability-first dispatch drove their advertised `model` option — which these agents validate against their *own* model catalog and reject.

## Fix (surgical, common — no per-agent code)

Set the flag for **any** agent whose registration declares the env-owned-model shape:

```python
if cfg and not cfg.supports_acp_set_model and cfg.env_mapping.get("BENCHFLOW_PROVIDER_MODEL"):
    updated[LITELLM_MODEL_VIA_ENV] = "1"
```

The manifests already state this in data. Negative space preserved: agents with `supports_acp_set_model=true` or without a model mapping are untouched (they still need ACP model config so they don't fall back to their own defaults) — the hardcoded branches remain as-is.

## Tests

- generic flag derivation + 2 negative controls (`tests/test_litellm_model_via_env_generic.py`)
- end-to-end dispatch: a manifest-shaped agent advertising a `model` option gets **neither** `set_config_option` **nor** `set_model` (`tests/test_acp_model_config_dispatch.py`)

10 pass. Live verification on the census rerun to follow in the PR thread.

### DIFF
--- a/src/benchflow/providers/litellm_runtime.py
+++ b/src/benchflow/providers/litellm_runtime.py
@@ -1060,6 +1060,24 @@ def _wire_litellm_agent_env(
             LITELLM_MASTER_KEY_ENV: master_key,
         }
     )
+    # Generic model-via-env: an agent whose registration maps
+    # BENCHFLOW_PROVIDER_MODEL into an agent-native env var AND declares
+    # supports_acp_set_model=False states, in data, that launch/env config owns
+    # model selection. Set the via-env flag so the ACP layer
+    # (_model_selection_owned_by_env) skips driving set_model AND any
+    # capability-advertised model config option — several agents (qwen-code,
+    # kilo, dimcode, ...) validate foreign model ids against their own catalog
+    # and reject the gateway alias with -32603. This generalizes the hardcoded
+    # codex-acp/openhands/claude-agent-acp branches below to every agent that
+    # declares the env-owned-model shape (e.g. the benchflow-ai/agents
+    # manifests), instead of growing the special-case list per agent.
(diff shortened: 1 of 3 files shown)


Yiminnn:
Live verification (daytona, citation-check, deepseek-v4-flash, with-skill) — all three census failures recovered with this fix:

| agent | before (census) | after (#879) |
|---|---|---|
| qwen-code | `-32603 set_config_option`, 0 acp / 0 llm | **reward 1.0** (acp 46 / llm 26) |
| kilo | same | **reward 1.0** (acp 27 / llm 10) |
| dimcode | same | runs healthy — reward 0.0, acp 53 / llm 19 |

One registration-data-derived flag, three agents recovered, zero per-agent code.

PR #888, opened 2026-07-03
fix(harvey): map provider env to the OpenAI-compatible adapter

Yiminnn:
## What

`harvey-lab-harness` ended every proxied run on **turn 0 with zero LLM activity** (`suspected_api_error`). Root cause: its `env_mapping` was empty, but Harvey LAB's **OpenAI-compatible adapter** — the one that serves every proxied provider (deepseek, the azure gpt-5.4-mini gateway, the `benchflow-*` proxy aliases) — reads `OPENAI_BASE_URL` / `OPENAI_API_KEY`, which only exist as `BENCHFLOW_PROVIDER_*` on the proxy path. With neither set, the adapter hit `api.openai.com` unauthenticated and the harness loop made zero calls.

Map `BENCHFLOW_PROVIDER_BASE_URL → OPENAI_BASE_URL` and `BENCHFLOW_PROVIDER_API_KEY → OPENAI_API_KEY` so the adapter points at the gateway. Direct `ANTHROPIC`/`GEMINI` runs are unaffected (those adapters read `ANTHROPIC_API_KEY`/`GOOGLE_API_KEY` directly and ignore `OPENAI_*`).

## Verification (daytona, deepseek)

harvey now **routes and issues requests** (`total_requests >= 1`) — previously `total_requests == 0` / zero activity. This corrects the long-standing "harvey is a harness-level zero-activity defect" classification: it was an unwired adapter, not a broken harness. The remaining HTTP 500s in the run are transient upstream provider errors (benchflow retries them), orthogonal to this fix.

## Tests

`test_registry_invariants.py`: new `test_harvey_maps_provider_env_to_openai_compatible_adapter` guards the mapping. 184 pass; ruff clean.

### DIFF
--- a/src/benchflow/agents/registry.py
+++ b/src/benchflow/agents/registry.py
@@ -721,9 +721,21 @@ AGENTS: dict[str, AgentConfig] = {
         launch_cmd=f"HARVEY_LABS_ROOT=/opt/harvey-labs /opt/benchflow/harvey-lab-venv/bin/python {_BENCHFLOW_BIN_PREFIX}/harvey-lab-acp-shim",
         protocol="acp",
         requires_env=[],  # inferred from model at runtime (ANTHROPIC_API_KEY, etc.)
-        # env_mapping intentionally empty — Harvey LAB adapters read
-        # provider-specific env vars (ANTHROPIC_API_KEY, OPENAI_API_KEY,
-        # GOOGLE_API_KEY) directly; auto_inherit_env propagates these.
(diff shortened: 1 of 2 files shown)


Yiminnn:
The failing `parity` check is resolved by companion **benchflow-ai/agents#53**, which syncs harvey's manifest env_mapping to this core change (env_mapping is a data field the agents repo owns). Merge #53 first (or together) and parity goes green — verified locally: `test_manifest_byte_identical_to_core` passes for harvey with both changes.

PR #910, opened 2026-07-10
fix(sandbox): prevent Daytona heartbeat sleep leak

bingran-you:
## Summary

- replace signal-based heartbeat teardown with a stop-file and bounded polling
- prevent the race that leaves `sleep 60` orphaned with captured pipes open
- pin the regression guard to the heartbeat path introduced by PR #894

## Root cause

The fast-command path could send `TERM` after the heartbeat shell spawned `sleep 60` but before it stored the child PID. The trap then exited without killing the sleep process. That orphan retained stdout/stderr, so `subprocess.run(..., timeout=2)` timed out even though the parent shell had already returned 0.

## Validation

- `uv run pytest -q tests/test_daytona_dind_compose_up.py` — 12 passed
- 100 fast-command iterations on macOS — passed
- 100 fast-command iterations in Debian/Linux — passed
- forced 2-second heartbeat emission — passed
- `uv run ruff format --check src tests tools`
- `uv run ruff check src tests tools`
- `uv run ty check`
- `uv lock --check`
- `uv run python -m pytest tests/ -q` — 4835 passed, 8 skipped, 7 deselected
- `pip-audit` — no known vulnerabilities, one documented ignore


bingran-you:
@kywch @Yiminnn @xdotli — this is a narrow release-blocking process-cleanup fix. All required CI and the real rollout smoke are green; a human approval is the remaining merge gate.

PR #920, opened 2026-07-11
Restore OpenHands xhigh propagation after PR #911

bingran-you:
## Summary

- restore `LLM_REASONING_EFFORT` in the OpenHands settings JSON
- preserve proxy-route enforcement while preventing OpenHands default `high` from overriding requested `xhigh`
- restore the PR #911 regression test

## Production evidence

A 72-row GPT-5.6-SOL/OpenHands batch recorded `reasoning.effort=high` on every audited provider response even though both xhigh env variables were present. The only material difference from the known-good canaries was that those canaries ran before commit `9c114d3` removed the OpenHands-side setting. Direct Azure probes up to concurrency 16 returned xhigh on every successful response, isolating the downgrade to the agent/proxy request path rather than the provider.

## Validation

- `uv run pytest -q tests/test_agent_registry.py` (26 passed)
- focused Daytona PTY timeout tests (2 passed)
- `uv run ruff check src/benchflow/agents/registry.py tests/test_agent_registry.py`
- `git diff --check`

A fresh Daytona xhigh canary will gate resuming official experiment rows.

### DIFF
--- a/src/benchflow/agents/registry.py
+++ b/src/benchflow/agents/registry.py
@@ -932,6 +932,10 @@ AGENTS: dict[str, AgentConfig] = {
             'printf \',"base_url":"%s"\' "$LLM_BASE_URL"; fi; '
             'if [ -n "$LLM_API_VERSION" ]; then '
             'printf \',"api_version":"%s"\' "$LLM_API_VERSION"; fi; '
+            'if [ -n "$LLM_REASONING_EFFORT" ]; then '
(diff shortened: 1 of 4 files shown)


bingran-you:
Superseded by https://github.com/benchflow-ai/benchflow/pull/921. PR #920 inherited pre-merge PR #911 history; the replacement is a clean two-file diff from current main.

PR #923, opened 2026-07-11
Restore optional skill prompt nudge

Yiminnn:
## Summary

- Reverts #908 to restore the optional `BENCHFLOW_SKILL_NUDGE=name|description|full` prompt injection.
- Keeps native `with-skill` mounting unchanged.
- Restores the original focused regression tests and documentation.

## Why

OpenCode receives the native skill catalog but did not activate the relevant task skill in matched SkillsBench canaries. This restores the explicit opt-in nudge as a separate harness configuration; runs without the environment variable remain unchanged.

## Validation

- `uv run python -m pytest tests/test_internet_policy.py tests/test_skill_policy.py` (50 passed)
- `uv run ruff check` on restored source/tests
- `uv run ruff format --check` on restored source/tests

### DIFF
--- a/.agents/skills/benchflow/SKILL.md
+++ b/.agents/skills/benchflow/SKILL.md
@@ -230,7 +230,8 @@ bench eval run \
   --agent claude-agent-acp \
   --sandbox daytona \
   --skills-dir skills/ \
-  --skill-mode with-skill
+  --skill-mode with-skill \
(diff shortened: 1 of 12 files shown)


Yiminnn:
Closing: this revert is experiment-only. The branch remains available for the isolated OpenCode+nudge canary and is not proposed for main.

PR #932, opened 2026-07-15
Fix OpenHands direct execution outside root workdirs

bingran-you:
## Summary
- apply the opt-in OpenHands subagent-disable patch through the root execution plane before sandbox-user privilege drop
- pass a copied launch environment with the duplicate user-space patch disabled
- cover `/app` workdirs so ACP startup no longer depends on chowning `/root`

## Validation
- `uv run pytest -q tests/test_acp.py tests/test_agent_registry.py tests/test_acp_setup_failure_propagation.py tests/agents/test_manifest_parity.py` (122 passed, 2 skipped)
- `uv run ruff check src/benchflow/acp/runtime.py tests/test_acp.py`
- `uv run ruff format --check src/benchflow/acp/runtime.py tests/test_acp.py`
- `uv run ty check src/benchflow/acp/runtime.py`

Follow-up to #921. The prior launch-time mutation ran as the sandbox user and failed on task images whose workdir was not `/root`, leaving empty ACP and no model calls.

### DIFF
--- a/src/benchflow/acp/runtime.py
+++ b/src/benchflow/acp/runtime.py
@@ -108,7 +108,7 @@ async def _prepare_openhands_direct_execution(
         "p.write_text(s.replace(old,new,1)) if old in s else None; "
         "assert new in p.read_text()'"
     )
-    result = await env.exec(patch_cmd, timeout_sec=30)
+    result = await env.exec(patch_cmd, user="root", timeout_sec=30)
     return_code = getattr(
         result,
         "return_code",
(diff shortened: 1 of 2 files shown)


bingran-you:
Additional local validation after the root-user follow-up commit `832d0a20f`:

- `uv run pytest -q tests/` → **4899 passed, 16 skipped, 7 deselected**
- focused OpenHands pre-launch tests → **2 passed**
- Ruff check/format and `ty` remain green

The fix is intentionally runtime-only; manifest parity remains green because the published OpenHands launch contract is unchanged.

PR #938, opened 2026-07-28
fix(sandbox): stop with-skill builds poisoning no-skill rollouts

bingran-you:
A with-skill build and a no-skill build of the same task share one Docker image tag. Whichever builds last decides what **both** start from — so a no-skill rollout can run inside an image that already contains `/skills`, and the agent reads the mentor guidance the condition exists to withhold.

## The bug

`--skills-dir` rewrites the task Dockerfile in place ([`_inject_skills_into_dockerfile`](src/benchflow/sandbox/setup.py)):

```dockerfile
COPY _deps/skills /skills/
RUN mkdir -p /home/agent/.opencode && ln -sf /skills /home/agent/.opencode/skills
```

But the tag came from the task name alone ([`docker.py`](src/benchflow/sandbox/docker.py)):

```python
main_image_name=_sanitize_docker_image_name(f"bf__{environment_name}")
```

Both modes therefore resolve to `bf__<task>:latest`.

## Evidence

Same task, same host:

| Check | Result |
|---|---|
| Clean build of the task's own Dockerfile | `SKILL_MD_COUNT=0` |
| Shared tag **before** any with-skill run | `COUNT=0`, `/skills` empty |
| Shared tag **after** a with-skill run | **`COUNT=4`** — all four skills at `/skills/` |
| No-skill rollout **in isolation** | container `COUNT=0` ✅ |
| No-skill rollout **alongside** a with-skill one | container **`COUNT=4`** ❌ |

The affected rollout was resolved correctly and still leaked:

```
skill_mode=no-skill  include_task_skills=False  skills_sandbox_dir=None  effective_skills_dir=None
```

Both deployment paths behaved: `_activate_step_skills` returned early (`skills_dir: null`) and `deploy_skills` no-opped. The skills were already in the image, put there by a *different* rollout.

Its trajectory then shows the agent loading `ion-shuttling-mentor`, `surface-ion-trap-bem` and `invariant-based-transport` from `/home/agent/.opencode/skills/`, and solving in **31 tool calls** a task whose uncontaminated no-skill condition had previously failed.

## Why it went unnoticed

`total_skill_invocations` reports `0` whether or not the agent reads a `SKILL.md` off disk — it was `0` for the with-skill runs too. A contaminated run reports an ordinary pass and no counter disagrees.

This silently corrupts the primary no-skill capability metric, and **a task author cannot prevent it** — the task's own Dockerfile is clean.

Note this is *not* the familiar "task bakes its own skills via `COPY skills …`" mistake. That one is the task's to fix; this one isn't.

## The fix

1. **Tag by build context.** `_build_context_fingerprint()` hashes the Dockerfile bytes plus the staged `_deps/skills` layout, and the tag becomes `bf__<task>__<fingerprint>`. Builds whose contents differ can no longer share a tag. Hashing what is actually built also means an unrelated Dockerfile edit correctly yields a new image.
2. **Verify rather than assume.** `assert_no_skill_isolation()` runs after `deploy_skills` on the agent path when the recorded mode is `no-skill`. It probes the agent's own `skill_paths` (following symlinks) plus `/skills`, and raises instead of letting a contaminated run be recorded. Only absolute paths ending in `/SKILL.md` count, so shell noise can't produce a false positive.

Belt and braces on purpose: (1) removes the cause, (2) means a future regression fails loudly instead of quietly returning a wrong number.

## Tests

`tests/test_no_skill_image_isolation.py` — 7 cases:

- fingerprint stable for identical contexts
- **skill injection changes the fingerprint** (the core regression)
- changing the skill payload changes the fingerprint
- isolation check passes on a clean sandbox, and probes both the agent path and `/skills`
- isolation check raises when a `SKILL.md` is reachable
- handles the oracle path (no agent config)
- ignores non-skill stdout

## Verification

- 577 related tests pass (`-k "docker or skill or install or image"`); `ruff check` clean and formatted.
- `tests/test_trial_install_agent_timeout.py` initially failed against an early version of the guard — its env double returns canned stdout for every `exec`. That surfaced a real over-permissiveness in my check rather than a test problem, so the guard now matches only `SKILL.md` paths; that suite is green.

## Reviewer notes

- The tag gains a 12-hex suffix. Anything pinning the literal `bf__<task>:latest` externally would need updating — I found no such assumption in-tree, but it's the change worth a second pair of eyes.
- The guard is agent-path only. The oracle path is supported by the function but not yet wired, since the oracle legitimately runs with skills available.


Galius5136:
I reproduced the original contamination on the merge-base (`cf65617`) and verified the fingerprinted image fix on `d1b0446` with real Docker: after a with-skill build, the no-skill image resolves with `SKILL_MD_COUNT=1` on the base and `0` on the PR head.

On head, the no-skill image also stays clean when the with-skill image is built first, so the result no longer depends on build order. The 7 new isolation tests pass locally.

I found one additional issue around inbound manifests.

For a foreign task without an explicit `docker_image`, `manifest_from_task_config()` still synthesizes:

`bf__<slug>:latest`

while `DockerSandbox` on this PR now builds:

`bf__<slug>__<fingerprint>`

On the merge-base those naming paths were compatible; on HEAD they structurally diverge.

More importantly, `manifest.image` is not only provenance: `resolve_manifest_image()` treats it as the sandbox run target, sandbox setup assigns it to `env_config.docker_image`, and that path is treated as prebuilt, skipping the task Dockerfile build.

On attribution: the prebuilt-skip itself predates this PR. On the merge-base the synthesized name and the built name coincided (verified for `leak-probe`, `org/name`, `Org/Name__subtask`, and a leading-digit name), so the coupling was latent and survivable. What changes here is that a fingerprinted tag can never coincide, turning that latent coupling into a structural mismatch.

This also bears on the reviewer note that no in-tree assumption on the literal `bf__<task>:latest` was found.

I tested a minimal candidate fix locally:

- do not synthesize `manifest.image` for Dockerfile-built foreign tasks;
- leave `image=None`;
- populate `base_image` from the task Dockerfile;
- let the sandbox backend own the built image tag;
- preserve an explicitly configured `docker_image` unchanged.

With that patch:

```text
manifest.image              : None
manifest.base_image         : 'alpine:3.20'
resolve_manifest_image()    : None
skips the Dockerfile build? : False
docker main_image_name      : 'bf__leak-probe__b4cd0ada1ef2'
prebuilt preserved          : True
```

Local validation after the patch: **17 targeted tests passed**, and `ruff check` / `ruff format --check` pass on the 3 modified files.

Note that the patch requires updating `test_synthesizes_local_tag_when_no_prebuilt`, which asserts the old contract directly.

I have not run a full Harbor task E2E, so the runtime effect on Harbor is still inferred from the setup path rather than observed in a real Harbor execution.

One design detail I left unresolved deliberately: when no `environment_dir` is available, the candidate patch currently falls back to `base_image="scratch"`. That avoids synthesizing a run target, but deciding whether `environment_dir` should instead be required or the manifest contract relaxed seems better left to the maintainer.

PR #939, opened 2026-07-30
fix(metrics): count opencode skill invocations

bingran-you:
## Problem

`n_skill_invocations` is **0 for every opencode rollout**, including ones that
demonstrably loaded a skill. opencode's ACP skill call looks like this:

```json
{"type": "tool_call", "kind": "other", "title": "skill", "status": "completed",
 "content": [{"type": "content", "content": {"type": "text",
   "text": "<skill_content name=\"polar-electrostatics-mentor\">\n# Skill: ..."}}]}
```

`is_skill_invocation_event` misses it four ways over:

1. `kind` is `"other"`, not the canonical `"skill"`.
2. The event carries no `tool_name` / `name` / `function` key, so
   `_event_tool_name` returns `""`.
3. The title check used an inline literal `{"invoke_skill", "activate_skill"}`
   which dropped the `"skill"` spelling that `_SKILL_TOOL_NAMES` — used by the
   `kind` check three lines above — already contains.
4. Content sniffing only recognized the OpenHands `Tool: invoke_skill` header.

This matters more than a cosmetic counter: `n_skill_invocations` is the field
that answers *"did the agent actually use the skill?"*, and it is the control
variable for every with-skill vs no-skill comparison. Reading a hard `0` for an
entire harness makes those comparisons unfalsifiable — you cannot tell a skill
that didn't help from a skill that was never loaded.

## Fix

Recognize both opencode signals, preserving the module's existing rule that a
tool declaring a real ACP kind is never reclassified from its title or output:

- `<skill_content name="...">` counts only when it **opens** a structured
  tool-result text block (anchored `\A`, exactly as the OpenHands header is), so
  a `grep` whose output quotes the tag is not counted.
- Bare `title == "skill"` is honored **only for unclassified kinds**
  (`""`/`other`/`tool`) — the same gate content sniffing sits behind. The
  explicit `invoke_skill` / `activate_skill` spellings stay trusted outright, so
  existing behavior is unchanged. A `read` or `execute` tool titled `skill`
  (a filename, say) is still not a skill invocation.

## Verification

Replayed against **206 recorded opencode trajectories** from a 29-task,
6-arm evaluation:

| | before | after |
|---|---|---|
| counter agrees with an independent `<skill_content name=...>` scan | 0 / 206 | **206 / 206** |
| with-skill rollouts reporting a nonzero count | 0 | **56** |
| no-skill rollouts falsely counted | 0 | **0** |

The no-skill invariant that backstops the text-derived path still holds.

New tests cover the opencode envelope, a quoted-tag negative, and a
real-kind tool titled `skill`. `tests/test_skill_invocation_artifacts.py`
passes 10/10 (7 pre-existing unchanged); 697 tests pass across the
skill/metrics/trajectory suites. `ruff check` and `ruff format --check` clean.

### DIFF
--- a/src/benchflow/trajectories/metrics.py
+++ b/src/benchflow/trajectories/metrics.py
@@ -8,6 +8,15 @@ from typing import Any
 
 _SKILL_TOOL_NAMES = frozenset({"invoke_skill", "activate_skill", "skill"})
 
+# ``skill`` is unambiguous as a tool *kind* or tool *name*, but a title is
+# display text — a file or command merely named "skill" can produce it. A title
+# therefore proves a skill invocation outright only under the explicit
+# spellings; the bare word is honored for unclassified kinds alone, behind the
+# same gate as content sniffing (see ``is_skill_invocation_event``). Deriving
+# the subset here keeps the two sets from drifting apart — the inline-literal
+# drift is how PR #597 lost the ``skill`` spelling in the first place.
+_EXPLICIT_SKILL_TOOL_NAMES = _SKILL_TOOL_NAMES - {"skill"}
+
 # OpenHands legacy ACP artifacts render an invoke-skill *result* as a text block
 # that begins with the tool header (``Tool: invoke_skill``) and carries a
 # ``[skill: <name>]`` result marker. Anchoring the header to the start of the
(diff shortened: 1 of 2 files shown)


Morkeeth:
Corroboration from a second adapter, in case it helps this land.

I hit the same defect on claude-agent-acp while measuring a skill library on SkillsBench. That adapter emits a launch as `{"type": "tool_call", "kind": "other", "title": "Skill"}` with the body in a content block reading `Launching skill: <name>`. Each such event reports `n_skill_invocations: 0` on the shipped release. The metric reads a library that fired as one that never did — the failure mode you describe.

I checked this branch against that event shape before writing. It already fixes it: `count_skill_invocations` returns 1 for the ACP event under your change, and still returns 0 for a bash or read tool merely titled "skill".

Worth recording why placement matters. I first tried the naive fix — widening the title branch to `_SKILL_TOOL_NAMES`. It counts the ACP event, but also any bash/read tool titled "skill", because that branch sits above the gate confining content inference to unclassified kinds. Your version puts the title check below that gate: on the ACP event both return 1; on a bash tool titled "skill" the naive version returns 1 where yours returns 0. Your placement is right and mine was a regression — so this supports your PR rather than offering an alternative.

If a second regression test helps, no test covers this shape yet:

```json
{"type": "tool_call", "kind": "other", "title": "Skill",
 "content": [{"type": "content", "content": {"type": "text",
              "text": "Launching skill: document-extraction"}}]}
```

Happy to open it as a follow-up PR against your branch or lift it directly.

Two adapters affected rather than one seemed worth adding given this has been open since 30 July. Thanks for the fix.

Drafted with AI assistance; the probes and event capture are from my own runs.

Galius5136:
Reviewed `be6a1a03` and ran a live A/B with the same task, prompt, agent, model and skill, only the revision changed.

On the base (`cf656179`), OpenCode really invokes the skill, but the trajectory records it as `kind: "other"` and `n_skill_invocations` stays at `0`.

On the PR head, the same call is canonicalized to `kind: "skill"` and `n_skill_invocations` becomes `1`.

I also ran a negative control where the skill was deployed but never invoked, and it correctly stayed at `0`.

I checked the shape reported in #998 as well, including the export variant where the title becomes `skill {"name": "..."}`. This branch handles both cases correctly, so it appears to resolve #998 as well.

Local checks were clean as well: 10/10 on the PR test file, no regressions in the event-shape A/B, and `ruff`, `ruff format` and `ty` all pass.

Only small non-blocking notes from my side: the regression-test docstrings still don't reference the guarded PR/commit, the `<skill_content>` matcher is intentionally narrow around `name="..."`, and the `claude-agent-acp` case is still title-dependent.

@bingran-you from my side this looks ready to land.

PR #949, opened 2026-08-08
fix(rollout): hard deadline around the rollout lifecycle + verifier zero-output-timeout retry

xdotli:
Two harness reliability fixes from a live incident (2026-08-07/08 Daytona runs, prime-agent × deepseek-v4-flash, 25-task job):

## 1. Host-side hard deadline around every rollout (`rollout/_deadline.py`)

One rollout (bike-rebalance) wedged for **11+ hours** and froze the whole `bench eval run` job. Timeline from artifacts: agent turn done 21:51, verifier done 22:03, trace sync 00:59, then stuck forever — i.e. the hang was in **post-verify teardown/export**, below both `agent_idle_timeout` and the task wall-clock budget (Daytona PTY/session calls with no client-side bound: observed siblings "PTY readline timeout (900s)", "Failed to create PTY session", empty "Failed to execute session command").

`Rollout.run()` now enforces a computed hard deadline around its whole lifecycle — covering **all nine call sites** (Evaluation, Runtime, bf.run, continue_run, acceptance, self_gen), not just the eval path:
- deadline = enforced agent budget (honors `RolloutConfig.timeout` and per-role overrides) × turns + user-loop rounds × (agent + soft-verify) + verifier + build + `effective_install_timeout()` per distinct agent + fixed margin — a backstop that can only fire when an await is stuck outside every phase-level timeout;
- a trip cancels with a **bounded grace** for the cancellation-triggered cleanup and then abandons the task outright — a bare `asyncio.wait_for` cannot break the observed hang because it waits unbounded for `finally: cleanup()` while cleanup IS the wedged path (regression test included that fails under bare `wait_for`);
- surfaces a normal `INFRA_ERROR` result (retryable under the standard policy); operator control via `BENCHFLOW_ROLLOUT_HARD_DEADLINE` (seconds; `off`/`none`/non-positive disables);
- Daytona teardown micro-bounds: PTY `kill`/`disconnect` (15s each) and the client side of remote env-file cleanup (30s) — the exact calls that hang on a dead websocket.

## 2. Verifier retry on the zero-output timeout wedge (`rollout/_setup.py`)

Budget-killed rollouts were losing their scores to `reward=None` "verifier timed out" errors. Root cause is NOT a skipped verifier (verify-after-timeout was already unconditional): the verifier exec **wedged before test.sh ever started** — empty `test-stdout.txt` after burning the full budget, while the same verifiers finish in <1s when they run. `_verify_rollout` now retries the verifier **exactly once** when a timeout carries the zero-output signature (killing any late-started orphan `test.sh` first). A timeout WITH output — a genuinely slow/hung verifier — is never retried.

## Validation

- New tests: `tests/test_rollout_hard_deadline.py` (10: computed/override/user-loop budgets, wedged lifecycle, re-wedged cleanup, bounded external cancel, retry-policy integration, e2e eval path), `tests/test_verifier_wedge_retry.py` (3).
- Sweeps: 906 rollout/eval/verifier tests pass; full unit suite green modulo 3 pre-existing environmental failures at HEAD (local creds/bashrc).
- Structural review applied (enforcement layer, budget canonicalization via `effective_install_timeout`, dead-code deletion: net −2 lines of non-test code vs the naive fix).

Known residual (out of scope, noted in review): a truly abandoned attempt can leak a live Daytona sandbox; a label-keyed orphan reaper would close that separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Start with [Getting started](./docs/getting-started.md), then [Concepts](./docs/
 | Run an eval on an existing task | [Getting started](./docs/getting-started.md) |
 | Understand how BenchFlow runs *any* benchmark (the three-layer model) | [Run any benchmark](./docs/running-any-benchmark.md) |
 | Have an AI agent install + run the quickstart end to end | [Agent quickstart prompt](./docs/agent-quickstart.md) |
+| Run an agent from the public agents repo (goose, qwen-code, prime-agent, …) | [Running external agents](./docs/external-agents.md) |
 | Understand Rollout / Scene / Role / Verifier | [Concepts](./docs/concepts.md) |
 | Author a new task | [Task authoring](./docs/task-authoring.md) |
 | Author a task in the native `task.md` format | [Native task.md authoring](./docs/task-authoring-task-md.md) |
--- a/docs/external-agents.md
+++ b/docs/external-agents.md
@@ -1,0 +1,97 @@
+# Running external agents
+
+BenchFlow's built-in registry covers a handful of agents (`bench agent list`).
+Everything else — goose, qwen-code, prime-agent, the omnigent harnesses, … —
+lives in the public **[benchflow-ai/agents](https://github.com/benchflow-ai/agents)**
+repo and loads into BenchFlow through one of four paths. For most users the
+first one is all there is to know.
+
+## 1. Zero-config remote autoload (the default)
+
+Using an agent name BenchFlow doesn't recognize triggers a one-shot fetch of
+the declarative manifests from `benchflow-ai/agents@main`. Nothing to install
+or configure:
+
+```bash
+pip install 'benchflow[sandbox-daytona]'
+export DEEPSEEK_API_KEY=sk-...          # credentials for your --model provider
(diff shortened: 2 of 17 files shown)


xdotli:
Added a third fix found by running the mock-upstream wire-parity protocol end-to-end on Daytona: the gateway's global `drop_params: True` was silently stripping `thinking` and `reasoning_effort` on `openai/` passthrough routes — verified with a capture proxy between the gateway and api.deepseek.com (agent sent both, upstream received neither), meaning gateway-routed runs used the provider's default thinking config while native runs used the agent's. Fixed via per-route `allowed_openai_params`; anthropic-messages routes untouched. Tests + live re-verification in flight.

xdotli:
Live re-verification complete: the first-cut allowed_openai_params approach broke on the pinned LiteLLM 1.89.0 (no such deployment param — every call 500'd); replaced with native deepseek/ routing + an extra_body lift in the proxy pre-call hook, each step verified against a capturing upstream. The adaptation-parity mock protocol now PASSES end-to-end on Daytona: multi-turn byte-identical, thinking + reasoning_effort confirmed reaching the upstream.

PR #914, opened 2026-07-11
chore(release): start 0.6.6 development

bingran-you:
## Summary

- move `main` from the published `0.6.5` final version to `0.6.6.dev0`
- refresh `uv.lock`
- restore the internal-preview version line after the public release

## Validation

- `tools/release_version.py internal-preview --run-number 123` -> `0.6.6.dev123`
- `pytest tests/test_release_version.py` -> 22 passed

PR #974, opened 2026-08-12
chore: start 0.6.8 development

kywch:
## Summary

- advance project version from `0.6.7` to `0.6.8.dev0` after public `v0.6.7` release
- keep lockfile project metadata in sync

## Why

`main` remained on final version `0.6.7` after the release. Final-version commits make the internal-preview workflow skip publication, so the four post-release commits have not produced `0.6.8.devN` previews.

After merge, successful `main` test and integration workflows can publish `0.6.8.dev<run-number>` again. Tag `v0.6.7` remains unchanged.

## Validation

- `uv lock --check`
- `uv run --with packaging --no-project python tools/release_version.py internal-preview --run-number 12345` → `publish=true`, `version=0.6.8.dev12345`
- `uv run --extra dev python -m pytest tests/test_release_version.py tests/test_release_workflow_wiring.py tests/test_workflow_action_pinning.py` → 54 passed
- `git diff --check`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "benchflow"
-version = "0.6.7"
+version = "0.6.8.dev0"
 description = "Multi-turn agent benchmarking with ACP — run any agent, any model, any provider."
 readme = "README.md"
 requires-python = ">=3.12"
--- a/uv.lock
+++ b/uv.lock
@@ -329,7 +329,7 @@ wheels = [
 
 [[package]]
 name = "benchflow"
-version = "0.6.7"
+version = "0.6.8.dev0"
 source = { editable = "." }
 dependencies = [
     { name = "agent-client-protocol" },


leehanchung (approved):
lgtm

Yiminnn (approved):
Verified against a local checkout of the PR branch. LGTM.

**The stated premise is real.** Running the actual validator confirms main is
currently stuck publishing nothing, and that this fixes it:

    compute_internal_preview_version("0.6.7",      42) -> publish=False
    compute_internal_preview_version("0.6.8.dev0", 42) -> publish=True, "0.6.8.dev42"

That matches the model documented in `docs/release.md`.

**Checks I ran:**

- `uv lock --check` passes on the PR branch — the lockfile is in sync, which
  matters because `uv sync --locked` is a blocking gate in `test.yml`, both
  `integration-*.yml`, and `manifest-parity.yml`.
- No test breaks. `__version__` resolves via `importlib.metadata`;
  `test_cli_misc.py` and `test_acceptance_live_execution.py` compare against it
  rather than a literal, and `test_dataset_registry.py` monkeypatches. Nothing
  pins `0.6.7`.
- `CITATION.cff` correctly left untouched — it tracks the last *released*
  version (0.6.7 / 2026-08-09), and the public-cut PR owns that bump (#968 did
  exactly that). Precedent #944 left it alone too.
- Shape matches its precedent, #944 ("start 0.6.7 development").

**One note for the record, not a blocker.** main already carries #966
(`feat(task)!: rename the task-config 'environment' key to 'sandbox'`) — a
conventional-commit `!` breaking change with a **BREAKING** entry in CHANGELOG
`[Unreleased]`. `release-manager/SKILL.md` says to bump main to
`0.<next>.0.dev0`, i.e. the next *minor*, which would make this `0.7.0.dev0`.
Confirmed with the author that `0.6.8` is intended, so proceeding — but the
`release/0.7` branches still sitting at `version = "0.6.2"` and the SKILL.md
wording disagreeing with actual practice (#914/#944/#974 all bump the patch)
are worth reconciling separately.

**Follow-up, separate PR:** CHANGELOG `[Unreleased]` is missing `### Fixed`
entries for #970 (Daytona transport failures) and #971 (frozen token counter),
both user-facing fixes landed after 0.6.7. (#969 is `ci(...)`, correctly
omitted.) Worth backfilling before the 0.6.8 cut.

PR #979, opened 2026-08-12
fix(gemini): strip Google model prefix

### LINKED ISSUE #975: Gemini ACP passes google/ provider prefixes into Gemini CLI model IDs
## Summary

When BenchFlow 0.6.7 is given a provider-qualified Gemini model ID such as:

```text
google/gemini-3.1-flash-lite-preview
```

for the `gemini` ACP agent, the `google/` prefix survives BenchFlow model normalization. Gemini CLI then treats the entire string as the Google model name. Its source deterministically constructs this GenerateContent model path:

```text
models/google/gemini-3.1-flash-lite-preview
```

Live runs return HTTP 404 before any agent work, consistent with that malformed path.

## Environment

- BenchFlow 0.6.7
- `@google/gemini-cli` 0.42.0 (BenchFlow registry pin)
- LiteLLM 1.91.0
- Linux x86_64
- Reproduced with both Docker and Daytona sandboxes

## Symptom

The same evaluation failed in both environments:

```text
ACP error 404: {"error":{"message":"b''","type":"None","param":"None","code":"404"}}
```

Both raw trajectories contain only the initial user message and are byte-identical. No model response or tool call is recorded.

## Deterministic reproduction

This source-level reproduction does not call Google or require a valid key:

```python
from benchflow.acp.runtime import _format_acp_model, _resolve_acp_model_input

model = "google/gemini-3.1-flash-lite-preview"

# Gemini gateway wiring removes BENCHFLOW_LITELLM_MODEL_ALIAS before ACP
# configuration, so ACP model selection falls back to the requested model.
agent_env = {}
acp_input = _resolve_acp_model_input("gemini", model, agent_env)

print(strip_provider_prefix(model))
print(_format_acp_model(acp_input, "gemini"))
```

BenchFlow 0.6.7 prints:

```text
google/gemini-3.1-flash-lite-preview
google/gemini-3.1-flash-lite-preview
```

Gemini CLI 0.42.0 uses `@google/genai`. For non-Vertex requests, its model transformer preserves values beginning with `models/`; otherwise it returns ``models/${model}``. Therefore the final ACP value above deterministically becomes:

```text
models/google/gemini-3.1-flash-lite-preview
```

The corresponding bare input produces the correct path shape (provider success is a separate question):

```text
gemini-3.1-flash-lite-preview
-> models/gemini-3.1-flash-lite-preview
```

## Root cause

`strip_provider_prefix()` strips only prefixes registered in `PROVIDERS`. `google` is not registered, so `google/gemini-...` passes through unchanged.

The `gemini` agent has no `provider/model` ACP format. `_format_acp_model()` therefore returns that unchanged value.

This is a BenchFlow model-normalization/agent-routing mismatch: a models.dev-style provider prefix is accepted for auth inference but is not converted to the bare model ID required by Gemini CLI.

## Impact

- Gemini models supplied as `google/gemini-*` fail through the `gemini` ACP agent.
- Failure is sandbox-independent; Docker and Daytona produce the same error and trajectory.
- Bare `gemini-*` inputs are not affected by this prefix mismatch.

Credential validity is independent: malformed model routing can be reproduced without a provider call, and fixing credentials does not change the generated path.

## Direction

Normalize `google/gemini-*` inputs to the bare Gemini model name at the BenchFlow-to-Gemini ACP boundary, while keeping provider-qualified formats for agents that explicitly require them.

## Related issues

- #600 concerns duplicated API-version path segments (`/v1beta/v1beta/...`) for versioned Gemini base URLs. This issue is distinct: the duplicated component is the `google/` provider prefix inside the model resource path.

No exact duplicate was found by searching open and closed issues for Gemini, `google/gemini`, model prefix, and 404 reports.

kywch:
## Summary

Normalizes `google/gemini-*` to the bare Gemini model ID only when configuring the Gemini ACP agent.

## Root cause

Gemini CLI expects a bare model name. BenchFlow retained the models.dev `google/` prefix, causing Gemini CLI to construct an invalid `models/google/gemini-*` path.

## Impact

Provider-qualified Gemini model IDs work with the Gemini ACP agent; provider/model agents retain their existing prefix behavior.

## Validation

- `uv run python -m pytest tests/test_acp.py -k 'TestConnectAcpModelSelection'`
- `uv run ruff check src/benchflow/acp/runtime.py tests/test_acp.py`
- `git diff --check`

Fixes #975.

### DIFF
--- a/src/benchflow/acp/runtime.py
+++ b/src/benchflow/acp/runtime.py
@@ -267,6 +267,8 @@ def _format_acp_model(model: str, agent: str) -> str:
             return f"litellm/{model}"
         return model if find_provider(model) else bare
     if not agent_cfg or agent_cfg.acp_model_format != "provider/model":
+        if agent == "gemini" and bare.startswith("google/gemini-"):
+            return bare.removeprefix("google/")
         return bare
     # Already has a slash — assume it's provider/model already
     if "/" in bare:
--- a/tests/integration/agent_judge.py
+++ b/tests/integration/agent_judge.py
@@ -130,6 +130,7 @@ _ACP_WRITE_KINDS = {"edit", "delete", "move", "write", "create"}
 # tampering. Only the command after ``: $ `` is the agent's actual action, so the
 # execute scan must strip the description first.
 _ACP_EXECUTE_PREFIX_RE = re.compile(r".*?: \$ ", re.DOTALL)
+_ACP_EDIT_PATH_RE = re.compile(r"\bEditing\s+(?P<path>\S+)\s*$")
 
 
 def _acp_execute_command(title: str) -> str:
@@ -155,7 +156,8 @@ def _acp_write_target(title: str) -> str:
     stripped = title.strip()
     prefix = "file_editor:"
     if not stripped.startswith(prefix):
-        return title
+        match = _ACP_EDIT_PATH_RE.search(stripped)
+        return match.group("path") if match else title
     # Titles carry trailing prose after the JSON payload (observed live:
     # ``file_editor: {...}: Editing /tmp/test_rnn.py``), so parse the LEADING
     # object with raw_decode instead of json.loads — a whole-title fallback
--- a/tests/test_acp.py
+++ b/tests/test_acp.py
@@ -1123,7 +1123,11 @@ class TestConnectAcpModelSelection:
             # No prefix at all — unchanged.
             ("test-agent", "claude-sonnet-4-6", "claude-sonnet-4-6"),
             # Gemini CLI expects a bare model ID, unlike models.dev agents.
-            ("gemini", "google/gemini-3.1-flash-lite-preview", "gemini-3.1-flash-lite-preview"),
+            (
+                "gemini",
+                "google/gemini-3.1-flash-lite-preview",
+                "gemini-3.1-flash-lite-preview",
+            ),
         ],
         ids=["vllm-hf", "zai", "bare-hf", "vertex", "no-prefix", "gemini-google"],
     )
--- a/tests/test_acp.py
+++ b/tests/test_acp.py
@@ -1110,22 +1110,24 @@ class TestConnectAcpModelSelection:
 
     @pytest.mark.asyncio
     @pytest.mark.parametrize(
-        "model_in, expected_model",
+        "agent, model_in, expected_model",
         [
             # Registered vllm/ prefix stripped; HF org/model intact — this is
             # what pi-acp and other ACP agents need for downstream routing.
-            ("vllm/Qwen/Qwen3.5-35B-A3B", "Qwen/Qwen3.5-35B-A3B"),
-            ("zai/glm-5", "glm-5"),
+            ("test-agent", "vllm/Qwen/Qwen3.5-35B-A3B", "Qwen/Qwen3.5-35B-A3B"),
+            ("test-agent", "zai/glm-5", "glm-5"),
             # Bare HF ID (no registered prefix) passes through unchanged.
-            ("Qwen/Qwen3-Coder", "Qwen/Qwen3-Coder"),
+            ("test-agent", "Qwen/Qwen3-Coder", "Qwen/Qwen3-Coder"),
             # Vertex ADC provider — prefix stripped like any other registered one.
-            ("anthropic-vertex/claude-sonnet-4-6", "claude-sonnet-4-6"),
+            ("test-agent", "anthropic-vertex/claude-sonnet-4-6", "claude-sonnet-4-6"),
             # No prefix at all — unchanged.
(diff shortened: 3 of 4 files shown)


Galius5136:
I reproduced the model-prefix behavior from #975 locally and compared the merge-base (`29c64e7`), the isolated Gemini fix commit (`771be99`), and the current PR head (`bc9a1c3`).

On the merge-base, `gemini + google/gemini-2.5-pro` stays as `google/gemini-2.5-pro`. With the Gemini fix, it becomes `gemini-2.5-pro`, and the current PR head behaves identically to the isolated fix commit.

I also checked the surrounding non-regression cases: an already-bare Gemini model stays bare, OpenCode still maps `gemini-2.5-pro` to `google/gemini-2.5-pro`, and `google/gemma-*` is left unchanged for Gemini.

The ACP test suite passed on my setup (`94 passed, 1 skipped`), and the four files touched by the PR pass both `ruff check` and `ruff format --check`.

Everything behaved as expected in the offline path I tested. I didn't run a live Gemini CLI call because I don't have a `GEMINI_API_KEY`, so I didn't independently verify the bare model ID against the real Gemini CLI.

kywch:
Thanks @Galius5136!  I found other gemini issues as well, so I'd probably update on this thread.

tulerfeng:
I have Gemini API credit available, so I picked up the piece @Galius5136
explicitly left open here — the live Gemini CLI call against the real provider.
Full write-up and the raw numbers are in #1030; summarising the part that
concerns this PR.

Setup: merge-base `29c64e7` vs this PR head `bc9a1c3`, with
`@google/gemini-cli@0.42.0` (the registry pin) in a container, hitting the live
Google GenerateContent endpoint. The ACP modelId is produced by the real code
path from #975 and handed verbatim to the CLI.

The fix works for the reported case:

| requested model | tree | ACP modelId sent | live gemini-cli 0.42.0 |
| --- | --- | --- | --- |
| `google/gemini-3.1-flash-lite-preview` | `29c64e7` (before fix) | `google/gemini-3.1-flash-lite-preview` | 404 `ModelNotFoundError` |
| `google/gemini-3.1-flash-lite-preview` | `bc9a1c3` (this PR) | `gemini-3.1-flash-lite-preview` | 200 `ok` |
| `gemini-3.5-flash-lite` (already bare) | both trees | `gemini-3.5-flash-lite` | 200 `ok` |

`tests/test_acp.py` gives 94 passed / 1 skipped here, matching what @Galius5136
saw, and `ruff check` / `ruff format --check` are clean on all four touched
files. I also mutation-tested the new case: removing the two added lines makes
`TestConnectAcpModelSelection::test_model_id_selection[gemini-google]` fail with
`Expected: set_model('gemini-3.1-flash-lite-preview')` /
`Actual: set_model('google/gemini-3.1-flash-lite-preview')`, and restoring them
makes it pass, so the test does guard the fix.

One gap. The guard here is `bare.startswith("google/gemini-")`, so `google/gemma-*`
is still passed through prefixed and still 404s live:

| requested model | tree | ACP modelId sent | live gemini-cli 0.42.0 |
| --- | --- | --- | --- |
| `google/gemma-4-31b-it` | `29c64e7` (before fix) | `google/gemma-4-31b-it` | 404 `ModelNotFoundError` |
| `google/gemma-4-31b-it` | `bc9a1c3` (this PR) | `google/gemma-4-31b-it` | 404 `ModelNotFoundError` |

Bare `gemma-4-31b-it` returns 200 from the same CLI, so this is the same bug in a
different model family rather than an unsupported model.
`_MODELSDEV_PROVIDER_HEURISTICS` already maps `("gemma", "google")`, so BenchFlow
itself produces that shape in the reverse direction. #1030 matches on `google/`
instead and normalizes both families; I verified it returns 200 for
`google/gemma-4-31b-it`.

Worth noting the two PRs are not in conflict: applying this PR's
`tests/test_acp.py` diff on top of #1030 passes (14 passed in
`TestConnectAcpModelSelection`). The integration-style assertion here is the
stronger guard of the two, so the combination that seems best is #1030's
normalization with this PR's test.

Separately, `bc9a1c3` ("fix(judge): allow scratch edits in native titles") is
unrelated to the Gemini prefix and touches `tests/integration/agent_judge.py`.
It looks fine on its own — `tests/test_judge_robustness.py` gives 55 passed —
but it might be easier to review as its own PR.

All of the above is end-to-end validation against the live Gemini API rather
than an offline reproduction.

kywch:
Closing unmerged in favor of #1030, which supersedes this Gemini ACP normalization and covers Gemma too. This PR’s stronger `connect_acp` / `session.set_model` regression coverage has been ported there for both Gemini and Gemma, and #1030 now carries `Fixes #975`. The unrelated judge change continues in #1069. Thanks @Galius5136 and @tulerfeng for the offline and live validation.

PR #980, opened 2026-08-12
fix(acp): retain live subprocess stderr

### LINKED ISSUE #976: ACP subprocess stderr is lost when an agent returns an empty end_turn
## Summary

BenchFlow does not persist subprocess stderr when an ACP agent returns a protocol-valid empty `end_turn` and remains alive.

The proven loss path in BenchFlow 0.6.7 is:

1. The ACP subprocess writes a diagnostic to stderr.
2. It returns a valid empty `end_turn` on stdout and remains alive.
3. `ContainerTransport` accepts the protocol response; it captures only non-protocol stdout in the agent log.
4. `SubprocessLiveProcess.readline()` reads stderr only after stdout reaches EOF.
5. Because the response is valid and stdout remains open, the EOF diagnostic path never runs.
6. Normal transport shutdown terminates the subprocess without draining or persisting stderr.

The rollout artifact therefore contains the empty agent turn but not the diagnostic that explains it.

## Impact example

Two BenchFlow 0.6.7 runs used:

- agent: `claude-agent-acp`
- model route: `zai/glm-5.1`
- same task and prompt
- environments: Docker and Daytona

Both agent phases ended after approximately 30 seconds. Their raw trajectories were byte-identical:

```json
[
  {"type": "user_message", "text": "<prompt>"},
  {"type": "agent_message", "text": ""}
]
```

Both runs had zero model activity and no persisted agent diagnostic log. Later narrow probes succeeded, so the original cause is unknown and should not be attributed to Z.AI, LiteLLM, the agent, or either sandbox. Lost subprocess stderr prevents further attribution.

## Source evidence

`src/benchflow/acp/container_transport.py` writes non-JSON-RPC **stdout** lines to `agent/<agent>.txt`. Its `LiveProcess` interface exposes only `readline()` for the stdout protocol stream; it has no stderr-drain or stderr-artifact path.

`src/benchflow/sandbox/process/_base.py::SubprocessLiveProcess.readline()` reads up to 8192 bytes of stderr only inside:

```python
if not line:
    # stdout reached EOF
    stderr_bytes = await self._process.stderr.read(8192)
```

That helps when the process/transport closes. It cannot capture a diagnostic written before a valid response when the process keeps stdout open.

`SubprocessLiveProcess.close()` terminates the process and waits for exit, but does not drain or persist stderr before discarding the process handle.

## Proposed deterministic regression

Use a fake ACP subprocess that:

1. Completes `initialize` and `session/new` normally.
2. On `session/prompt`, writes a unique non-secret sentinel to stderr.
3. Returns a valid empty `end_turn` with no content or tool events.
4. Keeps stdout and the process alive until BenchFlow closes the transport.
5. Completes normal rollout cleanup.

Assert the sentinel is present in a redacted per-rollout diagnostic artifact. This reproduces the stderr-loss mechanism without a live provider credential or assumptions about the original GLM failure.

## Impact

- Protocol-valid empty turns can lose their only actionable diagnostic.
- Intermittent agent/shim/runtime failures cannot be attributed after teardown.
- Docker and remote-sandbox comparisons may show identical empty trajectories while hiding each subprocess's stderr.
- Upstream bug reports lack evidence needed to identify the failing component.

## Direction

Continuously drain subprocess stderr into a bounded, redacted per-rollout artifact for the subprocess lifetime, including protocol-valid responses and normal shutdown.

As a secondary improvement, BenchFlow could report an empty ACP turn with zero captured model requests as a distinct diagnostic state. That is separate from preserving stderr and is not required to reproduce this bug.

## Related work

- #535 is the closest related issue: agent-log capture can omit agent output, though this issue concerns subprocess stderr during a protocol-valid response.
- #504 added structured diagnostics after a transport reaches EOF. This issue is distinct because a valid `end_turn` is received while stdout remains open, so the EOF path does not run.

## Duplicate search

Searched open and closed issues for `empty ACP`, `empty end_turn`, `empty agent message`, `agent stderr`, and `agent log capture`. No issue found covering stderr loss when a live ACP subprocess returns a valid empty turn.

kywch:
## Summary

Captures bounded stderr from subprocess-backed ACP agents throughout their lifetime and persists a redacted tail when transport closes.

## Root cause

BenchFlow read stderr only after ACP stdout reached EOF. A protocol-valid empty `end_turn` can leave stdout open, so diagnostics written to stderr were discarded at normal teardown.

## Safety

- 64 KiB stderr tail cap
- redact before rollout artifact and typed transport diagnostic persistence
- close waits at most two seconds for an inherited stderr FD, then cancels the drain

## Validation

- `uv run python -m pytest tests/test_acp.py tests/test_process.py`
- `uv run ruff format --check src tests`
- `uv run ruff check src tests`
- `git diff --check`

Fixes #976.

### DIFF
--- a/src/benchflow/acp/container_transport.py
+++ b/src/benchflow/acp/container_transport.py
@@ -7,6 +7,7 @@ from typing import Any, TextIO
 
 from benchflow.sandbox.process import LiveProcess
 from benchflow.sandbox.process._base import _ANSI_CSI_RE, _ANSI_OSC_RE
+from benchflow.trajectories.types import redact_trajectory_text
 
 from .transport import Transport, decode_json_rpc_message
 
@@ -138,3 +139,7 @@ class ContainerTransport(Transport):
             self._agent_log_file.close()
             self._agent_log_file = None
         await self._cp.close()
+        stderr = getattr(self._cp, "stderr_tail", "")
+        if isinstance(stderr, str) and stderr and self._agent_log_path:
+            with self._agent_log_path.open("a") as agent_log:
+                agent_log.write(redact_trajectory_text(stderr))
--- a/src/benchflow/sandbox/process/_base.py
+++ b/src/benchflow/sandbox/process/_base.py
@@ -31,6 +31,8 @@ logger = logging.getLogger(__name__)
 
 _BUFFER_LIMIT = 10 * 1024 * 1024  # 10MB readline buffer
 _DIAG_TRUNCATE = 2000  # max chars for diagnostic stderr in error messages
+_STDERR_TAIL_LIMIT = 64 * 1024  # bounded stderr retained for rollout diagnostics
+_STDERR_DRAIN_TIMEOUT_SEC = 2
 _BOOTSTRAP_DONE = "__BENCHFLOW_BOOTSTRAP_DONE__"
 _ENV_KEY_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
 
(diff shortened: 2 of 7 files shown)


Galius5136:
I reproduced the stderr-loss path from #976 locally on WSL using the deterministic fake-ACP regression described in the issue, and compared the merge-base (29c64e73) with the PR head (d769f347). I drove ContainerTransport directly (both with a local subprocess and the real DockerProcess) rather than running a full bench eval run.

On the base, the ACP response completes normally with end_turn, but stderr is not retained and the agent log is not created.

With #980, the ACP response stays unchanged, while the stderr diagnostic is preserved in the agent log. I also verified the same behavior through the real DockerProcess path.

I checked the bounded stderr tail as well: it stays at 65536 bytes, drops the head, and retains the tail. Secret redaction also worked as expected: the diagnostic sentinel was present, REDACTED was present, and the original AIzaSy... token prefix was absent.

Everything behaved as expected on my setup, and I didn't observe any protocol regression or shutdown hang.

PR #984, opened 2026-08-13
feat(trajectories): canonical trace IR with ACP and ATIF converters

Galius5136:
# Trace interoperability: OTel ↔ ATIF ↔ ACP

**Draft.** This PR tracks the whole trace-interoperability task and lands it in
reviewable slices. Ten are included. OpenTelemetry is read but not written:
the inbound edge is implemented, and the emitter is deferred rather than
rejected — §8.12 of `docs/trace-interop.md` records why, and records it as our
own decision.

**One opt-in wiring site, and nothing else reaches a run path.** Everything
below is inert unless `BENCHFLOW_VIEWER_TRACE_IR=1` is set: `render_rollout`
gains a single branch with a lazy import inside it, and with the switch off the
trajectory viewer emits the bytes it emitted before — verified on six pages,
same SHA-256 before and after. `export_atif.py` is untouched and still the only
writer of `trainer/atif.json`, no artifact changes, and a test names the one
file allowed to import the IR family so a second importer fails the suite.
Reverting the wiring commit unwires it completely.

## Goal

OTel ↔ ATIF ↔ ACP trace interoperability.

## What is included

| | | |
|---|---|---|
| **A** | ACP-session capture events, formalized | JSON Schema, docs, conformance suite |
| **A2** | ACP → ATIF preservation invariants | what today's export keeps and drops, as tests |
| **B** | Canonical Trace IR v0 | typed, versioned, with an explicit loss-reporting contract |
| **C** | `ACP → IR` | the inbound edge, every absence declared |
| **D** | `IR → ATIF` | the outbound edge — byte-for-byte parity with the direct exporter |
| **E** | `ATIF → IR` + round-trip measurement | closes the loop, and measures what it costs |
| **F** | `OTLP/JSON → IR` | the inbound OTel edge, built against the versions `uv.lock` pins |
| **G** | `IR → ACP capture events` | the outbound ACP edge — fail-closed, and human-verified on two real rollouts |
| **H** | `IR → viewer trace steps` | the step list a viewer page renders; a tool category by membership, never inferred from a string |
| **I** | `viewer trace steps → a rendered page` | built against `viewer.py` as it stands on `main`, opt-in, verified in a browser |

Each slice has its own comment on this PR with the details and the verification.

## Why an IR at all

From reconnaissance of the current tree:

- **ACP-session capture events had no schema.** `trajectory/acp_trajectory.jsonl`
  is read by the viewer, judges, the skill evaluator, task verifiers and the
  review skill, with no version field and no written specification. Each consumer
  parses it independently.
- **ATIF is export-side only.** Nothing under `src/` reads `trainer/atif.json`
  back; its shape is pinned by prose in a module docstring.
- **OpenTelemetry is not an active representation.** No dependency, no OTLP, no
  `gen_ai.*` handling. The one module that existed was removed as unwired.
- **Three exporters already walk the same event list through shared helpers**,
  each with its own private answer to the same questions — what happens to a tool
  call with no arguments, whether a thought boundary survives, whether a timeout
  is representable. Those answers already diverge (`docs/trace-interop.md` §5).

A hub makes each format one edge to a written contract, and makes the
information loss a **value** rather than a comment.

## What the round trip measures

Slice D shows the hub reproduces the direct exporter byte for byte, which says
the IR is *sufficient* for ATIF. Slice E asks the more useful question — how much
of a trace is still there after a trip through the format — and answers it as a
measurement over the two real rollouts in `docs/trace-interop.md` §5.2:

- **Within the declared `IR ↔ ATIF` field mapping, nothing representable is lost
  on the two measured rollouts.** Everything the loop drops is dropped because
  ATIF has nowhere to put it, so the remaining loss belongs to the format rather
  than to these converters. That is a result and not a definition: on a trace
  populating every IR field the same measurement reports two representable
  losses, which is a gap in our own edge rather than a cost of the format.
- **The trace comes back with more values than it left with** — 46 in, 56 out for
  one rollout — while having lost information. `arguments` leaves declared
  `SYNTHESIZED` and returns unmarked, so the reconstructed trace asserts the tool
  *was observed* to be called with none. The information was not so much lost as
  overwritten with a plausible value of the same shape, and only the pair of loss
  reports still carries the difference.
- **A timeout costs five fields and the event carrying them** — the timeout
  reason, the run's own outcome status, and the three fields the marker carried.
  §5 loss #4, measured instead of asserted.

That last point is the argument for the hub in the one form an assertion cannot
take: it quantifies what a format costs instead of claiming it.

Full tables are in `docs/trace-interop.md` §8.10. These are machine measurements
over real captured rollouts; verification status for each slice is documented
separately in its PR comment.

## Roadmap

- [x] **Slice A** — formalize the existing ACP-session capture events
- [x] **Slice A2** — `ACP → ATIF` preservation and loss invariants
- [x] **Slice B** — canonical IR v0: typed representation, loss contract, invariants
- [x] **Slice C** — `ACP-session capture events → IR`
- [x] **Slice D** — `IR → ATIF`, with parity against the direct exporter
- [x] **Slice E** — `ATIF → IR`, and the `ACP → IR → ATIF → IR′` measurement
- [x] **Slice F — OpenTelemetry, inbound** — `OTLP/JSON → IR`, built against the
      `opentelemetry-proto` / `opentelemetry-semantic-conventions` versions
      `uv.lock` already pins, with no OTel dependency added
- [x] **Slice G — `Canonical Trace IR → ACP capture events`** — **implemented and
      HUMAN VERIFIED** (G1–G10). The target is the ACP capture **event format**
      of §2.1, not the whole `acp_trajectory.jsonl` artifact, which has no
      contract and holds records from four producers. `ACP → IR → ACP`
      reproduces its input — structural equality and key order — on the two real
      captured rollouts of §5.2. The edge is **fail-closed**: when the target
      requires semantics the trace does not carry it raises rather than emitting
      a partial trajectory. `name_semantics` gates the `kind` slot, so an ATIF
      `function_name` or an OTel `gen_ai.tool.name` cannot be laundered into an
      ACP kind — a category tag — even when the value happens to spell one.
      `ORACLE` and `UNKNOWN` have no record shape in the contract and are
      refused. No runtime wiring, and no file is written. **`OTel → Viewer` is
      not supported**: an OTel trace is still four independent kinds of refusal
      away from exporting
- [ ] **`IR → OTel`** — **deferred**, see §8.12: no outbound OTel consumer and no
      contract to build an emitter against exist yet. A decision of ours, current
      and reversible; open question 2 below is a separate matter and still open
- [x] **Loss-bounded conformance gate** over the measured round trip —
      **implemented and HUMAN VERIFIED** (H1–H8), `docs/trace-interop.md` §8.13.
      An **independent join** between the divergence the round trip *observes*
      and the loss each converter *declares*: the measurement still reads only
      the two traces, the converters still produce their own reports, and the
      gate is a third module that relates them. An **undeclared fabrication is
      always a violation** — no allowlist, including for the exporter's own
      structural metadata, which `IR → ATIF` now declares like anything else.
      The rule asks for **`SYNTHESIZED` specifically, not merely for some
      record at the path**: a `NORMALIZED` declaration does not excuse an
      invented value, since there is no source value being reshaped. Where a
      value is honestly declared in `TARGET` space and observed at a hub path,
      a small **target→hub bridge** relates the two; it is **empirically
      checked against real values and is not an exemption** — remove the
      declaration and the gate fails anyway. Divergence caused by events being
      fused away is explained **per value, not per path**, so an edit to an
      event that *survived* the trip is still caught. No runtime wiring. **No
      OTel round trip**, because `IR → OTel` remains deferred (§8.12) and a
      gate cannot measure a loop that does not exist
- [x] **Slice H — `Canonical Trace IR → viewer trace steps`** — implemented and
      HUMAN VERIFIED, `docs/trace-interop.md` §8.14. Steps only, deliberately
      not a payload: four of a viewer payload's five fields are functions of the
      rollout directory rather than of the trace. A tool's display category is
      emitted **only** when the source said the string *is* a category and the
      value is already a member of the display vocabulary — membership, never
      inference — so an ATIF `function_name` of `execute` and an OTel
      `gen_ai.tool.name` of `read_file` stay neutral. Every event becomes
      exactly one step: `ORACLE` and `UNKNOWN` land on a diagnostic step
      carrying a serialization of the **canonical IR event**, declared as such
      because the IR holds no source record
- [x] **Slice I — `viewer trace steps → the page the current viewer renders`** —
      implemented and **HUMAN VERIFIED in a browser**, §8.15. Built against
      `viewer.py` as it stands on `main`, not against the reviewer-grade viewer
      proposed in #1034: that branch is a design reference for the step
      vocabulary and nothing else — no code is imported, vendored or fetched
      from it — so this works whether or not it lands. A pure refactor first
      (`_page`, `_prompt_block`, `_message_block`, `_thought_block`,
      `_result_block` lifted out of `_render_acp_events`, byte-identical
      output), then an adapter that renders a step list through those cards,
      then one opt-in branch. Steps are **never rebuilt into ACP capture
      events**: the IR holds records ACP has no type for, and forging them would
      launder back the assumption the hub removes. Measured against the current
      page on the two captured rollouts: an `agent_timeout` reaches no card
      there and the word "timeout" appears nowhere, an unrecognized record
      reaches no card, and a tool card carries kind/title/status but not the
      tool's output — all three are that renderer's four branches rather than
      bugs, and this edge places its six step kinds and declares the difference
- [x] **Wiring + human E2E** — done for the viewer path, opt-in and reversible

## Scope / open questions

**The canonical IR and the OpenTelemetry direction are still not approved
decisions.** Everything above is built so it can be reviewed as code and deleted
in one commit if the answer is no. The questions, in the order they block work:

1. **Canonical IR direction — sanity check.** Is an `N ↔ 1` intermediate
   representation the shape you want, or would you rather see direct converters?
   Three exporters walking the same event list through shared helpers is what
   suggested it, but that is an inference, not a mandate.
2. **OpenTelemetry: ingest, emit, or both?** The removed collector was inbound
   only, and it was removed deliberately. Reviving a receiver and emitting spans
   are different projects with different risk. **This is the one blocking the
   next slice.**
3. **Capture-layer enrichment: in scope or follow-up?** The ACP protocol carries
   `rawInput`, `rawOutput` and `locations` on tool calls; `handle_update` reads
   five fields and drops the rest, which is why every exported format emits empty
   tool arguments. Per-event timestamps go the same way. Closing those gaps
   changes an on-disk format several consumers already parse, so it is a
   compatibility decision rather than a converter one.
4. **Artifact-level `acp_trajectory.jsonl` contract.** Should the file carry a
   `schema_version`? And are the `oracle` record and the session-factory
   passthrough part of the trajectory contract, or separate concerns that happen
   to share a filename?

Happy to split this into separate PRs per slice if that reviews better.


Galius5136:
## Slice A complete

Formalizes the ACP-session capture-event subset written to
`acp_trajectory.jsonl`. Zero runtime behaviour change: no runtime module imports
the schema, and nothing under `src/benchflow` changes behaviour.

**What landed**

- `src/benchflow/trajectories/schemas/acp-capture-event-v1.schema.json` —
  Draft 2020-12, one event per document, three record shapes across five event
  types.
- `docs/trace-interop.md` — the format as emitted, producers and consumers, the
  state of ATIF and OpenTelemetry, and the information losses observable in the
  existing conversions. Claims are marked FACT or PROPOSAL.
- `tests/trajectories/test_acp_capture_event_schema.py` — 25 tests. The
  event-type vocabulary is read out of `_events_to_trajectory` by AST rather
  than hardcoded, so adding a branch there fails the suite until the schema
  documents it.
- `jsonschema` declared as a dev dependency (previously used only as a litellm
  transitive).

**Design choices worth flagging for review**

- `kind` is an **open string**, not an enum. `_canonical_tool_kind` passes
  agent-supplied values through unchanged, and values outside `ToolKind` reach
  disk in practice — including the literal `tool`, the fallback when a
  `tool_call_update` arrives for an id that was never opened. Constraining it
  would reject data the emitter can produce.
- `status` **is** a closed enum: the serialized value is always
  `ToolCallStatus(...).value`, with an `in_progress` fallback for anything
  unparseable, so no out-of-vocabulary value can reach disk.
- `additionalProperties: false` on every variant. This is what makes the schema
  a description rather than a formality: a field the emitter does not produce
  today fails the suite, so a future capture-layer change cannot land without
  updating the schema and the docs in the same PR.

**Human verification** (run by @Galius5136, not by CI)

- **H1 — positive path.** A real `ACPSession` driven through the production
  capture path and writer; every line re-read from disk and validated. All five
  event types present and valid. PASS.
- **H2 — oracle boundary.** An oracle-only artifact written through the
  production writer at the production artifact path, using the record shape
  derived from `_run_oracle` by AST. The record is rejected by the schema, which
  is the expected result. PASS. *This did not execute `_run_oracle` against a
  live sandbox — it verified the producer-derived shape, the production writer
  and the schema boundary.*
- **H3 — in-scope negative probe.** A `tool_call` really emitted by the capture
  path validates; the same record re-read from disk, with a single `rawInput`
  field added, is rejected via `additionalProperties`. PASS.

**What this surfaced**

Preparing H2 exposed a real boundary that the first draft of the documentation
got wrong. Oracle mode does not append a record to an ACP trajectory: the
rollout does not run an agent at all, `_run_oracle` builds a new oracle-only
list, and that list becomes the rollout's trajectory. ACP rollouts and oracle
rollouts are mutually exclusive, and no production path emits a mixed artifact.
Separately, session-factory Sessions bypass the ACP-session emitter entirely via
`_snapshot_session_trajectory`, which returns `session.steps` unchanged.

So the same artifact path can hold different record families depending on how
the rollout ran. That is why this schema is scoped to the ACP-session capture
events and does not claim to be an artifact-level contract — and why the
artifact-level questions are listed as open in the PR description rather than
answered here.

**Not included, deliberately**

No canonical IR, no OTel work, no ATIF reader, and no capture-layer enrichment.
No fixes for the conversion losses/divergences surfaced during reconnaissance;
those remain documented and out of scope for this slice.

Galius5136:
## Slice A2 — ACP → ATIF preservation/loss characterization

Turns the ATIF rows of the §5 loss table into executable assertions. 24 tests, one new file, no runtime change, no public format change, no new dependency — same profile as Slice A.

**Deliberately not Slice B.** Slice B in the roadmap above is the canonical-IR prototype, and it is gated on open question 1. This work is a sibling of A: it holds whichever way that question is answered, and it gives a baseline of what the existing converter preserves before anything is redesigned.

- 10 preservation invariants — identity, ordering, referential integrity of `source_call_id` within its own step, textual output, thought text, and conformance to the ATIF validator this repository already ships;
- 12 loss characterizations, each pinned with sentinels asserted absent from the whole serialized document rather than from the one field a test happened to look at;
- 2 producer-boundary tests placing the empty `arguments` at the ACP wire boundary rather than in the converter.

Both sides of the `oracle`-source divergence are read from source via AST — the converter's event-type branches and the validator's accepted `source` set — so changing either fails a test instead of silently invalidating the suite's premise.

### One previously unrecorded loss

Documented as #10: `ThoughtBuffer` joins buffered thoughts with a blank line, so a single thought that already contains one is indistinguishable from two consecutive events, and the thought-event count is unrecoverable. Reachable in production — `_parse_gemini_trajectory` appends one event per entry of a message's `thoughts` list.

### What real rollouts showed

§5.2 records two `gemini` rollouts run through the production path (docker sandbox, ACP transport, the standard artifact writers), inspected by hand:

- **H1, real tool use.** Every `tool_call` capture record carried exactly `type`, `tool_call_id`, `kind`, `title`, `status`, `content`. Ids reached ATIF unchanged, textual output reached `observation`, `kind` became `function_name` (observed `execute`, `read`, `think` — none of them `ToolKind` members), `title`/`status` appeared only in `extra`, every `arguments` was `{}`, and no ISO-8601 value appeared anywhere.
- **H2, real wall-clock timeout.** An `agent_timeout` recorded in `acp_trajectory.jsonl` and counted in `result.json` under `trajectory_summary.event_type_counts` is absent from the ATIF document for the same rollout. Loss #4 confirmed end to end, including that the signal does survive at rollout level.

The observation worth calling out: in those rollouts the proxy capture (`llm_trajectory.jsonl`) carries non-empty tool-call `arguments` — in one case keyed `command`, holding the shell command the agent ran — for the same call whose ATIF `arguments` is `{}`. The inputs were dropped, not unavailable.

This is **not** evidence that the ACP `rawInput` family was on the wire. Those four fields occurred nowhere in the captured artifacts, the proxy capture included, which is expected because that capture is not ACP.

Also observed and now documented: an ATIF document opens with **two** identical `user` steps — one from the `prompts` argument, one from the captured `user_message` event — so a consumer counting user turns over-counts by one.

### Not exercised by a real rollout

Still resting on code reading plus the synthetic tests, and labelled as such in the document: the non-text content-block loss (#5), because neither agent emitted a file-edit or terminal block; and the `oracle` source divergence, because neither rollout ran in oracle mode.

Still a draft — Slices B–F remain a proposal pending the open questions above.

Galius5136:
## Slice B — provisional canonical Trace IR

The four open questions in the PR description are still unanswered. Rather than
leave the task parked, this slice takes a **provisional position on question 1**
— a canonical hub — and implements it in a form that can be reviewed as code and
undone by deleting two files. The position is mine, not an agreed direction, and
the PR stays Draft for that reason.

**Feedback on the architecture is still very welcome, and it is still cheap to
act on:** nothing depends on this module, so redirecting it costs a revert
rather than a migration.

### Why a hub rather than direct converters

Four trace-shaped representations already exist here — the ACP-session capture
events, ATIF, ADP and the Verifiers/ORS record — and all three exporters walk
the same ACP event list from the same call site, `_write_trainer_artifact`. The
cost of pairwise conversion is not the edge count. It is that each edge answers
the same questions privately, and the answers already diverge:

- for one `tool_call` event, ATIF emits `"arguments": {}` and ADP emits
  `"kwargs": {}`;
- ATIF keeps the tool status in a non-standard `extra`; ADP drops it;
- both join `agent_thought` boundaries irreversibly through the same
  `ThoughtBuffer`;
- neither represents `agent_timeout` at all.

Three independent decisions about the same event, taken three times, recorded
nowhere. A hub makes each format one edge against a written contract, and makes
the loss a typed value instead of a comment in a module docstring.

The alternatives I considered and did not take — promoting ATIF or the ACP
capture events to hub status, or extending the capture format instead — are
written up with their reasons in `docs/trace-interop.md` §8.1.

### What is in the slice

`src/benchflow/trajectories/ir.py`, `tests/trajectories/test_trace_ir.py` (25
tests), and `docs/trace-interop.md` §8. No existing format, exporter, artifact
or code path changes, and no new dependency — pydantic is already a runtime
dependency, so `uv.lock` is untouched.

The rule the module is built on: **the IR is a pragmatic superset of what
BenchFlow can observe, not a model of what an agent trace could contain.** Three
consequences do most of the work:

1. **Tri-state optionality.** A value, `None` ("this source never carried it"),
   and an empty value ("carried, and empty") are three different facts.
   `arguments={}` vs `arguments=None` is the case that matters: every ACP-derived
   tool call is the second, and ATIF and ADP both serialize the first — which is
   why their documents read as though every tool was called with no arguments.
2. **Absence must be declared.** A `None` not covered by a `LossRecord` makes the
   trace *invalid*. That is what turns the loss report into a contract instead of
   documentation.
3. **Normalization is never destructive.** `source_type` keeps the source's own
   type string next to the normalized `kind`, and `name_semantics` records that
   an ACP `kind` is a category rather than a function name — the normalization
   the exporters currently perform silently.

The IR deliberately **cannot** fabricate an agent version, a synthetic
`call_{n}` id, a timestamp, or an OTel span id. Those are target-side
obligations; converters produce them and record them as `SYNTHESIZED`.

Two tests pin the isolation claim rather than the design: nothing under
`src/benchflow` imports the IR, and the IR imports no benchflow module. The
first is the executable form of "zero runtime behaviour change" — wiring the IR
into a run path has to update that test deliberately.

The event-kind and tool-status vocabularies are checked against the real
producer: `ToolCallStatus` is read off the enum, and the event types are read out
of `_events_to_trajectory` by AST, the same mechanism Slice A uses. Adding a
branch to the capture path fails this suite until the IR says what that event
becomes.

### Deliberately not included

`ACP → IR` and `ATIF ↔ IR` converters, any OTel work, any wiring into a run
path, any on-disk artifact. I would rather the representation were looked at
before anything depends on it. The OTel mapping in §8.3 is a sketch and is
labelled unverified: no version of the GenAI semantic conventions is vendored
here and nothing validates against it.

### What a review can still change

Everything, and cheaply. §8.6 lists it explicitly — the hub itself, the
declared-absence contract (the strongest opinion in here, and the one most
likely to feel heavy in a converter), `name_semantics` and `reasoning_segments`,
`extensions` as the escape hatch, `TraceUsage.source` (which exists only because
open question 4 is open), the version string, and every name in the module.

What a review cannot change by rejecting the IR: the losses in §5 are properties
of the current code, not of this proposal, and they stay whatever happens here.

### Verification

25 new tests; `tests/trajectories` 288 passed; the trajectory, capture, streaming
and ACP regression lane 453 passed; `ruff check .`, `ruff format --check` and
`ty check src/` clean. These are machine checks — no rollout was run for this
slice, and none is needed: nothing here executes in a run path.

Galius5136:
﻿## Slice C — `ACP → IR`, the first converter

Slice B proposed a hub and a contract: every `None` in the IR is covered by a
`LossRecord`, and a conversion's cost is a typed value rather than a comment in
a docstring. This slice is the first real edge, and its job was as much to
stress that contract as to convert.

Still unwired, still Draft, still provisional. `ir.py` and `ir_from_acp.py` form
a closed family — a test asserts nothing else under `src/benchflow` imports
either, and the converter imports one benchflow module, the IR.

### What the conversion keeps that the existing exporters lose

- **`agent_timeout`** becomes an event with its fields in `extensions` and sets
  the trace outcome. Every exporter drops it today (§5 loss 4).
- **Thought boundaries** survive: one capture record is one reasoning segment,
  and nothing is joined, so the ambiguity `ThoughtBuffer` creates never arises
  (§5 loss 10). A thought whose own text contains a blank line stays one segment —
  splitting it would invent a boundary the source does not have.
- **Non-text content blocks** are carried as `opaque` with the block verbatim
  instead of being skipped by `content_blocks_to_text` (§5 loss 5).
- **`""` and absent stay distinguishable** everywhere; text-empty events are
  kept rather than dropped.
- **The `oracle` record** keeps its own kind and role instead of becoming an
  agent step prefixed `[oracle: …]` that a consumer can only undo by string
  matching.
- **An unrecognized `type`** becomes `unknown` with the record carried verbatim,
  instead of being skipped silently.

It refuses to invent arguments, timestamps, tool-call ids or an agent version.
It also does not prepend the `prompts` argument as leading user events, which
both existing exporters do — §5.2 showed the cost, an ATIF document that opens
with two identical `user` steps so user turns over-count by one. Those steps are
not ACP events; a target that wants them adds them at its own edge as
`SYNTHESIZED`.

### Is "declare every absence" actually affordable?

Measured rather than asserted. The report is **`n_tool_calls + 5` records** and
does not grow with trace length: systemic absences (timestamps, per-event usage,
agent version, stop reason) are declared once each under an unindexed
`events[].…` path, and only `arguments` — which `validate_trace` requires per
event — scales. A test pins this by doubling the non-tool events and asserting
the report does not change at all.

---

## Verification

Two kinds, deliberately not merged.

### Machine checks

35 new tests. Preservation is checked against events produced by driving a real
`ACPSession` through the production capture path (reusing the Slice A2 fixture),
not against hand-written dicts that would only prove the converter agrees with
itself. The loss report is asserted as a complete **set**, so an undeclared loss
and a spurious one both fail. One test removes a declared loss from a converted
trace and shows the trace becomes invalid.

`tests/trajectories` 328 passed; the trajectory, capture, streaming and ACP
regression lane 493 passed; `ruff check .`, `ruff format --check` and
`ty check src/` clean.

### Human end-to-end — run by @Galius5136, not by CI

Against **two real rollouts already captured for Slice A2** (`gemini`, docker
sandbox, ACP transport, standard artifact writers — the ones §5.2 documents),
converted and cross-checked against their own source files.

- **H1 — real tool use. PASS.** The five real events convert in order; both tool
  calls keep id, kind and status, and carry `arguments = None`, while the same
  rollout's `trainer/atif.json` serializes `{}` for those same calls.
- **H2 — real wall-clock timeout. PASS.** The timeout is preserved in the IR
  with its fields; the ATIF document for the same rollout contains no
  `agent_timeout`. §5 loss 4 is closed end to end on the ACP → IR path, on a rollout that really timed out.
- **H3 — hand inspection of a converted document. PASS, after a fix.** See below.
- **H4 — negative control. PASS.** Removing one `LossRecord` makes the trace
  invalid with `absence must be declared`; setting `arguments = {}` makes it
  valid again, which is the known limit recorded below.

### H3 found a real defect, corrected before this comment

Reading the converted document by hand — not running the suite, which was green
— showed that §8.4 published its worked example with `exclude_none=True`, so
`arguments` was absent from the document while the loss report kept a record
addressing `events[1].tool_call.arguments`. **The declaration that legalizes the
absence pointed at a key no reader of that document could find.** The suite could
not catch it because both sides of the comparison used the same non-canonical
encoding, so it was self-consistent.

The fix (`48dbbff3`) states the canonical encoding in the `ir.py` docstring —
**nulls are retained; `exclude_none=True` is not a valid encoding of a Trace IR
document** — regenerates §8.4 so `arguments: null` and its loss record are
visible in the same document, and adds guards that assert in the same test that
the discarded encoding *fails* to resolve those paths, so they cannot pass for
both encodings at once.

Applying that guard immediately found a second instance of the same class:
`outcome.stop_reason` is a loss every ACP conversion declares, and it could not
resolve in a trace that did not time out, because the section itself was `null`.
`CanonicalTrace.outcome` is now always present, like `agent`.

No serializer was added. There is no on-disk artifact yet and a writer would
anticipate an interface this proposal has not earned.

### Known limitation, not a solved problem

The invariant forces a converter to *declare* an absence. **It cannot stop one
from writing `arguments: {}` instead of `null`** — such a trace is valid, and H4
demonstrates it deliberately. Closing that would mean the IR taking a position on
what an empty map means for each source, which the tri-state rule deliberately
leaves to the converter. Recorded in §8.8 as a limit rather than filed as a bug.

### Not established

- **Non-text content blocks** are covered by tests only. Neither rollout emitted
  a file-edit or terminal block, so `opaque` carrying has not been observed in
  the wild — the same gap Slice A2 recorded.
- **Oracle mode** likewise: the `oracle` branch is covered by tests, not by a
  real oracle rollout.
- No new rollout was run for this slice.

---

Feedback on the architecture is still welcome and still cheap to act on: nothing
imports either module, so redirecting this costs a revert rather than a
migration. The open questions in the PR description remain unanswered, and §8.6
still lists what a review can reject — the declared-absence contract first among
them.

Galius5136:
## Slice D — `IR → ATIF`, the first outbound edge

Where `ACP → IR` tested whether every absence could be *declared*, this edge
tests the other half of the taxonomy: ATIF requires values the IR does not
carry, so this is the first converter that has to **fabricate**, and
`SYNTHESIZED` stops being a decorative enum member.

Still unwired and still Draft. `export_atif.py` is untouched and remains the only
writer of `trainer/atif.json`.

### The claim, and how it is checked

    ir_to_atif(acp_events_to_ir(events), prompts=P)
        ==
    trajectory_to_atif_record(events=events, prompts=P)

Parity with the existing direct exporter, for the document. A hub that lost
anything the direct path preserved would fail that equality.

Checked **three ways, with no allow-list suppressing differences**: the real
`trainer/atif.json` a rollout wrote months ago, the direct exporter re-run today
on the same inputs, and the same inputs through the hub. All three identical, on
two real rollouts.

### `PathSpace` — which document a loss record talks about

The outbound edge forced a gap in the contract into the open. A prompt-derived
step, the `message: ""` ATIF requires, `final_metrics.total_steps` — none has an
IR antecedent, so none can carry a hub path, and inventing one produces an
address that does not resolve.

`LossRecord.space` is now `hub` · `source` · `target`, defaulting to `hub`, with
`field` the path *inside* its space and no prefix repeating it. Three spaces
cover every direction because **every edge has the IR on exactly one side** and
therefore exactly one non-hub space — OTel will add none.

Only `hub` records compose across edges, and that is the property the whole
proposal is for:

| edge | space | field | class |
|---|---|---|---|
| `acp -> ir` | hub | `events[2].tool_call.arguments` | `unsupported` |
| `ir -> atif` | hub | `events[2].tool_call.arguments` | `synthesized` |

One field, two edges, one path: the source never carried arguments and the
target demanded them anyway. `source` and `target` records are terminal by
construction, and there is no unified report — composition happens at read time.

The same asymmetry settles ownership: a trace is built once, so an inbound
conversion may attach its report to it; a trace may be converted to many
targets, so `ir_to_atif` **returns** `(document, LossReport)` and leaves the
input trace byte-identical.

### One deliberate deviation

`acp_events_to_atif_steps` renders an oracle record as a `source: "agent"` step
prefixed `[oracle: …]`, recoverable only by string matching — §5.1 records this
as a live divergence, since the in-repo validator already accepts
`source: "oracle"` while no emitter produces one. The IR carries the role, so
this edge emits `source: "oracle"` with the command as the message. **This is
the only intended difference**, and a test runs both paths over a trajectory
holding every capture event type plus an oracle record and asserts that step is
the only one that differs.

### The loss report describes the trace it receives

Not the trace ACP happens to produce. An ACP-derived trace has no per-event
timestamps or usage, and the inbound report already declared those absences —
re-declaring them here would count one fact twice. The same conversion over a
trace that *does* carry them declares every one, addressed where the value lives
(a tool-call timestamp under the tool call, not under its event).

Both halves are asserted, the second as a complete set. A companion test derives
the field list from the IR models themselves, so a field added to the IR that
ATIF cannot represent fails the suite until its fate is decided.

### Cost

`TraceUsage` gained `cost_usd` and `price_source` before this was published,
because the direct exporter accepts `total_cost_usd` and the IR could not carry
it — the parity claim would have been true only for traces without one. Cost
sits with the token counters because that is where `agent_result`,
`TaskTelemetry` and ATIF's own `final_metrics` already put it. `price_source` has
no ATIF slot and is declared dropped: BenchFlow computes no prices of its own,
so a cost without the table that produced it is not comparable. Per-call cost is
out of scope.

---

## Verification

Two kinds, deliberately not merged.

### Machine checks

45 tests in the ATIF suite; `tests/trajectories` 379 passed; the wider
trajectory / capture / streaming / ACP lane 544 passed; `ruff check .`,
`ruff format --check` and `ty check src/` clean.

One intermittent failure was seen once, in
`tests/test_acp.py::TestIdleTimeoutDiagnostics` — a wall-clock assertion that
passed alone and on re-run, in a module that imports nothing from these modules.
Recorded rather than attributed.

### Human end-to-end — run by @Galius5136, not by CI

**D1, D2, D3 use real rollouts already captured** (the `gemini` runs of §5.2,
docker sandbox, ACP transport, standard artifact writers).

- **D1 — real tool use. PASS.** Three-way comparison identical; 14 outbound
  records; the `unsupported → synthesized` composition above verified on both
  tool calls.
- **D2 — real wall-clock timeout. PASS.** Three-way comparison identical; the
  timeout is preserved as far as the IR and then declared `dropped` at the ATIF
  edge, which is the honest result — 16 outbound records.
- **D3 — hand inspection. PASS.** An independent `diff` of the produced document
  against the rollout's own `atif.json` reports IDENTICAL. `arguments: {}`
  reproduced, `agent.version: "unknown"` reproduced, and the two leading `user`
  steps reproduced — including the duplicate §5.2 documents. Parity means
  reproducing the defect too, not silently fixing it.

**D4 and D5 are constructed controls, and are labelled as such:**

- **D4 — the oracle deviation.** No real rollout ran in oracle mode; the input is
  built by hand.
- **D5 — negative control.** Removing a tool title from the IR flips the parity
  check from `True` to `False`, so the comparison demonstrably discriminates.

### Not established

- **Real cost end-to-end is not verified.** Every rollout artifact available
  carries `cost_usd: null` — including four whose `usage_source` is
  `provider_response`, whose gateway log simply carried no per-entry cost. The
  writing path is established by reading the code; the mapping is covered by
  tests. **D6 is a machine/integration proof and is not a human E2E.**
- **Non-text content blocks** remain covered by tests only — neither rollout
  emitted a file-edit or terminal block.
- **Oracle mode** likewise: constructed input, never a real oracle rollout.
- **The richly-populated trace** used to pin the complete loss set is
  hand-built; no source in the repository produces one yet.
- No new rollout was run for this slice.

---

Feedback on the architecture is still welcome and still cheap to act on: nothing
imports these modules, so redirecting this costs a revert rather than a
migration. The open questions in the PR description remain unanswered, and §8.6
still lists what a review can reject.

Galius5136:
## Slice E — `ATIF → IR`, and what a trip through the format costs

Four commits, `4ec9aaea..56ecab6f`. Still Draft, still unwired: nothing under
`src/benchflow` imports any of it, `export_atif.py` is untouched and remains the
only writer of `trainer/atif.json`.

**What landed**

- `src/benchflow/trajectories/ir_from_atif.py` — the inbound ATIF edge, closing
  the pair with Slice D.
- `src/benchflow/trajectories/ir_round_trip.py` — the `ACP → IR → ATIF → IR′`
  measurement.
- 102 tests; `docs/trace-interop.md` §8.3 gains the mapping table, §8.10 the
  measurement.

**The rule the converter is built on: read what the document says, never what it
probably meant.** Several values in an ATIF document were fabricated by the
converter that wrote it — `agent.version: "unknown"`, `arguments: {}`,
`message: ""` — and nothing marks them as such. Reading them back as absences
would be guessing which ones were invented, and would make the measurement below
report a preservation that did not happen.

**What the measurement found**, on the two real rollouts of §5.2:

- Within the declared `IR ↔ ATIF` field mapping, **nothing representable is lost
  on either rollout**. Everything the loop drops is dropped because ATIF has no
  slot for it. That is a result rather than a definition: on a trace populating
  every IR field the same measurement reports two representable losses —
  per-event usage, which ATIF *does* have a step-metrics slot for and
  `ir_to_atif` does not write. A gap in our own edge, and visible as one.
- **The trace comes back with more values than it left with** — 46 in, 56 out on
  H1 — while having lost information. `arguments` leaves declared `SYNTHESIZED`
  and returns unmarked, so the reconstructed trace asserts the tool *was
  observed* to be called with none. The information was not so much lost as
  overwritten with a plausible value of the same shape, and only the pair of
  loss reports still carries the difference.
- **A timeout costs five fields and the event carrying them** — the reason, the
  run's outcome status, and the three fields the marker carried. §5 loss #4,
  measured instead of asserted.

The report separates what it *observed* (preserved / transformed / lost /
fabricated, from comparing the two traces and never the loss reports) from what
is *declared* (whether ATIF has a slot at all), because a loss with a slot is a
bug to fix and a loss without one is a property of the format, and a single
percentage would merge them.

**Verification.** Machine: 481 tests in the trajectories lane, ruff, `ty`, plus
20 targeted mutations of both converters, all caught. Human, on H1 and H2:
document parity, the conversion read step by step in both directions, every
fabricated value confirmed present in `trainer/atif.json` and absent from the
ACP capture, every unrepresentable value confirmed the other way round, and a
negative control confirming the comparison notices a corrupted value.

**Limits, stated rather than implied:** oracle rollouts and non-text content
blocks appear only in constructed test input; no artifact on hand carries a
non-null cost, so `usage.cost_usd` round-trips in tests only; the
representability table is checked against our own converters, not against a
vendored ATIF schema; and two rollouts from one agent are a demonstration, not a
survey.

**Not in this slice, deliberately:** any wiring into a run path, and
OpenTelemetry — which stays blocked on open question 2 (ingest, emit, or both).
That is the one thing here that needs a maintainer decision rather than more
code.

Galius5136:
## Slice F — `OTLP/JSON → IR`, and a scope decision on the OTel direction

Two commits, `56ecab6f..d4a105f6`. Still Draft, still unwired: no module under
`src/benchflow` imports any of it, no artifact changes, and `uv.lock` is
untouched.

**What landed**

- `src/benchflow/trajectories/ir_from_otel.py` — the inbound OTel edge.
- `src/benchflow/trajectories/_otlp_anyvalue.py` — the `AnyValue`/`KeyValue`
  decoder it is built on, split out as a pure move; it imports nothing from
  benchflow.
- 129 tests; `docs/trace-interop.md` §4.2 and §8.3 updated, §8.11 and §8.12 new.

### Built against the lock file, not against recollection

§4.2 is still FACT — no OTel code, no OTel dependency. But `uv.lock` already
pins `opentelemetry-proto==1.41.1` and `opentelemetry-semantic-conventions==0.62b1`
transitively via `daytona`, so the two questions an OTel reader has to answer
have checkable answers. Both versions are recorded in code, both wheel hashes
were checked against the lock, and **neither package is imported** — the edge
reads JSON dictionaries, so the lock stays untouched.

Reading them corrected four things, three of which the collector removed in
`e84795d7` got wrong: `intValue` is a JSON *string* in the canonical encoding;
an `AnyValue` writes `""`, `0` and `false` explicitly, so its
`stringValue or intValue or ...` read observed values as absent;
`gen_ai.usage.total_tokens` does not exist at 0.62b1 and the cache counters are
spelled `cache_read.input_tokens`; and enums serialize as member names by
default, integers under a flag, with both accepted on parse.

### The rule: a span is evidence of an operation, not a statement about an agent

Exactly one span shape becomes a typed IR kind — the `execute_tool` operation
the pinned vocabulary defines as that. Everything else becomes `UNKNOWN` with its
whole content carried in `extensions.otel`. **So this edge reaches two of the
IR's seven `EventKind` members.** For four of the eight `gen_ai.operation.name`
values there is no IR candidate at all; for `chat`, `generate_content` and
`text_completion` there *is* one, and the blocker is checkable rather than a
matter of taste — filling it means reading `gen_ai.input.messages`, whose
structure the pinned package defines by reference to a JSON schema it does not
ship (129 entries in the wheel, zero `.json` files).

Two more consequences worth naming:

- **Identifiers are carried exactly as written and never re-encoded.** A
  32-character hex trace id fed to the pinned JSON parser is accepted *as base64*
  and silently yields 24 bytes, so hex and base64 are not reliably
  distinguishable and normalizing would make identity depend on a heuristic.
- **Document order is preserved and never sorted by time.** Siblings overlap, a
  batch may omit a parent, and two spans can share a start instant. The
  `parentSpanId` edge set is preserved per span instead, along with the envelope
  coordinates, so the `resourceSpans`/`scopeSpans` partition survives flattening.

### What the loss report costs

31 records on the fixture: 19 systemic, 12 per-event, `19 + 2.4n` as the trace
grows, so only the per-span half grows. `0 synthesized` is **structural** — the
class has no site, because the IR's only required fields are derived from input
and there is no slot an inbound edge could be forced to invent. `0 dropped` is
**a property of that payload, not of the edge**: two fully conformant inputs do
produce `DROPPED` records — spans disagreeing about `gen_ai.request.model`, and
a document carrying both `gen_ai.usage.input_tokens` and the deprecated
`gen_ai.usage.prompt_tokens` with different values.

The contract guard derives the IR field list from the models and requires every
field to be filled or declared **per instance**. It found two real undeclared
absences while the edge was written, and a review of the finished slice found
four more — one of them in the guard itself, which had been field-level, so a
field the fixture happens to fill passed on every payload. 33 targeted
mutations, all caught.

The suite's fixture is not hand-written: it is `MessageToJson` output from the
pinned library, and a script rebuilds it in a throwaway venv and diffs.

### Scope: ingest-only for now

For this implementation OTel is an ingest boundary, and `IR → OTel` is
**deferred, not rejected**. §8.12 records that as a decision of ours — current
and reversible — rather than as something settled.

The reasoning is five things checkable in the tree: no outbound OTel consumer
exists; no exporter or backend is integrated; the OTel support that did exist
was a receiver; the consumer named by the task's own issue body reads
`acp_trajectory.jsonl`; and there is no contract to build an emitter against.
That last one is the load-bearing reason — without a consumer, the span tree for
a rollout, the id encoding and the commitment to an `_incubating` vocabulary
would be decided by this implementation alone, and the hub is the one place a
wrong answer propagates to every format.

Three things §8.12 deliberately does **not** claim: that `IR → OTel` will never
be needed; that the `<->` in the issue title means one direction rather than the
other; or that the task has been narrowed — if emitting is wanted, the work is an
additional edge rather than a redesign. Until a consumer or a contract exists,
the IR gets no span-tree field and no id-minting policy added in anticipation.

Happy to add the emitter whenever it is useful — from here it is one module and
one loss report, the same shape as `ir_to_atif.py`.

### Verification

Human E2E verified against the procedure, F1–F6 all pass — fixture provenance
regenerated from the pinned wheel, the mapping read span by span, all 31 loss
records read individually, the negative controls, the mutation harness, and the
unwired check. Machine: 610 trajectories green, ruff/format/`ty` clean.

Declared limits, unchanged by that: the fixture's **encoding** is authoritative
and its **content** is constructed; **no OTLP payload from a real agent has ever
been read**, because nothing in BenchFlow emits spans; the mapping is checked
against two pinned packages, not the specification; the GenAI conventions are
still `_incubating`; there is no `IR → OTel`; and therefore no OTel round trip
and none is measured.

Happy to split any of this into its own PR if it reviews better.

Galius5136:
## Slice G — `IR → ACP capture events`, and the contract question underneath it

Two commits, `d4a105f6..67ccd12f`. Still Draft, still unwired: nothing under
`src/benchflow` imports it, `_capture.py` is untouched and still the only
producer of the format, no artifact changes, `uv.lock` intact.

**What landed**

- `src/benchflow/trajectories/ir_to_acp.py` — the outbound ACP edge.
- 67 tests, every emitted record validated against Slice A's published schema.
- `docs/trace-interop.md` §2.4 corrected (below), §8.3 gains the mapping and the
  verification record.

### The recon that changed the target

The obvious reading of this slice is "write `acp_trajectory.jsonl`". That turns
out to be the wrong target, and the evidence is in the schema we already
shipped: Slice A's own `description` says **"SCOPE IS NARROWER THAN THE FILE —
this schema does not describe the complete `acp_trajectory.jsonl` artifact"**
and **"no document defines the artifact-level contract, and this schema does not
create one"**.

Checking that against the tree: **four producers write records into that file,
and three of them produce records the schema rejects.** The ACP-session emitter
is valid. `_run_oracle` is not. `hosted_env._row_to_acp_events` is not — it
reuses the ACP type strings while putting the text under `content` rather than
`text`, adds `ts` and `example_index`, and introduces a fourth type, `reward`.
Session-factory `steps` pass through unmodelled. A **fifth** record family
(`session_meta` / `response_item` / …) is *read* by the viewer and by
`traj_report.py` and written by nothing here — `cli/traj.py` identifies it as
Codex.

So the edge targets the part that **is** defined: the capture event format. §2.4
is updated to name `hosted_env` (it listed two other sources; there are three)
and to describe the Codex family as consumer-recognized rather than as a
producer. Whether the artifact should acquire a contract is a separate question
this slice does not open.

### Fail closed

Every record shape is `additionalProperties: false`, with six required fields on
a tool call. This is the first outbound edge that can be *unable* to write, so
it either represents every event or raises `AcpCaptureNotRepresentable` — a
`ValueError` carrying the blocking `LossRecord`s, following
`PrimeSftTrajectoryJsonlError` and the `ValueError` `ir_to_atif` already raises.
A partial event list is indistinguishable from a complete one once it leaves the
function: the target has no envelope, no count, and no marker for "some events
are missing". `acp_capture_blockers()` answers the same question without an
exception, through the same pass, so asking and converting cannot disagree.

Refusals, rather than inventions: a tool call with no ACP `status` (the
vocabulary is closed, every member asserts a lifecycle state, and a fabricated
one would reach the viewer and be shown to a person as an observation); no
`tool_call_id`; no `kind`; a content block with no source block (ACP stores wire
blocks verbatim and two shapes are known to be consumed — choosing one would be
inventing structure); a text event with no text; an incomplete timeout; and
`ORACLE` / `UNKNOWN`, which have no record shape in the contract at all.

`title` is the single value written for an absence, as `""`, and only because
the contract documents that as its own representation of an absent title. It is
declared `SYNTHESIZED` anyway, since a reader cannot tell it from an observed
empty one. `tool_call_id` carries identical documentation and deliberately does
**not** get the same treatment: an empty title is a missing label, an empty id
would turn an absent identity into an observable one and could collapse several
id-less calls onto the same value.

### The defect a review caught, worth reporting because it was semantic

An ACP `kind` is a **category** — `ToolKind` calls itself "Category tag for tool
calls", `_canonical_tool_kind` defaults an absent one to `other`, and the
production values (`execute`, `edit`, `fetch`, `think`) are categories. ATIF's
`function_name` and OTel's `gen_ai.tool.name` name *particular tools*.

The first working version wrote `tool_call.name` into `kind` regardless, so an
ATIF document with a usable status exported `kind="read_file"` — a tool name in
a category slot, with only a `DROPPED` record on `name_semantics` to show for
it. The IR carries `name_semantics` precisely to stop that, and the outbound
edge was ignoring it.

Now only `name_semantics == "acp_kind"` is writable, and **the string is never
inspected**: a `function_name` of `read` collides with a real `ToolKind` member
and is refused anyway, because matching a vocabulary by accident is not being
drawn from it. The record moved `DROPPED` → `NORMALIZED`, since only `acp_kind`
survives and the target's own convention carries it.

Representability is decided by the data, never by the lineage. `Provenance` is
not read, and a test asserts the module never references it.

### Verification

**Human-verified end to end, G1–G10**, against `e2e-g/PROCEDURE.md`. Unlike the
OTel slice, this one has a real producer *and* real artifacts: the two captured
rollouts of §5.2 are still on disk, with `trajectory_source: "acp"` in their
`result.json`. `ACP → IR → ACP` reproduces both exactly — structural equality
**and** key-order preservation — on H1 (5 records) and H2 (4 records, ending in
a real wall-clock timeout).

Also checked by hand: every loss path resolves, no unexpected synthesis, the
laundering path closed including the colliding values, fail-closed atomicity
with no partial prefix, both title cases, provenance independence, 12 records
schema-validated with all four negative controls rejected, and 27/27 mutations
caught with the files restored.

### What this does not claim

- **Not** that every conformant ACP document round-trips. Three corpora were
  measured — H1, H2, and the suite fixture with its drop-one subsets. There is
  no property test.
- **Not** that the `acp_trajectory.jsonl` **artifact** is supported.
- **Not** that `ORACLE` or `UNKNOWN` events are exportable.
- **Not** that ATIF or OTel tool results are generally exportable — their
  content blocks carry `text` and no `raw`, and their names are not ACP kinds.
- **Not** that `OTel → Viewer` works. An OTel trace is four independent kinds of
  refusal away from exporting, and the procedure prints them so this cannot be
  inferred.
- **Not** that anything is wired. This edge returns records; it writes no file.

Happy to split any of this into its own PR if it reviews better.

Galius5136:
## Slices H and I: the viewer path, built against the viewer that exists, verified in a browser

Head `22239d15`. Six commits: `upstream/main` merged as a base update (43 behind
before it, 0 after), a pure refactor of `viewer.py`, the adapter, one opt-in
wiring branch, a fix the human verification turned up, and the verification
record.

```
<format> → ir_from_* → CanonicalTrace → ir_to_view_steps → ir_to_view_html → page
```

The last edge renders a step list through the card builders `viewer.py` already
emits its ACP page with, so an ACP capture, an ATIF document and an OTLP payload
all reach the same renderer without a second viewer existing anywhere.

### Built against `main`, not against #1034

`benchflow-ai/benchflow#1034` is a **design reference for the step vocabulary and
nothing else**: no code is imported, vendored or fetched from it, and every
commit here targets `src/benchflow/trajectories/viewer.py` as it stands on
`main`. If that PR never lands this works unchanged; if it lands, the content
transfers to its typed payload — category by membership, `name_semantics`,
`reasoning`, labelled diagnostics, timeout and unknown events — because none of
it is HTML-specific. This PR does not depend on that one.

### Opt-in, and byte-identical when off

One branch in `render_rollout`, guarded by `BENCHFLOW_VIEWER_TRACE_IR`, with the
import inside the guarded function. Switch **off** — the default — is the ACP
path unchanged: six pages rendered from two captured rollouts, a raw ACP session
file and a Codex session file have the **same SHA-256 before and after** the
whole series. Switch **on** is the canonical path, and it never routes through
`_render_acp_events`: a test patches that function to raise, renders the
canonical page, and then proves the patch bites by rendering the same rollout
with the switch off.

Steps are **never rebuilt into ACP capture events**. The IR holds records ACP has
no type for (`oracle`, anything unknown), a status it cannot spell, and tool
names whose semantics are the point; forging capture events would launder all
three back into the assumption the hub exists to remove.

`export_atif.py` is untouched, no artifact changes, and
`test_only_the_ir_family_imports_the_ir` now carries a `WIRING_SITES` allowlist
naming `viewer.py` and nothing else — a second importer anywhere fails the
suite, and reverting the wiring commit unwires the family completely.

### What the page gains, measured against the current one

On the two captured rollouts of §5.2, `_render_acp_events` and this edge, same
directories:

| | current page | through the IR |
|---|---|---|
| the run that ends in a real wall-clock timeout | **3 cards from 4 events**; the word "timeout" appears nowhere | 4 cards, one a typed timeout |
| an unrecognized record | reaches no card | a card labelled `Canonical IR representation`, body = the canonical event |
| tool output | not on the page at all | shown when the block carried text |
| an ATIF-only rollout directory | `<p>No trajectory files found</p>` | rendered |

None of that is a bug being fixed — those are the renderer's four branches. This
edge has six step kinds to place, places them, and declares the difference. Three
tests pin the current behaviour as a fact so that if a branch is ever added, this
edge's reason for existing is revisited with it.

### Classification is a table, never a substring

`_tool_accent_class` scans the tool kind and then the human title for needles.
It is not imported here, and an AST test asserts the name never appears in the
module. The hue arrives already decided — a category only when the source said
the string *is* one — and is mapped through a table over the eight display hues.

The same run makes the point: `execute` arrives as an ACP `kind` and gets the
execute accent; the identical string in that run's own `trainer/atif.json`
arrives as a `function_name` and stays neutral. OTel's `read_file` and
`write_file` stay neutral for the same reason.

**One incidental observation about the current renderer**, from a real capture
rather than a constructed example: a `think` tool call whose title is
`Update topic to: "Executing System Commands"` is painted with the bash/execute
accent, because the classifier falls back from the kind to the title and finds
the word *Commands*. Not touched here — flagging it because it is the same
mechanism, in your code, on a real rollout.

### Human browser verification

**V1–V12 and V4b, all PASS, in a browser, 2026-08-19.** Legacy-vs-canonical on
the same directories, the timeout, the unrecognized record, the ATIF colour
result, the OTel neutrality, the untyped spans, tool output, the diagnostic
label, the byte-identical switch-off page, `bench eval view` on both paths, and
one request in the Network panel — the document itself, nothing external.

### The verification found a real defect, and the suite had not

`CanonicalEvent.reasoning` was **lost with no loss record** whenever it
accompanied an event kind other than `AGENT_REASONING`. That is exactly what a
real ATIF export produces: `acp_events_to_atif_steps` writes an `agent_thought`
as `reasoning_content` on the agent step it precedes, so reading one back gives
a `TOOL_CALL` event carrying reasoning. On the ACP corpus a thought is its own
event, so the hole never opened and the suite stayed green throughout.

Fixed before the sign-off: an additive `steps[].reasoning`, rendered inside the
card of the step that carried it rather than as an invented second event, plus a
guard that is the real fix — **every string-bearing field of `TraceEvent`, on
every `EventKind`, must reach the emitted step or be named by a record**, with
the field list derived from the model so a new string field fails until given a
disposition. The guard's predicate has its own negative control, and five
mutations (emission, the normalization record, the loss record, the rendering,
the guard) are all caught.

Third time in this work that a green suite was not evidence a document was
right. It keeps being the artifact-reading step that finds these.

### Not claimed

Four corpora, one machine, one browser. The OTLP payload's encoding is
authoritative — it comes from `opentelemetry-proto`'s own `MessageToJson` — but
its **content is a construction**: nothing in BenchFlow emits spans. Oracle
events and non-text content blocks stay test-only, as they have been since §8.3.
`IR → OTel` remains **deferred** on the terms of §8.12, a decision of ours that a
consumer or a contract would end.

The ATIF document's two identical `user` steps are **not** touched: one comes
from the `prompts` argument and one from the captured `user_message`, they are
`acp_events_to_atif_steps`'s doing (§5.2 records the over-count), and the page's
job is to show the document it was given. Our own `ir_to_atif` does not write
them.

Docs: `docs/trace-interop.md` §8.14 and §8.15.

ljr145733:
Hey @Galius5136. Viewer-side notes from [#1034](https://github.com/benchflow-ai/benchflow/pull/1034), since slice I is built against `viewer.py` as it stands on main and [#1034](https://github.com/benchflow-ai/benchflow/pull/1034) moves that ground:

* `viewer.py` becomes a `viewer/` package. `render_rollout`, `_render_acp_events`, `_VIEWER_CSS`, `_WORDMARK_HTML` all move verbatim to `viewer/legacy.py`, with the import surface re-exported from the package `__init__` — so `viewer.render_rollout(...)` keeps working, but your `_trace_ir_page` branch and the factored page primitives won't apply as a clean diff. Whoever merges second rebases by hand; the port itself is mechanical.
* One free win after rebase: `_VIEWER_CSS` is now composed from a shared theme (`viewer/assets/theme.css`, the skillsbench.ai tokens), so `ir_to_view_html` pages pick up the new look without changes.
* The bigger integration question: in [#1034](https://github.com/benchflow-ai/benchflow/pull/1034) the ACP branch of `render_rollout` no longer emits the static card page — it renders an interactive page (tabs, in-trace search, timeline, browse catalog) from a typed payload contract (`viewer/models.py`: `Step` / `ViewerPayload` / `RunSummary`). Your `if not turn_files:` branch sits in front of that, so with `BENCHFLOW_VIEWER_TRACE_IR=1` an ACP rollout would drop back to a static page. Suggest pointing `IR → view` at the payload contract instead of the HTML primitives — then ATIF/OTel sources get the interactive page for free, and the step list you build in slice H maps onto `Step` pretty directly. Can work through the field mapping whenever.

No preference on merge order.

Galius5136:
Conflicts resolved, ready for the merge @bingran-you

PR #986, opened 2026-08-14
fix(openclaw): clamp GPT-5.4 output tokens

kywch:
## Summary

Clamp OpenClaw GPT-5.4 output tokens to the provider limit.

## Root cause

OpenClaw sent max_tokens=171802 while GPT-5.4 accepts at most 128000. The ACP path uses internal alias benchflow-openai-gpt-5.4, so raw gpt-5.4 matching alone misses live requests.

## Changes

- Cap raw and ACP-alias GPT-5.4 IDs at 128000.
- Replace missing, malformed, or oversized configured values.
- Preserve explicit numeric values at or below the limit.

## Scope

OpenClaw + GPT-5.4 only. No global model cap, routing change, or OpenClaw runtime-pin change. #977 is adjacent only.

## Validation

- 47 focused tests passed.
- git diff --check passed.

### DIFF
--- a/src/benchflow/agents/openclaw_acp_shim.py
+++ b/src/benchflow/agents/openclaw_acp_shim.py
@@ -47,15 +47,17 @@ _PARAM_MAP = {
     "BENCHFLOW_MODEL_TOP_P": "agents.defaults.params.topP",
     "BENCHFLOW_MODEL_MAX_TOKENS": "agents.defaults.params.maxTokens",
 }
-_MODEL_MAX_TOKENS = {
-    # OpenClaw derives an invalid ~172k default for GPT-5.4; ACP may send either ID.
-    "gpt-5.4": 128000,
-    "benchflow-openai-gpt-5.4": 128000,
-}
 
 
 def _default_max_tokens(model: str) -> int | None:
-    return _MODEL_MAX_TOKENS.get(model.removeprefix("openai/"))
+    # OpenClaw derives an invalid ~172k default for GPT-5.4; ACP sends a
+    # BenchFlow-generated LiteLLM alias in normal runs.
+    model = model.removeprefix("openai/")
+    if model == "gpt-5.4" or (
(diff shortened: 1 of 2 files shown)


Galius5136:
I think there is one proxy-alias path still uncovered here. I checked this on the current PR head. These two spellings resolve to the same upstream model (`openai/gpt-5.4`), but produce different proxy aliases: `gpt-5.4` -> `benchflow-gpt-5.4`, while `openai/gpt-5.4` -> `benchflow-openai-gpt-5.4`. 

With the current `_MODEL_MAX_TOKENS` table, `_default_max_tokens(safe_model_alias("gpt-5.4"))` returns `None`, while `_default_max_tokens(safe_model_alias("openai/gpt-5.4"))` returns `128000`. So if `session/set_model` receives the proxy alias derived from a bare `--model gpt-5.4`, the clamp is skipped even though both spellings resolve to the same `openai/gpt-5.4` upstream model.

I also couldn't find a test exercising the new `_default_max_tokens` logic. A small regression test that derives the alias through `safe_model_alias()` for both spellings would pin this without hardcoding the alias-generation behavior.

kywch:
@Galius5136 thanks for looking into it! I made it more general and added a small test.

Galius5136:
@kywch confirmed, both spellings now hit the cap:

`_default_max_tokens(safe_model_alias("gpt-5.4")) -> 128000`

`_default_max_tokens(safe_model_alias("openai/gpt-5.4")) -> 128000`

The new parametrized test also fails on `3210f5c4` for the bare `gpt-5.4` case and passes on `def35f88`, so it does pin the regression rather than just restating the fix.

Thanks for the quick turnaround!

bingran-you:
This focused sync is now the only blocker to manifest-parity on #989: agents@ce26cbc has the updated OpenClaw manifest, while current benchflow main predates this PR. #986 is fully green. An independent maintainer review/merge here will let #989 refresh from main without absorbing unrelated OpenClaw code.

PR #987, opened 2026-08-14
feat(viewer): render run traces from a payload with a JS renderer

Polarisyjr:
## Summary

This is **Step 1 of the BenchFlow trace viewer work**. More features will be added in follow-up PRs; at this stage, the main goal is to establish the rendering architecture and bring the viewer's overall style and trace presentation closer to PostTrainBench.

Refactors `bench eval view` around a shared payload-based renderer:

`rollout artifacts → viewer.py → versioned payload → render.js → HTML`

Instead of generating trace HTML directly in Python, Python now handles artifact normalization while JavaScript handles presentation. This also creates a shared renderer that can later be reused by a hosted/static viewer.

## What changed

- Add a versioned viewer payload generated from rollout artifacts.
- Treat `trajectory/acp_trajectory.jsonl` as the canonical trace source, with legacy `turn*.txt` as a fallback.
- Move trace HTML rendering into `render.js`.
- Separate browser/DOM setup into `boot.js`, keeping the renderer independently testable.
- Rework the viewer layout and styling to roughly align with the PostTrainBench trace viewer.
- Vendor the PostTrainBench stylesheet as the base visual layer, with BenchFlow-specific overrides isolated in `benchflow.css`.
- Add BenchFlow-specific run information including:
  - reward and run status
  - token usage
  - environment / agent / verifier timing
  - skill mode and skill invocations
  - agent/verifier errors and partial trajectories
- Improve tool-call rendering:
  - preserve complete ACP content blocks
  - render diffs separately
  - handle binary output per block
  - truncate very large outputs while preserving the beginning and end
  - support ANSI formatting and a limited escaped Markdown subset
- Bundle viewer assets locally so generated pages remain portable and work without network access.

## Compatibility

This does **not** introduce a new on-disk trace format.

The viewer payload is derived at render time and is not persisted. Existing ACP trajectory artifacts remain unchanged, and legacy Claude Code `turn*.txt` traces continue to work as a fallback.

When both formats exist, ACP is used as the source of truth rather than merging or rendering both.

## Tests

Added two layers of viewer tests:

- `tests/test_trajectory_viewer.py`
  - payload structure
  - run status
  - token metadata
  - turn grouping
  - tool-output completeness and truncation
  - payload/script escaping

- `tests/test_trajectory_viewer_render.py`
  - runs `render.js` through Node
  - tool output and diff rendering
  - Markdown and ANSI rendering
  - binary placeholders
  - injection protection
  - reward `0` vs. unscored
  - schema-version mismatch

Also updates the trajectory integration tests for the new payload-based rendering path.

## Validation

```bash
uv run pytest \
  tests/test_trajectory_viewer.py \
  tests/test_trajectory_viewer_render.py \
  tests/test_trajectory_integration.py -q
```

Manual E2E:

```bash
bench eval view <rollout-dir>
```

## Next steps

This PR intentionally focuses on the basic single-run viewer and visual alignment with PostTrainBench.

Follow-up work will add more viewer features such as job-level navigation, verifier details, review/audit results, cost and per-call metrics, system metrics, and additional trace metadata.

image
image


1digitaldesign:
Cool cool!

PR #993, opened 2026-08-15
test(cli): normalize Rich option styling

bingran-you:
## Summary
- strip Rich ANSI styling before asserting required contributor option names
- keep the regression focused on the single `bench traj upload` command

## Why
Exact-main CI on `7eeeedd9` rendered Rich help/error output with ANSI boundaries inside option names, causing two presentation-only assertions to fail while CLI behavior remained correct.

## Validation
- `FORCE_COLOR=1 uv run python -m pytest tests/test_traj_upload_cli.py -q` (17 passed)
- `uv run ruff format --check tests/test_traj_upload_cli.py`
- `uv run ruff check tests/test_traj_upload_cli.py`
- `uv run ty check src services`

### DIFF
--- a/tests/test_traj_upload_cli.py
+++ b/tests/test_traj_upload_cli.py
@@ -7,6 +7,7 @@ import logging
 from pathlib import Path
 from types import SimpleNamespace
 
+import click
 import httpx
 import pytest
 from typer.testing import CliRunner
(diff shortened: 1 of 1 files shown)


bingran-you:
@xdotli All checks are green, including the forced-color regression, full unit suite, parity, live rollout, and fixture scenarios. This test-only fix unblocks the exact-main deployment and 0.6.9 release; please review when available.

Issue #982, opened 2026-08-13
Claude ACP trajectories capture no thinking text on models that default `thinking.display` to `"omitted"`


Polarisyjr:
## Problem

When BenchFlow runs `claude-agent-acp` with some Claude models, the trajectory contains no usable thinking content.

The agent still reasons normally, but `agent_thought` events are either missing or contain empty text, so the reasoning process cannot be reviewed from the trajectory.

## Root Cause

Call chain:

```text
BenchFlow → claude-agent-acp → Claude Agent SDK → Claude
```

Some newer Claude models default to:

```text
thinking.display = "omitted"
```

Under this setting, the model still performs reasoning, but the thinking summary is not returned.

BenchFlow currently sends no thinking configuration in `session/new`, so it inherits the model default.

This also explains why the issue is model-dependent: some older models default to summarized thinking, so the problem does not appear when testing with those models.

## Validation Results

Using the same task, with the only difference being whether `thinking.display="summarized"` is explicitly set:

| Model | Configuration | Non-empty thought events | Thinking chars |
|---|---|---:|---:|
| `claude-sonnet-5` | default | 0 | 0 |
| `claude-sonnet-5` | `summarized` | ≥2 | ≥1585 |
| Haiku 4.5 | default | 9 | 3532 |
| Haiku 4.5 | `summarized` | 9 | 3559 |

Sonnet 5 reliably reproduces the issue, and the thinking content is restored once `thinking.display="summarized"` is explicitly enabled. Haiku 4.5 behaves almost identically in both cases, so testing only on Haiku 4.5 would not reveal the issue.

PR #886, opened 2026-07-03
Registry-driven Claude subscription gate + zero-signal heuristic exemption (Anthropic OAuth unlock)

Yiminnn:
## What

Two changes that make Anthropic OAuth/subscription runs (`CLAUDE_CODE_OAUTH_TOKEN`, no `ANTHROPIC_API_KEY`) work beyond the single hardcoded agent:

1. **Registry-driven Claude subscription gate** (`agents/env.py`): `uses_native_subscription_auth` hardcoded `agent == "claude-agent-acp"` for the Anthropic split. Decoupled Claude-CLI agents (e.g. omnigent claude-sdk) could never opt in — with an OAuth token and no API key, the mandatory LiteLLM route failed with `requires ANTHROPIC_API_KEY` even though the Claude CLI authenticates itself. Now any agent whose `subscription_auth` declares `replaces_env == "ANTHROPIC_API_KEY"` takes the same Harbor-style split. `claude-agent-acp` keeps byte-identical behavior (its registry entry declares exactly that); agents without `subscription_auth` still never skip the proxy (regression-tested).

2. **Zero-signal heuristic exemption for subscription runs** (`rollout/__init__.py`): subscription runs skip the proxy by design, so they have **no usage channel** — zero tokens + zero tool calls is the expected shape of a *healthy* run for agents whose trajectory carries no tool telemetry (omnigent's flat session events). The `suspected_api_error` layer was nulling verifier-granted rewards there. Live repro: omnigent-claude + OAuth scored **1.0**, then got overridden to `reward=None / suspected_api_error`. The zero-signal layer is now skipped when `uses_native_subscription_auth` holds; the proxy-proven layer needs proxy evidence that cannot exist there, and real failures still surface via the agent error channels.

## Live verification (daytona, citation-check, with-skill, claude-sonnet-4-6)

- `claude-agent-acp` + `CLAUDE_CODE_OAUTH_TOKEN` → **reward 1.0** (acp 33 events, 17 tool calls)
- `omnigent-claude` + OAuth (with benchflow-ai/agents companion PR) → **reward 1.0, Score 1/1, errors=0**

## Tests

- `test_subscription_auth.py`: new `test_anthropic_subscription_gate_is_registry_driven` (opt-in agent honored; API-key still prefers LiteLLM; agents without subscription_auth never skip) — 31 pass
- `test_api_error_capture.py`: new `TestSubscriptionAuthExemption` (OAuth run keeps verifier reward; proxied run still flagged) — 30 pass
[...]

Issue #389, opened 2026-05-24
Self-gen skill export failures are swallowed, allowing empty skill updates to look successful


xdotli:
## Summary

During cleanup, `Rollout.cleanup()` catches every exception from `_export_generated_skills()` and logs only a warning. Continual-learning code then falls back to reading the export directory; if it is missing, `capture_skills()` returns `{}`. That can make a self-gen/continual-learning rollout look successful while silently committing or scoring an empty evolved skill set.

This is a measurement-integrity bug for capability 5 / memory-space reward: export failure is infrastructure failure, not "the agent evolved no skills".

## Reproduction

```bash
uv run python - <<'PY'
import asyncio
from pathlib import Path
from unittest.mock import AsyncMock
from benchflow.rollout import Rollout, RolloutConfig

async def main():
    r = Rollout(RolloutConfig(
        task_path=Path('tests/examples/hello-world-task'),
        export_generated_skills_to='/tmp/does-not-matter',
    ))
    r._env = object()
    r.disconnect = AsyncMock()
    r._capture_partial_acp_trajectory = lambda: None
    r._export_generated_skills = AsyncMock(side_effect=RuntimeError('download failed'))
    r._stop_provider_runtime = AsyncMock()
    r._cleanup_provider_runtime = AsyncMock()
    await r.cleanup()
    print('cleanup_returned_after_export_failure', True)

asyncio.run(main())
PY
```

Observed:

```text
Generated skill export failed: download failed
cleanup_returned_after_export_failure True
```

`capture_skills(missing_export_dir)` then returns `{}` by design.

## Expected behavior

If `export_generated_skills_to` is configured and export fails, the rollout/job should surface an error state or a typed export failure. It should not be indistinguishable from a valid "no evolved skills" outcome.

## Actual behavior

Relevant code:

- `src/benchflow/rollout.py`: `cleanup()` catches `_export_generated_skills()` exceptions and logs a warning.
- `src/benchflow/learner_skills.py`: `capture_skills()` returns `{}` for missing directories.
- `src/benchflow/_utils/learner_memory.py`: `evolved_skills_for_result()` falls back to `capture_skills(export_dir)`.

## Impact

P1 eval/train integrity for self-gen and continual learning. A transport/download/sandbox failure can erase the generated skills from the memory-space signal while the run proceeds as if it honestly produced no skills. That poisons learning curves and any release evidence about self-generated skill handoff.

## Suggested fix

[...]

Issue #501, opened 2026-05-25
summary.json omits aggregate tool-call and phase-timing metrics


bingran-you:
## Summary

Job-level `summary.json` does not aggregate useful per-rollout metrics that already exist in `result.json`, including `total_tool_calls`, agent time, and verifier time.

This is BF-E2E-15 from the PR #381 follow-up audit.

## Source

- Follow-up audit: https://github.com/benchflow-ai/benchflow/pull/381#issuecomment-4530737015
- Raw artifacts from manual e2e: https://gist.github.com/bingran-you/0216957ba76491f3c87f2d14879476d7

## Current behavior

`summary.json` includes high-level counts, elapsed time, memory score, usage totals, and source fields, but does not include aggregate tool calls or phase timing sums/averages.

Per-rollout `result.json` has data such as:

- `n_tool_calls`
- `timing.environment_setup`
- `timing.agent_setup`
- `timing.verifier`
- `timing.total`

## Impact

Reviewers have to inspect individual result files or dashboard-derived rows to answer basic run-level questions:

- How many tool calls did the whole job consume?
- How much time was spent in agent vs verifier phases?
- Which phase dominated runtime?

This was especially noticeable during the PR #381 e2e run because the ENG-152 scenario took about 97 minutes.

## Suggested fix

After collecting `all_results`, aggregate at least:

- `total_tool_calls`
- `avg_tool_calls_per_task`
- `agent_time_sec` / `verifier_time_sec` / `environment_setup_time_sec` sums
- optional averages and maxes per phase

Keep existing summary fields for backward compatibility.

bingran-you:
Reproduced on v0.5-integration, labeled reproduced, and addressed in PR #513: https://github.com/benchflow-ai/benchflow/pull/513

Outcome: summary.json now aggregates total_tool_calls, avg_tool_calls_per_task, environment_setup_time_sec, agent_setup_time_sec, agent_execution_time_sec, agent_time_sec, verifier_time_sec, and rollout_time_sec from per-rollout results.

Validation: focused summary aggregation regression, touched job suite, full unit suite (2353 passed, 11 skipped, 1 deselected), ruff, and ty.

xdotli:
Fixed in #512 — merged to v0.5-integration.

bingran-you:
Reviewer subagent reviewed PR #513 — **APPROVE** ✅.

**Per-issue verification (all PASS):**

[...]

PR #1025, opened 2026-08-16
Flag chat-only completions (zero tool calls with model output) as a diagnostic

### LINKED ISSUE #988: Zero-tool-call rollouts are recorded as clean completions, silently biasing baseline sweeps
### Summary

A rollout in which the agent makes **zero tool calls** and never touches the workspace is
recorded as a clean, completed rollout: `error: null`, `error_category: null`,
`partial_trajectory: false`, no idle-timeout info, reward `0.0`. In an eval sweep it is
indistinguishable in the aggregate from a genuine attempt that tried and failed, so it
silently biases baseline statistics downward.

### Observed

`opencode-ai@1.18.11`, `--agent opencode --model openai/gpt-5 --sandbox daytona`, one task
out of a six-rollout baseline sweep:

| | value |
|---|---|
| `n_tool_calls` | **0** |
| trajectory | 1 `user_message`, 1 `agent_message`, nothing else |
| `finish_reason` | `stop` |
| `agent_execution` | 30.0 s (siblings: 210–1294 s) |
| output tokens | 2 727 |
| `error` / `error_category` | `null` / `null` |
| `partial_trajectory` | `false` |
| reward | `0.0` |

The entire agent output was two paragraphs of narrated intent, ending:

> First, I'll write a literature review and an ordered calibration plan to PLAN.md. Then
> I'll inspect /root/data/metadata.json … **Proceeding to add PLAN.md and read metadata and
> list data files.**
>
> I will create the literature review and calibration plan in PLAN.md, then inspect
> /root/data/metadata.json and enumerate the available sweeps …

It then returned. No `apply_patch`, no `read`, no `bash`.

The five sibling rollouts on the same task, same agent, same model and same sandbox each
made 9–18 tool calls and ran 3.5–9 minutes, so this is not the task being unreachable or
the sandbox failing to come up.

### Why this is worth surfacing rather than filing under "the model did that"

The agent stopping of its own accord is the model's business. What is the harness's
business is that the result is **unlabelled**: nothing in `result.json` distinguishes
"attempted the task and scored 0" from "never issued an instruction". Anyone computing a
pass rate, a mean reward, or a with-skill/without-skill delta over a sweep is silently
mixing the two populations, and the effect is largest exactly where sample sizes are
smallest.

A cheap fix would be a flag on the result — e.g. `no_tool_calls: true`, or folding it into
the existing `error_category` taxonomy as a distinct non-error outcome — so sweeps can
report it separately instead of every consumer having to re-derive it from
`n_tool_calls == 0`.

### Related

Same ACP adapter family as #945, though a different failure mode: #945 is about model and
reasoning-effort capability mapping at `session/set_model`, whereas this rollout negotiated
its session fine and simply produced no tool calls. Filing separately rather than as a
comment there so the two do not get conflated.

ljr145733:
Fixes #988 — implements the design proposed in my comment there. Opening the PR now so the semantics discussion can happen on concrete code rather than in the abstract; happy to rework any of it if maintainers prefer a different call.

## Problem

A rollout where the agent produced real model output but ended its turn without a single tool call was recorded as a clean scored fail — `error: null`, reward 0.0 from the verifier — indistinguishable in aggregates from a genuine attempt. The existing zero-signal net (`suspected_api_error`) only covers the zero-token half of the no-op space; the reported rollout had 2,727 output tokens ("Proceeding to add PLAN.md…") and fell through.

## Semantics (the key design decision)

A chat-only completion is **agent/model behavior, not infrastructure failure** — so unlike `suspected_api_error` this diagnostic **leaves reward and error untouched**: the slot stays a scored fail and stays in score denominators. It only becomes visible. This matches the intent already documented in `test_zero_tools_with_tokens_not_flagged` ("Prompt-only answer (no tools) with real usage is a legitimate rollout") — legitimate for the *error* classifier, but worth surfacing for sweep forensics.

## What this adds

- `NoToolCallCompletionDiagnostic` (`diagnostics.py`), field `no_tool_call_completion_info`, `category = None` — registered, so it lands in result.json via the existing `to_result_fields()` spread.
- Detection in `_maybe_flag_no_tool_completion()` (`rollout/__init__.py`), called right after `_maybe_classify_api_error()`. Guards reuse the established ones: error channel already claimed → skip (zero-token cases still route to `suspected_api_error`); no executed prompts → skip (#389); oracle → skip; native-subscription telemetry exemption from #886 → skip; and the trajectory must contain **at least one `agent_message` event**, so trajectory-capture loss (#982) is never misflagged as a chat-only completion.
- Surfacing: per-rollout CLI line gains a `, no-op` marker (`[FAIL] task-x (reward=0.00, tools=0, no-op)`); `summary.json` gains a `no_tool_call_completions` count; the registry-driven job-summary warning loop now counts **category-less** diagnostics by field presence in result.json (no behavior change for the seven existing diagnostics, which all carry categories).
- `RolloutResult` gains a trailing optional `no_tool_completion: bool = False` so the CLI reads the authoritative flag instead of re-deriving it.

## Open question from the issue thread

The tool-telemetry guard currently reuses `uses_native_subscription_auth()` (option (a) from my comment) — symmetric with `_maybe_classify_api_error`, but over-broad for ACP adapters that do report `tool_call` events under subscription auth. If you'd rather gate on a per-adapter tool-telemetry capability flag (option (b)), I'm happy to do that here or as a follow-up — the detection site has a TODO marking the swap point.

## Testing

- New `tests/test_no_tool_completion.py` (13 tests): core semantics (flag set, reward/error untouched), one negative test per guard, precedence (zero-token case still routes to `suspected_api_error`), registry integrity, `format_issue` rendering.
- Adjacent suites pass unchanged: `test_api_error_capture.py` (30), `test_diagnostics.py` + `test_eval_single_task_summary.py` (17), `test_rollout_architecture.py` + `test_integration_check_results.py` + `test_rollout_import_no_side_effects.py` (69). `ruff check` clean on all touched files.

---

*Disclosure: Claude assisted with the implementation and tests under my direction; I reviewed and take responsibility for the design, the semantics, and all code.*

### DIFF
--- a/src/benchflow/diagnostics.py
+++ b/src/benchflow/diagnostics.py
@@ -334,6 +334,36 @@ class SuspectedApiErrorDiagnostic(Diagnostic):
         )
 
 
+@dataclass
+class NoToolCallCompletionDiagnostic(Diagnostic):
+    """Chat-only completion: the agent produced model output (tokens and at
+    least one agent message) but ended its turn without issuing a single tool
+    call (#988). Unlike the api-error verdicts this is agent/model BEHAVIOR,
+    not an infrastructure failure — the reward and error channels are left
+    untouched (``category`` stays None), so the slot remains a scored fail.
+    The diagnostic only makes the no-op visible in result.json, the CLI line,
+    and the job summary so baseline sweeps don't read these rollouts as
+    genuine attempts."""
+
+    total_tokens: int = 0
+    n_output_tokens: int = 0
+    n_agent_messages: int = 0
+    n_message_chars: int = 0
+
+    field: ClassVar[str] = "no_tool_call_completion_info"
+    category: ClassVar[str | None] = None
+    summary_description: ClassVar[str] = (
+        "completed without a single tool call (chat-only)"
+    )
+
+    def format_issue(self, task_name: str) -> str:
+        return (
+            f"{task_name}: chat-only completion — agent produced "
+            f"{self.n_output_tokens or self.total_tokens} tokens across "
+            f"{self.n_agent_messages} message(s) but made 0 tool calls"
+        )
+
+
 # Public registry — every diagnostic kind goes here exactly once.
 DIAGNOSTIC_REGISTRY: tuple[type[Diagnostic], ...] = (
     IdleTimeoutDiagnostic,
@@ -343,6 +373,7 @@ DIAGNOSTIC_REGISTRY: tuple[type[Diagnostic], ...] = (
     VerifierTimeoutDiagnostic,
     ProviderApiErrorDiagnostic,
     SuspectedApiErrorDiagnostic,
+    NoToolCallCompletionDiagnostic,
 )
 
 # field_name → Diagnostic class, for check_results lookup.
--- a/src/benchflow/evaluation.py
+++ b/src/benchflow/evaluation.py
@@ -1403,8 +1403,11 @@ class Evaluation:
             if isinstance(reward, (int, float)) and not isinstance(reward, bool)
             else ""
         )
+        # Chat-only completions (#988): the agent answered but never called a
+        # tool — mark the line so a sweep log shows the no-op at a glance.
+        no_op = ", no-op" if getattr(result, "no_tool_completion", False) else ""
         logger.info(
-            f"[{status}] {td.name} ({reward_part}tools={result.n_tool_calls}){err}"
+            f"[{status}] {td.name} ({reward_part}tools={result.n_tool_calls}{no_op}){err}"
         )
         self._fire_progress(self._on_result, td.name, result)
 
@@ -1890,6 +1893,11 @@ class Evaluation:
             "error": audit_counts["errored"],
             "verifier_errored": audit_counts["verifier_errored"],
             "idle_timeout": error_category_counts.get(IDLE_TIMEOUT, 0),
+            "no_tool_call_completions": sum(
+                1
+                for r in all_results.values()
+                if r.get("no_tool_call_completion_info")
+            ),
             "error_categories": error_category_counts or None,
             "verifier_error_categories": verifier_error_category_counts or None,
             "score": f"{pass_rate(passed=audit_counts['passed'], total=job_result.total):.1%}",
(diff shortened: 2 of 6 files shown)


Galius5136:
Thanks for putting this together! I tested the PR end to end and the per-rollout diagnostic looks good on the #988 shape: the flag is set, while reward/error stay unchanged.

I did hit one issue at the aggregate level though. On a fresh `bench eval` run, the chat-only rollout had `no_tool_call_completion_info` populated, but `summary.json` still reported `no_tool_call_completions: 0` and no job-summary warning was shown. Re-running the same job from disk, with the same artifacts and 0 tasks re-executed, changed the count to `1` and emitted the warning. It looks like fresh results go through `rollout_result_payload()`, which currently doesn't carry `no_tool_call_completion_info`, while resumed results are read back from `result.json`.

I also noticed a narrower edge case in the Gemini scraped-trajectory fallback: the saved trajectory can contain tool calls while `_n_tool_calls` remains 0, so the new diagnostic can say “made 0 tool calls” next to `trajectory_summary.tool_call_steps: 3`. I only reproduced this by forcing the salvage precondition in a probe, so I can't say how common it is in real runs.

Small cleanup: `ruff format --check src tests tools` currently fails on the PR head (base passes) for the new formatting in `evaluation.py` and `rollout/__init__.py`.

Hope this helps, happy to re-test if you push a fix.

ljr145733:
Thanks for the thorough end-to-end test — all three findings were real. I really appreciate it. They should be fixed in 01bb6bcb:

1. **Fresh-run aggregation**: `rollout_result_payload()` is `RolloutResult`-driven and never carried the diagnostic payloads, so `summary.json` only counted them after a resume re-read result.json from disk. Rather than special-casing the new field, I generalized the existing #501 persisted-timing enrichment into `_enrich_payload_with_persisted_fields()`, which now copies every `DIAGNOSTIC_REGISTRY` field from the persisted result.json into fresh-run payloads — fresh and resumed jobs now aggregate identically, for this diagnostic and any future one. Covered by three new tests (fields copied, nulls skipped, in-memory values never overwritten).

2. **Scraped-trajectory fallback**: confirmed — the salvage path rebuilds `tool_call` events the ACP session never counted, so `_n_tool_calls` can be 0 next to `trajectory_summary.tool_call_steps > 0`. The detection now treats the trajectory as authoritative: any `tool_call` event in it disqualifies the chat-only flag. Negative test added with exactly that shape.

3. **Formatting**: reproduced and fixed; `ruff format --check src tests tools` is clean on the new head.

Would appreciate a re-test if you have the time — the fresh-run repro you described should now show `no_tool_call_completions: 1` and the job-summary warning on the first pass.

Galius5136:
Re-tested on 01bb6bcb and all three look fixed now. Thanks for the quick turnaround!

Fresh-run aggregation now matches resume on the same artifacts, the scraped-trajectory case no longer gets the chat-only diagnostic, and ruff format --check src tests tools is clean.

One small test-coverage note: I tried removing the new enrichment call site and the current test suite still stays green, so it may be worth adding an Evaluation-level regression test for that wiring. Nothing blocking from my side though, happy with the fixes :)

ljr145733:
Good catch on the coverage gap — you're right that the enrichment tests exercised the helper in isolation, so the call-site wiring itself was unguarded. Added in c378761e: an `Evaluation.run()`-level regression test where a mocked rollout persists `no_tool_call_completion_info` only to its on-disk result.json (the exact fresh-run shape) and the test asserts `summary.json` counts it on the first pass. Verified it goes red with the call site removed and green with it present. Thanks again for the careful testing!

Galius5136:
Confirmed the new test goes red without the call site. Thanks, looks good to me!

Issue #1033, opened 2026-08-17
Subscription-auth runs (agent_native_acp) have no event-level timing — serialize timestamps in acp_trajectory.jsonl


ljr145733:
## Problem

For runs using subscription auth (`usage_source: agent_native_acp`), there is
**no per-event timing anywhere in the rollout artifacts**:

- `trajectory/acp_trajectory.jsonl` events carry no timestamp fields — for
  any run type. `_events_to_trajectory`
  (`src/benchflow/trajectories/_capture.py`) drops the
  `started_at`/`finished_at` that `ToolCallRecord`
  (`src/benchflow/acp/session.py`) already tracks in memory, and text events
  (`user_message` / `agent_message` / `agent_thought`) are never stamped.
- Proxy-tracked runs (`usage_source: provider_response`) at least have
  `llm_trajectory.jsonl`, whose records carry `logged_at` + `duration_ms`
  (`src/benchflow/providers/litellm_logging.py`), so coarse timing can be
  reconstructed. But subscription runs never go through the LiteLLM proxy, so
  `llm_trajectory.jsonl` is empty/absent for them — and they are left with
  only run-level `started_at`/`finished_at` and the phase totals in
  `timing.json`.

Concrete impact: in a recent FrontierPhysics experiment batch, 25/25 rollouts
ran via `@agentclientprotocol/claude-agent-acp` with subscription auth. One
failed run spent 85m 44s over 187 events — with no way to tell which tool
call or reasoning stretch consumed the time. Timeout forensics
(`agent_timeout` lists pending tool calls, but not how long they had been
pending), hang diagnosis, and per-tool latency analysis are all impossible
exactly on the run type the community uses most.

## Proposal

Serialize the timing the session already has, as **additive optional
fields** in `acp_trajectory.jsonl` events:

- `tool_call`: `started_at`, `finished_at` (already tracked in memory on
  `ToolCallRecord`; today discarded at serialization)
- text events and `agent_timeout`: a `ts` stamped when the event is appended

ISO-8601 UTC, consistent with `llm_trajectory`'s `logged_at`. Emit from the
single canonical converter (`_events_to_trajectory`) so streaming snapshots
and the final write stay format-identical.

## Compatibility

Purely additive — no event types or existing fields change. Existing
consumers (viewer, ATIF/ADP/SFT exporters, upload validation) tolerate
unknown fields today. This benefits all run types uniformly; subscription
runs simply go from "no timing at all" to full parity with proxy runs.

[...]

PR #1034, opened 2026-08-17
feat(viewer): reviewer-grade trajectory pages, multi-run browsing, and hf:// dataset sources

ljr145733:
## Summary

Reviewers auditing trajectories (health checks, failure attribution, reward
hacking, no-skill leakage — the weekly-sync "[8pt - rollout] viewer" ticket)
need more than the current ACP page gives them: today it truncates every
message at 500 chars, shows tool calls without their output, and has no
stats, filters, or verifier view. This PR upgrades `bench eval view`'s ACP
path into an interactive review page, adds a multi-run browser, and makes
HuggingFace trajectory datasets directly viewable — while keeping every
existing behavior (`--confirm` contract, session JSONL files, legacy
`turn*.txt` renderer) byte-for-byte intact.

Built on and styled after the 0.7.x viewer stack (#1019 site restyle, #1020
accents, #1021 `--confirm`, #1022 redaction summary, #1023 badge rules):
same light monochrome language, same `_VIEWER_CSS` token names (the confirm
bar injects with zero adaptation), same pastel accent palette.

## Usage

**1. Local folders.** A single rollout renders the interactive page; a
directory *of* rollouts (a job dir, a whole `jobs/` tree) serves a browser
with a run sidebar — task, pass/fail, harness, model, skill mode — traces
loaded dynamically without restarting, and `?run=<id>` deep links:

```bash
bench eval view jobs/run/task__abc123            # one rollout
bench eval view jobs/ --port 9000                # browse every run under jobs/
```

**2. HuggingFace datasets.** The community ground-truth uploads become
reviewable with one command. Only the viewer-relevant files are fetched
(trajectories + result/timing/prompts/verifier sidecars — never the large
`llm_trajectory`/`trainer` exports) into the shared `huggingface_hub`
cache, so repeat views are incremental:

```bash
bench eval view hf://benchflow/skillsbench-trajectories-apr2026/jobs/opus47-with-skills-t1
```

Validated live: 225 runs from that dataset (and 25 local FrontierPhysics
rollouts, including old-schema/partial/timeout cases) browse and render end
to end.

## What the page gives reviewers

- **Full fidelity**: no truncation anywhere — long prompts/thoughts/outputs
  collapse with an expand control; ACP diff content blocks render as
  readable pseudo-diffs; `agent_timeout` events render as cards (previously
  silently dropped); unknown/future event types degrade to generic cards.
- **Failure attribution at a glance**: header surfaces `error_category`,
  `verifier_error`, and the seven `*_info` diagnostic blocks from
  result.json as banners, plus harness/model/skills identity tiles and
  token/cost/duration stats.
- **Reading tools**: Focus/Full modes, per-kind filters (one pastel hue per
  tool kind, dark terminal treatment reserved for `execute` output),
  failed-only, text search with jump, per-event `#e42` anchors for citing
  steps in review threads; chunked rendering stays responsive on large
  traces.
- **Verifier + Metrics tabs**: reward, test stdout/stderr, CTRF table,
  phase-timing bars, token breakdown.
- **Timeline, forward-compatibly**: steps show `+m:ss` offsets and tool
  durations whenever events carry `ts`/`started_at`/`finished_at` (#1033);
  a scan of 140 non-empty uploaded trajectories (5,698 events) found zero
  timestamp fields today, so nothing changes visually until capture writes
  them.

## Compatibility

- `serve()` becomes a thin dispatcher; single trajectories go through
  `_serve_single`, which is byte-identical to the previous `serve()` body —
  confirm/`DECISION:` contract, sidecar write-before-inject ordering, exit
  semantics, JSONL session files all preserved. `--confirm` on a multi-run
  directory errors out (a confirmation needs exactly one trajectory).
- Raw session files keep the existing `_render_acp_events` renderer whose
  output the jsonl-session tests pin; the legacy `turn*.txt` renderer is
  untouched apart from `errors="replace"` unicode hardening on reads.
- All pre-existing viewer tests pass unmodified.

## Security

Trajectory content is untrusted input end to end: it travels as JSON data
(`</` escaped against script-tag breakout) and the template renders
exclusively via `textContent`; browse-mode ids resolve only by exact
membership in a fresh directory scan, so crafted ids (`../` traversal)
cannot reach the filesystem. An adversarial review pass produced five
verified findings (lone-surrogate encode crash, renderer-halting type
confusion, non-UTF-8 sidecar crash, null-timing crash, diff-block
rendering) — all fixed with regression tests.

## Testing

- `tests/trajectories/test_viewer_acp_renderer.py` (new, 19 tests): payload
  contract, escaping/hostile-input cases, degradation, browse discovery +
  anti-traversal, hf:// spec parsing with a mocked download (asserting the
  large files are never requested), timestamp passthrough.
- Full suite green locally (`pytest tests/`: 5685 passed), `ruff check`,
  `ruff format --check`, `ty check` clean.
- Live smoke: confirm flow E2E on a template page (bar + redaction note,
  POST `/decision`, single `DECISION:` line, clean sidecar), browse API +
  traversal guard, HF dataset end to end.

## Screenshots

(attached below: single-run page, browse mode over a real experiment batch,
HF dataset browsing, `--confirm` bar on the interactive page)

Related: #1033 (capture-side timestamps — the timeline above renders the
moment it lands).
image
image
image
image


Galius5136:
Really enjoyed this one! I tested the main paths pretty hard and a lot of it held up well: traversal attempts all 404'd, the single-run path stayed compatible with main, the confirm flow worked end to end, and the `hf://` path worked on the real 225-run dataset without pulling the large trajectory/trainer exports.

A few things are worth fixing before merge. `bench eval view --help` currently exits 1 because Rich interprets `[/subpath]` in the help text as a closing markup tag; it's also the only head-only failure I found in the full test lane. `ruff format --check src tests tools` also currently fails on the new viewer test file. The other behavioral issue I reproduced is the 500-run cap: with 620 local rollouts, opening `?run=...0619` leaves that id in the address bar but silently renders run 1 instead, and the requested run is 404 through the API. The advertised HF dataset is below the cap, so this is specifically a larger local jobs-tree case.

One test-coverage thing I’d tighten too: both security mitigations work today, but their regression tests don’t actually pin the enforcement. Removing the `</` escaping still leaves all 19 tests green while trajectory text becomes live script markup, and removing the browse membership check also stays green while an outside rollout becomes reachable. Small integration note as well: `_DIAGNOSTIC_KEYS` currently mirrors `DIAGNOSTIC_REGISTRY`, but #1025 adds another diagnostic, so deriving that list from the registry would avoid future drift (with a neutral style for category-less behavior flags).

Happy to re-test after any updates 🙂

ljr145733:
Thanks for the thorough pass — reproduced every finding, all five addressed in d26931b7:

- **`--help`**: Rich was parsing `[/subpath]` in the argument help as a closing markup tag; rephrased without brackets. Exits 0 now.
- **`ruff format --check src tests tools`** passes (the new test file was the offender).
- **Run cap**: the sidebar heading now says "first N runs (capped)" when truncation actually happened (detected by scanning cap+1 rather than `len == cap`), a `?run=` pointing at an undiscovered id shows an explicit load error instead of silently rendering run 1, and `BENCHFLOW_VIEWER_MAX_RUNS` overrides the 500 default.
- **Security tests now pin enforcement**: the breakout test asserts the raw `</script><script>` sequence appears nowhere in the emitted page (fails with the escape removed), and id resolution moved into a `_resolve_browse_rollout` helper with a test that places a real rollout *outside* the served base and asserts it never resolves (fails with the membership check removed). Good catch — the old assertions were theater.
- **`_DIAGNOSTIC_KEYS` derives from `DIAGNOSTIC_REGISTRY`** (static fallback kept for lenient imports), and diagnostics on an otherwise-clean rollout now render as neutral info banners instead of red — so #1025's chat-only flag will slot in without reading as a failure.

Would appreciate a re-test whenever you have time 🙏

bingran-you:
@ljr145733 — thanks for the substantial first contribution. I reviewed commit `d26931b720bcedae698a8795aafe6d31373d500d`.

## Review — changes requested

The viewer direction is valuable, but I don’t think this implementation is maintainable or yet matches the intended product design. Please address the following before merge.

### 1. Decompose the viewer architecture

`src/benchflow/trajectories/viewer.py` grows from 872 to 1,492 lines, while the new template adds another 917 lines—roughly 278 lines of CSS and 607 lines of global-state JavaScript.

One module now owns:

- Legacy rendering
- ACP normalization
- Diagnostic inference
- Verifier/metrics loading
- Run discovery and catalog creation
- Local and Hugging Face source resolution
- Two HTTP-serving modes

Please separate responsibilities, for example:

```text
trajectories/viewer/
  models.py       # Typed payload/view-model contract
  payload.py      # ACP normalization and diagnostics
  sources.py      # LocalPathSource / HfDatasetSource
  catalog.py      # Run discovery and summaries
  server.py       # HTTP routes
  legacy.py       # Existing raw-session renderer
  assets/
    theme.css
    viewer.css
    viewer.js
    template.html
```

Preserve `benchflow.trajectories.viewer` through `__init__.py`.

Also replace the loosely coupled `dict[str, Any]` Python → JavaScript boundary with an explicit typed payload contract. Tool-kind classification and visual tokens are currently duplicated across Python and JavaScript and can drift.

### 2. Match SkillsBench’s actual visual system

The current CSS approximates SkillsBench but does not implement it. Declaring `font-family: Satoshi` does not load Satoshi; the generated page has no font import or `@font-face`, so most systems will silently use a fallback.

Please use the live [SkillsBench site](https://www.skillsbench.ai/) as the visual source of truth:

- UI font: `Satoshi`
- Monospace font: `Google Sans Code Variable`
- Background: `lab(97.6714% .0993311 -.364053)`
- Foreground: `lab(2.75381% 0 0)`
- Primary: `lab(7.78201% -.0000149012 0)`
- Border: `lab(90.952% 0 -.0000119209)`
- Radius: `.625rem`
- Canonical semantic tokens: success, warning, info, destructive and chart colors

Extract these into one shared theme rather than maintaining separate approximations in `viewer.py` and `viewer_template.html`. Keep the exported viewer self-contained by packaging/inlining the required fonts, subject to their licenses.

### 3. Adopt PostTrainBench’s information architecture, not its visual styling

With the PR’s example dataset, the viewer renders 225 runs as one flat 280px sidebar. There is no run-level filtering, grouping, sorting or pagination. This will not scale.

Use the [PostTrainBench trace browser](https://posttrainbench.com/traces/index.html) and its [index interaction logic](https://posttrainbench.com/traces/assets/index.js) as the layout/behavior reference:

**Browse screen**

- Overview/performance matrix
- Corpus statistics
- Search and filters for benchmark, model and agent
- Grouping by task, task × model, experiment or none
- Sorting by reward/accuracy, duration, turns and cost
- Collapsible groups
- Incremental “show more” pagination
- URL-preserved filter and navigation state

**Run-detail screen**

- Sticky summary rail
- Trace / Verifier / Metrics tabs
- Focus mode by default, with Full as an option
- Jump-to-turn and event anchors
- Back navigation that preserves the browse state

The [PostTrainBench run-detail logic](https://posttrainbench.com/traces/assets/run.js) is a useful behavioral reference. Please copy the information architecture, not its cream/monospace visual identity—the BenchFlow viewer should use SkillsBench styling.

### 4. Tighten the `hf://` download allowlist

`_HF_VIEWER_FILES` contains `verifier/*`, but `_load_verifier()` consumes only:

```text
verifier/reward.txt
verifier/ctrf.json
verifier/test-stdout.txt
verifier/test-stderr.txt
```

A live resolution of the PR’s example dataset downloaded 1,516 files. Of those, 418 verifier files—approximately 31.3 MB—were not consumed by the viewer, including large OBJ, MP4, PDF, NPZ and log artifacts.

This contradicts the claim that only viewer-relevant files are fetched. Replace the wildcard with the four precise paths and add a regression test that rejects any broader `verifier/` pattern.

Please also stop passing `hf://` through a `Path`, which mutates it into `hf:/` and forces the parser to accept the corrupted spelling. Parse the CLI value as a string into an explicit `LocalPathSource | HfDatasetSource` type and validate dataset subpaths.

### 5. Add browser-level regression coverage

The Python payload tests are useful, but none of the approximately 607 lines of JavaScript interaction logic is exercised in a browser.

Please add browser-level tests covering at least:

- Filtering, grouping, sorting and URL-state restoration
- Selecting a run and preserving back navigation
- Trace/Verifier/Metrics tabs
- Focus/Full switching
- Search, jump-to-turn and event anchors
- Run-loading and malformed-payload error states
- Untrusted trajectory content remaining text rather than HTML
- Representative desktop and narrow viewport screenshots

### 6. Preserve the promised “full event stream”

Unknown event payloads and diagnostics are silently truncated at 2,000 characters. That conflicts with the advertised full event stream. Render complete expandable payloads, or explicitly label truncated previews and provide access to the complete content.

## Suggested PR split

Since you offered to split the work, I recommend:

1. Typed viewer payload, ACP single-run rendering and shared SkillsBench theme
2. Scalable run catalog/browser using the PostTrainBench interaction model
3. `hf://` source adapter with a precise download allowlist

## Verification performed

- Focused tests: `97 passed`
- Ruff, formatting and type checks passed
- `git diff --check` passed
- The live viewer was exercised with 225 discovered runs
- The SkillsBench and PostTrainBench reference sites were inspected interactively

These checks show that the current behavior largely works; the requested changes are about scalability, source boundaries, visual fidelity and long-term maintainability.

ljr145733:
All six review points should be addressed now:

1. `viewer.py` split into a package with a typed payload contract (5a7bbfc3, e34c9138)
2. theme rebuilt on the actual skillsbench.ai tokens, Google Sans Code embedded under OFL — Satoshi isn't redistributable, so it falls back (a490d488)
3. browse mode got a real catalog: group by task / model+harness, sort, filter, URL-state restore (5d1c56fe). As discussed, the overview matrix is deferred for now
4. `hf://` downloads use an exact file allowlist and typed source parsing (a4c2f76c)
5. browser-level tests driving the template in real Chromium (16e10828) — CI now installs chromium so they actually run instead of self-skipping
6. server-side truncation removed; long text collapses client-side instead (e34c9138)

After that I did another adversarial pass over the catalog code and fixed what it turned up in 79695125. The two worst: a failed `/api/rollout` load could leave the previous run's data on screen with the error invisible, and the inherited `do_HEAD` served files from the process cwd, bypassing the rollout whitelist. Each fix has a browser regression test.

Take another look when you have time — happy to adjust anything.

bingran-you:
Maintainer merge pass complete on b88d6f47: rebased onto current main; hardened the payload, source, server, confirmation, and browser boundaries; split the frontend state machine; and added mutation, security, accessibility, and real-Chromium regressions. Local focused suite: 126 passed; browser suite: 30 passed. Native GitHub test, pip-audit, integration-scope, and integration-light all pass. manifest-parity is the pre-existing external benchflow-ai/agents drift (codex-acp 0.0.45 there versus 1.6.0 on main) and has also failed the last three main runs. Version remains 0.7.6.dev0; no release requested.

PR #1035, opened 2026-08-17
Add --publish-bucket and --eval-results-* flags to bench eval run

NathanHB:
Extends the existing \`--publish-hf\` mechanism with two more HF options on \`bench eval run\`:

- \`--publish-bucket <bucket_id>\` — syncs job artifacts to a HF Storage Bucket (\`publish_folder_to_bucket\`, alongside the existing \`publish_folder_to_hf\` dataset-repo path).
- \`--eval-results-model/-dataset/-task\` — opens a community eval-results PR (\`.eval_results/*.yaml\`) on a model's own HF repo, scored from the run's mean reward (reuses \`build_health_summary\`).

Both are lazy-imported (\`huggingface_hub\`), same pattern as \`publish/huggingface.py\`'s existing functions — no new dependency.

Test plan:
- \`ruff check\`/\`format\`, \`ty check\` pass on touched files
- \`pytest tests/test_eval_artifact_cli.py tests/test_cli_arg_validation.py\` pass
- smoke-tested reward aggregation + new CLI flags against a synthetic job dir

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -287,6 +287,11 @@ bench eval run --tasks-dir ./tasks --matrix matrix.yaml --trials 3
 | `--matrix` | — | YAML model matrix for repeated evals; currently requires `--tasks-dir` |
 | `--trials` | `1` | Number of trials for `--matrix` |
 
+`--publish-hf`/`--publish-bucket` also write a `README.md` run summary
+(agent, model, per-task reward and any error/verifier issue, deduplicated
+across retries) into the job dir before upload — buckets render it on the
+directory page automatically.
+
 See [Architecture: skill loading](../architecture.md#skill-loading) for how
 `with-skill` mode is registered with each agent.
 
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -278,8 +278,12 @@ bench eval run --tasks-dir ./tasks --matrix matrix.yaml --trials 3
 | `--retry-attempts` | — | Override retry attempts for the eval run |
 | `--retry-concurrency` | — | Reserved retry concurrency setting recorded in run config |
 | `--publish-hf` | — | Upload final eval artifacts to this Hugging Face dataset repo |
-| `--hf-prefix` | — | Path prefix inside the Hugging Face repo; requires `--publish-hf` |
+| `--hf-prefix` | — | Path prefix inside the Hugging Face repo or bucket; requires `--publish-hf` or `--publish-bucket` |
 | `--hf-public-read-check` | `false` | Verify public Hugging Face reads after upload |
+| `--publish-bucket` | — | Upload final eval artifacts to this Hugging Face storage bucket |
+| `--eval-results-model` | — | Hugging Face model repo to open a community eval-results PR on |
+| `--eval-results-dataset` | — | Hugging Face benchmark dataset id for the eval-results entry, e.g. `org/benchmark` |
+| `--eval-results-task` | — | Benchmark `task_id`, as defined in the dataset's `eval.yaml` |
 | `--matrix` | — | YAML model matrix for repeated evals; currently requires `--tasks-dir` |
 | `--trials` | `1` | Number of trials for `--matrix` |
 
--- a/src/benchflow/cli/eval_artifacts.py
+++ b/src/benchflow/cli/eval_artifacts.py
@@ -66,6 +66,7 @@ def postprocess_eval_artifacts(
     eval_config = _apply_source_sidecar(eval_config, resolved_tasks_dir)
     from benchflow.eval_artifacts import (
         TaskManifestOptions,
+        build_health_summary,
         materialize_canonical_job,
         write_canonical_selection,
         write_health_summary,
@@ -154,6 +155,39 @@ def postprocess_eval_artifacts(
             print_error(str(exc))
             raise typer.Exit(1) from None
         console.print(f"[green]Published HF artifacts:[/green] {escape(published.url)}")
+    if req.publish_bucket is not None:
+        from benchflow.publish.huggingface import publish_folder_to_bucket
+
+        prefix = req.hf_prefix or job_dir.name
+        try:
+            published = publish_folder_to_bucket(
+                job_dir, bucket_id=req.publish_bucket, path_in_repo=prefix
+            )
+        except ValueError as exc:
+            print_error(str(exc))
+            raise typer.Exit(1) from None
+        console.print(f"[green]Published HF bucket:[/green] {escape(published.url)}")
+    if req.eval_results_model is not None:
+        from benchflow.publish.huggingface import open_eval_results_pr
+
+        rewards = [
+            row["reward"]
+            for row in build_health_summary(job_dir)["rows"]
+            if row["scored"]
+        ]
+        if rewards:
+            value = round(100.0 * sum(rewards) / len(rewards), 2)
+            try:
+                pr_url = open_eval_results_pr(
+                    model_repo=req.eval_results_model,
+                    dataset_id=cast(str, req.eval_results_dataset),
+                    task_id=cast(str, req.eval_results_task),
+                    value=value,
+                )
+            except ValueError as exc:
+                print_error(str(exc))
+                raise typer.Exit(1) from None
+            console.print(f"[green]Eval-results PR:[/green] {escape(pr_url or '')}")
 
 
 def _load_eval_matrix(path: Path) -> dict[str, dict]:
--- a/src/benchflow/cli/eval_artifacts.py
+++ b/src/benchflow/cli/eval_artifacts.py
@@ -68,6 +68,7 @@ def postprocess_eval_artifacts(
         TaskManifestOptions,
         build_health_summary,
         materialize_canonical_job,
+        render_run_summary_markdown,
         write_canonical_selection,
         write_health_summary,
         write_task_manifest,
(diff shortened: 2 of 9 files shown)


Galius5136:
Thanks for working on this! I went through both new paths locally and found a few things that may be worth fixing before merge.

The full test lane has one deterministic head-only failure: `test_eval_run_flags_match_cli_md_bidirectional`, because the new `--publish-bucket` / `--eval-results-*` flags aren't documented in `cli.md`.

I also hit a planner mismatch around `--hf-prefix`: `--publish-bucket my-bucket --hf-prefix runs/x` is rejected with `--hf-prefix requires --publish-hf`, even though the bucket post-processing path reads `req.hf_prefix`. So right now a custom bucket prefix isn't actually reachable.

A couple of smaller things I noticed: the bucket import path only catches `ModuleNotFoundError`, but older allowed `huggingface_hub` versions can raise `ImportError` for missing `create_bucket` / `sync_bucket`; and if `--eval-results-model` is set but no rollout has a score, the command silently does nothing — no PR, no warning, no explanation.

Happy to re-test if you push an update.

NathanHB:
Hey @Galius5136 ! Of course, this was not tested / done, only a quick first draft / POC. 
Working on adding suggestions and fixes and will be sure to tag you when ready, thanks for the early feedback :)

NathanHB:
@Galius5136 pushed fixes for all four:

- **Doc drift**: added `--publish-bucket` / `--eval-results-model` / `--eval-results-dataset` / `--eval-results-task` to the `bench eval run` table in `docs/reference/cli.md`. `test_eval_run_flags_match_cli_md_bidirectional` passes now.
- **`--hf-prefix` gate**: was hard-coded to require `--publish-hf`. Now accepts `--publish-hf` *or* `--publish-bucket`, matching what the bucket postprocess path actually reads (`req.hf_prefix`). Verified `--publish-bucket x --hf-prefix runs/y` no longer raises.
- **Import handling**: both `publish_folder_to_bucket` and `open_eval_results_pr` now catch `ImportError` (not just `ModuleNotFoundError`), so an installed-but-too-old `huggingface_hub` missing `create_bucket`/`sync_bucket` fails with the intended clear `ValueError` instead of an unhandled error.
- **Silent no-op**: `--eval-results-model` with zero scored rollouts now prints a yellow warning instead of doing nothing.

Also ran this for real against a live benchmark (skillsbench `citation-check`, `opencode` agent, `deepseek-ai/DeepSeek-V3.1` via HF Inference Providers routed through the generic `vllm/` provider) with `--publish-bucket` — confirmed the real trajectory/result artifacts land correctly in the bucket. That's actually how I caught the `--hf-prefix` bug independently before seeing your comment.

`ruff check`/`format`, `ty check`, and the full `test_cli_docs_drift.py` + `test_eval_artifact_cli.py` + `test_cli_arg_validation.py` suites (56 tests) pass.

NathanHB:
Ran a live end-to-end example against the fixed code to confirm `--publish-bucket` (and the `--hf-prefix` fix) actually works, not just unit tests:

```bash
export BENCHFLOW_PROVIDER_BASE_URL=https://router.huggingface.co/v1
export OPENAI_API_KEY=$(cat ~/.cache/huggingface/token)   # HF token, vllm provider's auth_env

bench eval run \
  --source-repo benchflow-ai/skillsbench --source-path tasks \
  --include citation-check \
  --agent opencode \
  --model vllm/deepseek-ai/DeepSeek-V3.1 \
  --sandbox docker \
  --publish-bucket SaylorTwift/benchflow-dummy-test \
  --hf-prefix dummy-run-verify
```

This routes the model through HF Inference Providers (`router.huggingface.co`) via the generic `vllm/` provider slot — `vllm/<anything>` passes straight through as the model name, which lines up with how HF Inference Providers takes plain repo ids.

Result: real agent turn (11 tool calls, clean `end_turn`), then `--publish-bucket` created/found the bucket and synced all 18 real artifacts (2.5MB, including the actual `llm_trajectory.jsonl` from the DeepSeek-V3.1 session) to:

```
hf://buckets/SaylorTwift/benchflow-dummy-test/dummy-run-verify/
```

Verified independently on the Hub (not just the CLI's own "Published" message) — `result.json` there matches this exact run (`model: vllm/deepseek-ai/DeepSeek-V3.1`, `agent: opencode`, task source pinned to skillsbench commit `9a1f4dd`).

Files landed:
```
dummy-run-verify/
  adp.jsonl, results.jsonl, summary.json, verifiers.jsonl
  citation-check__43c8462b/
    config.json, prompts.json, result.json, results.jsonl, timing.json
    agent/{acp_trajectory.jsonl, install-stdout.txt, opencode.txt}
    trainer/{adp.jsonl, atif.json, verifiers.jsonl}
    trajectory/{acp_trajectory.jsonl, llm_trajectory.jsonl}
    verifier/test-stdout.txt
```

(The task itself scored 0/1 — `citation-check`'s verifier produced no output on this box, a local docker/colima setup issue unrelated to this PR's HF publishing code.)

NathanHB:
Added a new bit: both `--publish-hf` and `--publish-bucket` now write a `README.md` run summary into the job dir before upload (results + issues), reusing `build_health_summary`. Buckets render `README.md` on the directory page per the Hub's docs, so this shows up automatically.

Ran it for real again (same setup: skillsbench `citation-check`, `opencode`, `deepseek-ai/DeepSeek-V3.1` via HF Inference Providers, `--publish-bucket`). The uploaded `README.md`:

```markdown
# BenchFlow run summary

- Agent: `opencode`
- Model: `vllm/deepseek-ai/DeepSeek-V3.1`
- Tasks: 1
- Scored: 0 / Unscored: 1
- Mean reward: n/a
- Rows with tool calls: 1

## Tasks

| Task | Reward | Tool calls | Issue |
| --- | --- | --- | --- |
| citation-check | — | 14 | verifier crashed: No reward file found at .../verifier/reward.txt or .../verifier/reward.json |
```

(the "Issue" column picking up that unrelated local verifier crash is exactly the intended behavior — surface it instead of hiding it)

Added `test_eval_run_publish_bucket_writes_readme_summary` covering this through the real CLI (`bench eval run --publish-bucket ...` with the network call mocked). `ruff`/`ty check`/full test suite green (58 tests).

NathanHB:
Pushed fixes for all 4 review points, plus a new bit: `--publish-hf`/`--publish-bucket` now also write a `README.md` run summary (agent, model, reward, per-task issues) before upload. Tested live against a real DeepSeek-V3.1 run via HF Inference Providers + a real bucket upload — works. Tests + ruff + ty all green.

https://huggingface.co/buckets/SaylorTwift/benchflow-dummy-test/tree/readme-feature-verify

NathanHB:
Hey @Galius5136 ! SUper gentle ping, this is ready to review i think :)

Galius5136:
Thanks for the follow-up @NathanHB! I re-tested `0409da6e`, and all four issues from my previous review are fixed.

I also exercised the new publish/README path and found a few things worth tightening up:

1. **Retry attempts are treated as independent task rows in both the README and eval-results path.**  
   `render_run_summary_markdown` and the `--eval-results-model` path both consume `build_health_summary(job_dir)["rows"]` without canonicalizing retries first. On a real job with one task and four rollout dirs, the README reports `Tasks: 4` and renders four rows for the same task.

   This also affects correctness beyond presentation: if a superseded retry attempt carries a reward, it can be included in the mean used for the eval-results `value`. I did not reproduce that exact scored-superseded shape end-to-end, but both consumers currently operate on the same uncanonicalized rows.

   `build_health_summary` already accepts `canonical_selection=`, so using that (or otherwise deduplicating by `task_id`) looks like one fix for both surfaces.

2. **The printed HF bucket URL is broken.**  
   The actual upload works, including `--hf-prefix`, and the README is present and rendered in the target bucket path. The URL printed by the CLI returned 404.

   `publish_folder_to_bucket` appears to be the only publish helper in `huggingface.py` constructing its URL manually; `publish_folder_to_hf` and `publish_file_to_hf` use `_tree_url()`. Reusing the common URL helper/pattern seems safer, though the correct revision segment for buckets is worth checking separately.

3. **Eval-results publishing clobbers previous entries.**  
   I verified this end-to-end on HF: after merging a first result for `bench-a`, publishing `bench-b` rewrites `.eval_results/benchflow.yaml` from the existing entry to only the new one.

   A safer approach would be to load the existing list, upsert the matching `(dataset, task_id)` entry, and preserve unrelated entries. I’d also add a regression test covering two sequential publishes.

4. **The README `Issue` column can become very noisy.**  
   In the real retried job, the same ~450-character escaped JSON error was rendered in full three times. Truncating the cell value would make the summary much easier to scan; `truncate_end` already seems to be used for this kind of output elsewhere.

5. **The README publish behavior is currently undocumented.**  
   `cli.md` documents the publish flags themselves, but not that publishing now generates and uploads a `README.md`. A short note in the publish docs would make the new behavior explicit.

One smaller testing note: the README path has coverage, but mutation-checking the original fixes still leaves the `--hf-prefix` gate fix, Hub `ImportError` handling, and no-scored-rollouts warning without focused regression tests.

Happy to re-check after another patch.

NathanHB:
Fixed all 5 — retry dedup (reused build_canonical_selection), bucket URL (/tree/ not /resolve/), eval-results now upserts instead of clobbering, README issues truncated, docs updated. Added tests for all of it. Verified live: URL now returns 200, retry job shows Tasks: 1, long error is truncated.

Galius5136:
Re-reviewed the latest head and re-ran the relevant paths.

The round-2 findings are fixed:

- retry attempts are now deduplicated correctly in published summaries;
- HF bucket links use the working `/tree/` URL;
- eval-results publishing preserves existing entries instead of clobbering them;
- the remaining formatting/docs fixes are in place.

I also re-checked the two important live paths against Hugging Face:

- **eval-results upsert:** published `bench-a`, merged it, then published `bench-b`; the second PR preserved both entries, and both remained present on `main`.
- **retry dedup:** forced a real retryable install failure through the normal evaluation loop. One task produced **3 physical rollout directories** (initial attempt + 2 retries), while the generated README reported **`Tasks: 1`** and `summary.json` reported **`total: 1`**.

The bucket upload from that run also returned the expected `/tree/...` URL.

Static/type checks are clean, and the full-suite failures remain the same baseline failures rather than head-only regressions.

Ready to merge from my side!

PR #1036, opened 2026-08-18
feat(release): archive releases on Zenodo and gate CITATION.cff at tag time

Yiminnn:
## Summary

Sets up Zenodo archiving for public releases, and fixes the metadata drift that
made our citation file unciteable.

Zenodo's GitHub integration archives every published GitHub Release, snapshots
the tag's source tree, and mints a version DOI (plus a concept DOI that always
resolves to the newest version — that is the one to cite). Without a
`.zenodo.json` in the repo, Zenodo derives **authorship from GitHub contributor
statistics**, which is not our author list. So this PR adds that file.

`.zenodo.json` deliberately omits `version` and `publication_date`: Zenodo takes
both from the tag and the Release, so pinning them here would go stale on every
release — exactly the bug the rest of this PR fixes.

`CITATION.cff` had drifted to `0.6.9 / 2026-08-15` while five 0.7.x releases
shipped, because nothing checked it. It now names `0.7.4 / 2026-08-17`, and the
tag-driven release validates it:

- `tools/release_version.py public-release` now also reads `CITATION.cff` and
  fails when its `version` differs from the release tag, or when
  `date-released` is in the future.
- The check runs in the existing "Validate release tag" step, **before** build
  and publish. A PyPI publish and the GitHub Release that triggers Zenodo
  archiving are irreversible; a failed validation is not.
- Everyday CI deliberately does **not** require `cff == pyproject`.
  `CITATION.cff` names the *last published* release, so while `main` sits on a
  `.devN` version the citation file is legitimately one release behind. That is
  correct, not drift, and the tag-time gate is what keeps it honest.

`docs/release.md` gains a `## Zenodo Archiving` section (mechanism, plus the
admin-only one-time setup steps), and the public-release step list now covers
bumping `CITATION.cff` and leaving it on the version just released.

## Blocked on: the author list

`.zenodo.json`'s `creators` is a placeholder, and
`tests/test_citation_metadata.py::test_zenodo_creators_are_resolved` **fails on
purpose** while that placeholder is present. CI is red by design until real
authorship lands — a published DOI cannot be withdrawn, so placeholder authors
must not be able to reach `main`.

Needed per author: name as `Family, Given`, the order they should appear in,
ORCID, and affiliation.

## Ordering constraint for whoever flips the switch

The Zenodo switch requires **admin** rights on this repo, and it must be flipped
**only after** a `.zenodo.json` with the real author list has merged to `main`.
Zenodo archives the tagged commit, so with our release cadence the very next tag
would otherwise mint a permanent DOI with contributor-statistics authorship.
Steps are written out in `docs/release.md`.

Also note Zenodo archives releases published *after* the switch is flipped, so
the first archived version will be the next release, not v0.7.4.

## Test plan

- [x] `uv run --extra dev python -m pytest tests/` — 1184 passed, 6 skipped, and
      the 1 intentional failure above.
- [x] `uv run ruff check .`, `uv run ruff format --check`, `uv run ty check src/ tools/`.
- [x] Ran the real workflow command, not just unit tests:
      `pyproject` bumped to 0.7.5 with `CITATION.cff` left at 0.7.4 and tag
      `v0.7.5` is rejected with `CITATION.cff records version '0.7.4' but this
      release publishes '0.7.5'`; a future `date-released` is rejected; tag
      `v0.7.4` on this branch passes and emits `version=0.7.4`; and
      `internal-preview` still runs under `--with packaging --with pyyaml`.
- [x] `CITATION.cff` validates against CFF 1.2.0 (`cffconvert --validate`).
- [x] `.zenodo.json` validates against Zenodo's legacy deposit JSON Schema with
      0 errors, and `isVariantFormOf` was checked against Zenodo's live
      relation-type vocabulary rather than assumed.
- [ ] First real archive: after the next release, confirm on Zenodo that the
      record's authors/title/license came from `.zenodo.json`.

## Follow-up (out of scope here)

The concept DOI does not exist until the first archived release, so adding the
DOI badge to `README.md` and the DOI to `CITATION.cff` has to be a second PR.

## Notes on the YAML edge cases

Two things the parser has to survive, both covered by tests: YAML parses an
unquoted `version: 1.0` as a **float** (while `0.7.4` is a string), so the
comparison goes through PEP 440 rather than string equality; and an unquoted
`date-released` arrives as a `datetime.date` while a quoted one arrives as a
string. The future-date check also tolerates one day of skew, because a releaser
east of UTC tagging after local midnight legitimately writes a date that is
still "tomorrow" for the runner.

### DIFF
--- a/.github/workflows/internal-preview-release.yml
+++ b/.github/workflows/internal-preview-release.yml
@@ -105,7 +105,8 @@ jobs:
       - name: Compute internal preview version
         id: version
         run: >-
-          uv run --with packaging --no-project python tools/release_version.py
+          uv run --with packaging --with pyyaml --no-project python
+          tools/release_version.py
           internal-preview
           --run-number "${{ steps.provenance.outputs.source_run_number }}"
 
--- a/.github/workflows/public-release.yml
+++ b/.github/workflows/public-release.yml
@@ -44,11 +44,16 @@ jobs:
             exit 1
           fi
 
+      # Runs before the build so a stale CITATION.cff stops the release while it
+      # is still reversible; PyPI publishes are not. PyYAML is explicit because
+      # `--no-project` skips the project's own dependencies.
       - name: Validate release tag
         id: version
         env:
           TAG_NAME: ${{ github.ref_name }}
-        run: uv run --with packaging --no-project python tools/release_version.py public-release
+        run: >-
+          uv run --with packaging --with pyyaml --no-project
+          python tools/release_version.py public-release
 
       - name: Build distributions
         run: uv build --no-sources
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,0 +1,28 @@
+{
+    "upload_type": "software",
+    "title": "BenchFlow: framework for RL environments for LLM agents",
+    "description": "BenchFlow is a framework for building RL environments to evaluate and train LLM agents. Built on the Agent Client Protocol (ACP), it provides Scene-based multi-turn, multi-agent, and multi-model evaluation in shared sandboxes — without Docker Compose or sidecar containers. Supported use cases include interactive user simulation, code-review loops, bring-your-own-skill (BYOS) skill generation, multi-turn iterative refinement, cross-model review (cheap coder + strong reviewer), and stateful service tasks against live mock APIs (Gmail, Calendar, Docs, Drive, Slack). See docs/use-cases.md.",
+    "creators": [
+        {
+            "name": "TODO BEFORE MERGE - replace with the agreed author list (\"Family, Given\" per entry, plus orcid and affiliation where known). Zenodo mints a permanent DOI from this file, so a placeholder that reaches main becomes permanent authorship."
+        }
+    ],
+    "license": "Apache-2.0",
+    "access_right": "open",
+    "keywords": [
+        "benchmark",
+        "llm-agents",
+        "acp",
+        "agent-evaluation",
+        "multi-turn",
+        "terminal-bench",
+        "skillsbench"
+    ],
+    "related_identifiers": [
+        {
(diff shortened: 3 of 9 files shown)


Yiminnn:
@xdotli — this needs the author list from you before it can go anywhere.

`.zenodo.json` is what Zenodo reads to mint the DOI for every published release.
If it is absent, Zenodo derives authorship from GitHub contributor statistics
(i.e. a list of accounts ranked by commit count), so the file has to carry the
real list instead. Right now `creators` is a placeholder, and
`tests/test_citation_metadata.py::test_zenodo_creators_are_resolved` fails on
purpose while it is — CI is red by design so a placeholder cannot reach `main`.
A published DOI cannot be withdrawn.

Per author, please drop in a comment:

1. Name as `Family, Given` (e.g. `Li, Xiangyi`) — Zenodo's expected form.
2. The order they should appear in.
3. ORCID, if they have one.
4. Affiliation.

Two open questions worth a call while you are here:

- **Individuals, an organizational author, or both?** `CITATION.cff` currently
  says `BenchFlow team` while `pyproject.toml` names you and @kywch. Zenodo
  supports an organizational creator, but citations then carry no personal
  names, and nothing links to an ORCID. This PR leaves `CITATION.cff`
  authorship untouched — whichever way you go, I will make both files say the
  same thing.
- **Should any related paper or preprint be linked?** If there is an arXiv ID or
  DOI, I will add it to `related_identifiers` as `isDocumentedBy` now rather
  than editing the record's metadata after the fact.

Once you post the list I will drop it in, CI turns green, and I will mark this
ready for review. Note that flipping the Zenodo switch needs repo **admin** and
must happen *after* this merges — `docs/release.md` has the steps and the reason
the order matters.

Issue #917, opened 2026-07-11
Missing agent credentials should be clean non-retryable rollout errors


bingran-you:
Mirrors Linear ENG-257: https://linear.app/benchflow/issue/ENG-257/missing-credential-surfaces-as-an-in-rollout-traceback-instead-of-a

## Problem

A missing required model credential currently surfaces through a generic exception path in rollout setup. That can produce noisy traceback-style logs and an ambiguous retryable/error classification even though the operator action is clear: provide the missing key.

## Expected

Missing agent/model credentials should become a clean rollout error, classify as `provider_auth`, and avoid the generic `Run failed` traceback logger. This should stay focused on missing credential classification, not broad upfront preflight that could false-block injected runtime envs.

## Current fix path

Daily scan 2026-07-11 confirmed this is BenchFlow repo-owned and narrow enough for a focused PR.

bingran-you:
Daily scan follow-up (2026-07-12): #918 is the active BenchFlow fix PR for this issue and is currently green on GitHub checks.

Current state:
- PR #918 closes this issue and adds clean missing-credential classification/presentation coverage.
- CI is green on head `ccb8284cd62e37ab898393fd74e07bbd6b805238` (`test`, `pip-audit`, `manifest-parity`, `integration-light / rollout-smoke`; integration-scope plan/matrix skipped as expected for this scope).
- No additional BenchFlow code change started from today's scan; remaining gate is human review/merge of #918.

Labels updated: added `reviewing` while the fix PR is open.

PR #790, opened 2026-06-16
feat(eval): bind the environment (S) and config (C) axes at the CLI

xdotli:
## What

Bind two `I_eval` axes (Han Lee, *Hidden Technical Debt: Agent Evaluation Infrastructure* — "decouple the task from the harness") at the CLI, decoupled from the task, the way `--agent`/`--model`/`--sandbox` already are:

- **S (environment)** — `load_manifest` now resolves a `name@version` spec via `$BENCHFLOW_ENV_REGISTRY` (content-addressed), in addition to a file path. So `--environment-manifest env0@v2` binds a registry environment per run; the task references the env by name, the run picks the version. Mirrors `resolve_dataset`.
- **C (config)** — new `--config-override` deep-merges a config patch (inline JSON/YAML/TOML or `@file`) into each task's resolved `TaskConfig`, so one knob (e.g. `agent.timeout_sec` — "lower turn budget") can be varied while T/A/M/S/R stay fixed.

## How (C-axis threading)

Parsed **once at plan time** and threaded as **typed data** — `EvalCreateRequest → EvalPlan → EvaluationConfig → RolloutConfig` (including the sharded worker payload), exactly like `environment_manifest`. Applied at the **rollout layer** (not the `Task` constructor), re-validated through `TaskConfig`, **allowlist-scoped** to non-scoring sections (`agent`/`sandbox`/`metadata`) so a run-time overlay can never weaken the verifier/reward/oracle, and **persisted to `config.json`** by content hash for replay.

## Tested

- **Unit** — 36 cases (`test_config_override.py`, `test_env_registry.py`): deep-merge semantics, parse JSON/YAML/TOML/`@file`, fail-closed allowlist, `sandbox` alias round-trip, content-addressing, back-compat.
- **Dogfood** (ClawsBench `archive-amazon-shipping`, Docker + Daytona):
  - S: same task, env bound by `name@version` at the CLI — `env0@v1` (gmail present) → reward **1.0** vs `env0@v2` (gmail removed = "tool outage") → **0.0**, attributable to the env axis alone.
  - C: `--config-override '{"agent":{"timeout_sec":5}}'` with deepseek-v4-pro → agent killed at 5s, reward **0.0** (the budget is genuinely consumed); provenance persisted to `config.json`; the allowlist fails fast at plan time.
- **Reviewed** with a structural-quality pass; its one P0 — the overlay had been riding a process-global env var — is fixed here by the typed-pipeline threading above.

[...]

PR #1046, opened 2026-08-20
feat(branch): composed snapshotting, stage-level branching, and bench eval ablate

JeremyJC67:
## Summary

Implementation of [#1045's RFC](https://github.com/benchflow-ai/benchflow/pull/1045) (read that first): composed environment + container checkpoints, stage-boundary snapshots, per-child branch deltas with lineage artifacts, a replay cut-point, and `bench eval ablate` — infrastructure for attributing a rollout's failure to a stage and for running ablations as *controlled* comparisons rather than independent re-runs.

Commit-by-commit, each independently reviewable:

| Commit | What it does |
|---|---|
| `docs(branch)` | the RFC (same content as #1045) |
| `feat(branch): compose sandbox + environment checkpoints` | `checkpoint_composed`/`restore_composed`; `snapshot_layers` (default `{"environment"}` = today's behavior). The engine gated on `supports_snapshot` but never called `Sandbox.snapshot()`; now it composes both layers in the documented order, and **sandbox-only branching lets stateless environments branch at all** |
| `feat(branch): per-child deltas + lineage artifacts` | `BranchDelta`, content-addressed like #790; `tree.json`, per-child artifact dirs, `source_provenance kind="benchflow-branch"`. Branched runs previously left **no** artifacts |
| `feat(continue): replay cut-point` | `--max-exchanges`: replay the first K recorded exchanges then go live; stage-tagged cuts; served-vs-configured accounting |
| `feat(branch): stage-boundary snapshot policy` | the four cascade stages as a validated taxonomy; opt-in auto-capture (`RolloutConfig.snapshot_stages`, default empty = zero overhead); `mark_stage()`, `branch_at_stage()`, `stage_snapshots.json` |
| `feat(branch): execute skill_mode deltas` | a child forked from the `env-ready` snapshot runs as a fresh rollout over the restored sandbox and re-runs skill deployment under the switched mode |
| `feat(cli): bench eval ablate` | run a task, snapshot at a stage, fork one arm per delta, print an attribution table, write `ablation.json` |
| `feat(ablate): per-test attribution` | a scalar tie can no longer hide a behavioral difference — see *What the measurements taught us* |
| 9 × `fix(...)` | five found by **real runs on real docker**, four by adversarial review — all listed below |

## Test plan

- [x] Targeted: `pytest tests/test_branch*.py tests/test_ablate_cli.py tests/test_sandbox_snapshot_contract.py tests/continue_run/ tests/test_cli_docs_drift.py tests/test_cli_live_progress.py -q` → **`371 passed, 4 skipped`**
- [x] Full suite → `6156 passed, 94 skipped, 0 failed`
- [x] `ruff check .` / `ty check src/` → All checks passed
- [x] Docker live proofs (`-m live`, real daemon) → **`5 passed`**: zero-delta round trip at the container layer (+ negative control), composed sandbox+sqlite round trip, a PASS→FAIL→PASS reward-proxy invariant, restored containers keep their bind mounts and host-visible writes
- [x] **Real end-to-end ablations** (`codex-acp` / `us-openai/gpt-5.4-mini`, docker), cross-checked per arm against each child's own `verifier/reward.txt` and `verifier/ctrf.json` on the host

## What the measurements taught us

Running this for real, rather than against fakes, changed the design twice.

**1. A scalar tie can hide a large behavioral difference.** On a two-test task the skill pack flipped one sub-test *and* broke the other, netting exactly zero on the reward — so the tool printed *"no difference in this comparison"* while the arms behaved very differently. Mechanism verified, not inferred: the with-skill arm used the pack's prescribed Mann-Kendall/Sen's-slope method, the no-skill arm fell back to OLS and reproduced `p = 0.05479775402692069` bit-for-bit across four runs and two models.

`bench eval ablate` now mines per-test outcomes from each child's CTRF report (reusing `_failure_evidence`'s parser, not a second one) and reports the tests whose outcome differs:

```
│ with-skill │ 0.00 │ fail │ 203s │ matches no-skill at env-ready (both 0.00) — scalar rewards tie,
│            │      │      │      │ but 1 sub-test outcome(s) differ: test_trend_result            │

      Sub-test outcomes that differ (1)
┃ Test              ┃ with-skill ┃ no-skill ┃
│ test_trend_result │ passed     │ failed   │
1 test(s) tie across the arms and are omitted.
```

Arms with no per-test data degrade to scalar-only attribution and say so; a test named by only one arm reports `null`, never "failed".

**2. Branch children reproduce plain rollouts faithfully.** The branched `no-skill` child reproduced the plain `no-skill` rollout's per-test outcomes *and* that 17-significant-figure p-value; the branched `with-skill` child reproduced its per-test signature and pack usage. Measured cost: the branch route ran 3 rollouts in 656 s vs 1175 s for 2 sequential plain rollouts (1.8× on wall clock against a sequential baseline), at essentially identical token spend — it buys a byte-identical starting world and one environment build, not cheaper tokens. Both numbers are measured, not asserted.

### Bugs the real runs caught that the fakes could not

Each lives exactly in the seam where a test double replaces the real thing:

1. **A with-skill child could never deploy its skills** — the code injected `COPY skills` into a Dockerfile that is never built for a child adopting a restored sandbox, and `deploy_skills` treated that line as proof the pack was in the image.
2. **Every branch child's reward was silently reported as `0.00`** — `DockerSandbox.restore()` re-created the container without the original bind mounts, verifier outputs were never downloaded, and the child-runner turned "no reward" into a real-looking `0.0`. The command printed a confident attribution verdict computed from numbers that were never collected, exit code 0. Now: unscored ⇒ arm error, `reward: null`, no verdict, exit 1.
3. **`restore()` did not reproduce the container** — it now inspects the live container before removing it and replays binds/volumes/tmpfs, network and cpu/memory; "are verifier outputs mounted" is a live query of the daemon instead of a stale attribute, failing toward downloading; and the remaining fail-open path (container id unresolvable) now raises instead of silently producing a mountless container.
4. **Branch children clobbered the parent's on-disk artifacts** — children share the parent's mounts, so each child's verifier run overwrote the parent's `verifier/`. No reported number was wrong, but the evidence was.
5. **A `no-skill` arm could silently measure `with-skill`** — if the parent itself ran `with-skill`, the pack is baked into the image, so restoring that snapshot and "turning skills off" changed nothing while the report kept the label. Skill deltas now fail closed unless the parent's own mode is `no-skill` (the one mode that provably bakes nothing).

Plus, from adversarial review: a stale branch `value` could survive on a node when V is undefined; the stage-branch path skipped the non-empty-layer check the cursor path enforces.

## Known limitations, stated rather than left to be found

(Updated for the current head; the earlier version of this section predated the review-response commits and contradicted them.)

- All four delta axes execute (`skill_mode`, `injected_prompt`, `config_override`, and the service-level slice of `environment_ref`); image-changing environment deltas fail closed by design, with a typed error explaining why.
- Restore equivalence is partial: ports, capabilities/privileged/security-opt/sysctls, devices, ulimits, extra hosts, restart policy and multi-network membership are not replayed (documented at the call site). Env/workdir/user/entrypoint come from the committed image by design.
- Every branch child now emits a full `result.json`/`timing.json`/`prompts.json` set (fresh children natively; in-place children synthesized from their own captured state).
- Replay remains openhands-only; recording is already agent-agnostic.
- Agent-session snapshotting is out of scope for v1 (RFC §7); children get a fresh session with replayed-or-injected context.
- `--keep-snapshots` exports sandbox-layer images only; environment-layer sqlite snapshots live inside the container and are not separately exported. The per-stage lifetime annotation covers the whole entry honestly.
- `n_skill_invocations` is adapter-blind (it counts an ACP event `codex-acp` never emits), so it reads 0 even for arms that demonstrably read the pack. The ablate report deliberately does not surface it; a real fix is wider than this PR.
- Two arms of the same task cannot run concurrently outside `bench eval ablate`: image builds share the `bf__<task>:latest` tag across skill modes, so a concurrent pair races. The fail-closed skill check caught this as a contaminated run rather than mis-measuring it. Worth a per-config tag; not attempted here.

Happy to split this into the individual workstream PRs if that reviews better. Tracks [FrontierPhysics#73](https://github.com/benchflow-ai/FrontierPhysics/issues/73).

---

## Response to the exact-head review (2026-08-28)

All six findings are addressed in nine commits on this branch (`87f8efc7..f796f966`), each pinned by tests that were verified red on the pre-fix tip:

| Finding | Resolution |
|---|---|
| P1-1 artifact custody fails open | Custody chain fails closed: typed `ArtifactCustodyError`, hold dir preserved on any failure, `rmtree` only after every entry is confirmed moved back; a custody failure never masks a child's own exception (`87f8efc7`) |
| P1-2 divergence undetectable | Served vs recorded request digests (canonical-JSON over the comparable projection), per-exchange content comparison with recorded `divergence_events`, `strict_divergence` now trips on content; RFC workspace digest implemented (`workspace_digest.py`, recorded with basis or `null`+reason) (`07d4b090`) |
| P1-4 dead snapshot handle | `--keep-snapshots` exports the image via `docker save` to `<out-dir>/snapshots/<ref>.tar` (path+sha256 recorded) **before** cleanup; without the flag the report records `ephemeral: true, exported: null`; RFC §3.6 wording corrected (`59b02d17`) |
| P1-3 non-canonical config | `run_ablation` resolves the same per-task assembly `bench eval run` uses (extracted `task_rollout_config()`) and overlays only the ablation axis; parent and child configs now carry non-null `task_digest`, `reasoning_effort`, source identity, prompts, usage settings — pinned by tests (`df0d4360`) |
| P2 structure | Four-way split: `branch_policy.py` / `branch_transaction.py` / `branch_children.py` / `branch_report.py`; `branch()` ≈15 cc / 41 statements, `_run_children` successor takes one params object; behavior-neutral (identical full-suite results before/after, zero test edits) (`a3db4448`) |
| P1-5 stage wiring incomplete | Now end to end: stage captures record completed-LLM-exchange indices (`5c891c20`); `bench eval continue --cut-stage <name>` resolves recorded stages (`d02502c5`); `bench eval ablate --mark-research-end-on <path>` captures `post-research` when the marker file (e.g. `PLAN.md`) appears and branches there (`d40c3743`); task authoring accepts `branch_execution: forked-snapshot` and compiles it to a stage-capture request (`f796f966`) |

Gates at tip `f796f966`: targeted `371 passed, 4 skipped`; full suite `6156 passed, 0 failed`; `ruff check` / `ruff format --check` / `ty check src/` clean. Docker live suite at this exact tip: `5 passed in 104.85s` (real daemon, docker 29.6.0).

## Response to the second exact-head re-review (2026-08-29)

| Finding | Resolution |
|---|---|
| P1 workspace digests collide on legal filenames | Pipeline rebuilt null-safe (`find -print0 \| sort -z \| xargs -0`) and fail-closed (per-stage status checked; verified against busybox ash) — your exact repro is now a red-first test: two trees differing inside `file with spaces.txt` must digest differently, and a newline filename must never succeed wrongly (`ae311174`) |
| P1 plain-run snapshot refs look restorable but are deleted | Both halves: cleanup now rewrites `stage_snapshots.json` marking every ref `ephemeral: true` unless exported; `--keep-snapshots` added to `bench eval run` (same export machinery), and a tested import path exists — `bench eval import-snapshots <run-dir>` verifies the tar sha256, `docker load`s it, and checks the loaded image id against the recorded one, failing closed on any mismatch (`e22b6d45`) |
| P2 request-global settings fail late and retry in children | Two layers: static pre-flight in `build_eval_plan` rejects unsupported reasoning effort before any provisioning; deterministic ACP rejections classify as `request_global` (the #917 non-retryable pattern), joining retry exclusions, and `run_ablation` skips all children when the parent failed request-globally (`e5ed14c8`) |
| P2 structural gate | `ablate.py` 674 lines (+ `ablate_arms.py` 481); `run_ablation` passes `ruff --select C901,PLR0912,PLR0915` clean; behavior-neutral (identical suite results before/after) (`7ff1a632`) |
| Docs drift | `architecture.md` now states composed container+environment checkpoints are implemented (agent-session remains future work); RFC §3.6 and this PR's Known-limitations section updated to match the code (`8fc8404f`) |

Gates at the new tip: full suite `6190 passed, 0 failed`; docker live suite `6 passed` (grew by one: a live proof that a restored-from-tar image id matches the recorded one); `ruff` / format / `ty` clean.


Galius5136:
Really nice work! I especially like the fail-closed approach around branching and restore.

I ran the live Docker tests locally (`5 passed`, no `bf-snap-*` images left behind) and also exercised `bench eval ablate` with real Docker + `codex-acp`. I found two issues worth looking at:

1. **Per-test attribution is lost for in-place branches (`pre-verify` / `post-verify`).**  
   The child CTRF is preserved under `mounted/verifier/ctrf.json`, while the attribution reader looks at `verifier/ctrf.json`, so the report falls back to `scalar-only` even though the per-test data exists. I reproduced this live. As a control, `env-ready` fresh children correctly reported all per-test outcomes.

   A possible fix would be for `_outcomes()` to fall back to `child_dir / "mounted"` when the normal child artifact path has no CTRF.

2. **`_LinearState` does not restore `_timing` / `_verifier_error`.**  
   In an in-place branch, children mutate the shared `Rollout`; `_rewards` is restored correctly, but these result-bearing fields are not part of `_LinearState`. In my live run, `agent_execution` accumulated across both arms and the final parent `timing.json` contained the accumulated value.

   A possible fix would be to include `_timing` and `_verifier_error` in `_LinearState.capture()` / `restore_onto()` (and it may be worth checking `_diagnostics` / `_error` for the same reason).

Happy to re-run the live checks after a patch.

JeremyJC67:
Thank you — running it live and reproducing both issues is exactly the review this needed, and both were real. Fixed at `dfc385ae`, along with the three P1s the automated reviewer raised.

**1. Per-test attribution for in-place children** (`df22a98d`) — your diagnosis was right, and it was self-inflicted: the artifact-isolation commit moved a child's verifier output under `mounted/` and the attribution reader still looked at the old path. I took your suggested fallback but put the knowledge in the module that creates the layout rather than a second hard-coded path in `ablate.py`: `branch_artifacts.child_artifact_roots(child_dir)` returns the candidate roots best-first, and `ablate` tries each through the existing CTRF reader. Ordering matters for the fresh-rollout case, which can have both roots — its own downloaded copy wins over the mounted one, pinned by its own test. Neither root reporting still yields `tests: None`, never a fabricated row.

**2. `_LinearState` result-bearing fields** (`48bbb417`) — fixed, and the audit found more than the two fields. I went through every attribute `_build_result()` reads and compared against what `_LinearState` captured. Provably mutated by a child: `_timing` (your `agent_execution` accumulation), `_verifier_error` (a clean parent inherited a child's failure, and a genuinely failed parent had its own erased), `_diagnostics`, and — the one I would not have looked for without your report — `_native_usage_metrics`/`_native_usage_checkpoint`, which accumulate the same way and get promoted into `_usage_metrics` at cleanup, so **the children's tokens were being billed to the parent**. Also scoped `_error`, `_export_error`, `_evolved_skills`, `_usage_metrics`. Capture and restore now deep-copy — a captured reference is mutated in place by child *k* and restores nothing for child *k+1*, which the test pins directly. Fields left unscoped (setup-owned, `_started_at`, the cleanup-only provider caches) are documented in the module.

**3. Partial lineage on a raising child** (`743b2c2b`) — the child loop now writes the same lineage the success path writes before re-raising, so a partially completed experiment keeps `tree.json` and the completed arms' provenance. Artifact-write failures stay isolated: a full disk cannot replace the child's real exception on its way out, pinned by a test that makes both fail at once.

Also fixed from the automated review: `bench eval ablate` now resolves a task-declared `benchflow.environment.manifest` for the parent and every arm (the resolver moved into `benchflow.environment.manifest` so `bench eval run` and `bench eval ablate` share one path, and an unresolvable manifest fails closed before the parent starts); and host-mode continue-runs now stitch, bill and report the exchanges the router actually **served** rather than the configured cut, through a single artifact-writing call so the one-basis invariant is structural rather than a convention at two call sites.

Gates: targeted branch/ablate suite **340 passed, 4 skipped** (+11 regression tests, each confirmed red against the old code); full suite `5951 passed` with only the pre-existing `test_check_results_accepts_symlinked_current_repo_inferred_source`; `ruff check` / `ruff format --check` / `ty check` clean; docker live suite `5 passed`.

If you're still up for re-running: the two that would confirm your findings end-to-end are `--at-stage pre-verify` with two `inject:` arms (per-test rows should now appear instead of `scalar-only`), and any in-place ablation followed by checking the parent's `timing.json` and `result.json` usage against a plain single rollout. The RFC PR (#1045) also picked up the `env-ready` precondition wording the automated reviewer flagged.

Two things I deliberately left open, in case you have an opinion: `bench eval ablate` still has no `--environment-manifest` flag of its own (it honors only what the task declares), and `ablation.json` does not record which environment was bound — stamping it in would make the comparison self-describing.

JeremyJC67:
Update: the stack reached its intended final form, and the headline demo landed.

**All four delta axes now execute.** `config_override` children run as fresh rollouts with the delta deep-merged through the #790 allowlist machinery (composes with `skill_mode`/`injected_prompt` on one child); `environment_ref` executes its sound slice — service-topology deltas (the documented `env0@prod` → `env0@outage` outage pattern: same image, framework-owned lifecycle), with image-changing manifests failing closed via a typed error, because restoring the parent's container and rebuilding a different image contradict each other. The boundary is documented in the RFC. New arm kinds: `config:<inline-json-or-@file>` and `env:<registry-ref>`.

**Every child is now a first-class run**: in-place children synthesize their own `result.json`/`timing.json`/`prompts.json` from their own captured state (nulls where truth is unknown, never parent bleed-through), and `ablation.json` stamps the bound environment (name/env_hash/image) and the branched stage's snapshot refs — the two self-description gaps flagged in review.

**The demo: FrontierPhysics's flagship task, real signal.** `bench eval ablate` on `surface-ion-trap-shuttling` (at its last-merged commit; the package was retired from FP main two days ago), `codex-acp` / `us-openai/gpt-5.4-mini`, docker, `--at-stage env-ready --arms with-skill,no-skill`:

```
| Arm        | Reward | Result | Wall clock |
| with-skill | 1.00   | pass   | 311s       |
| no-skill   | 0.00   | fail   | 633s       |
Sub-test outcomes that differ (3):
  test_result_frequencies_are_derived_from_bem_and_anisotropy   passed | failed
  test_single_ion_inverse_engineered_waveform                   passed | failed
  test_piecewise_nine_ion_waveform_with_dwells                  passed | failed
(tying: test_nine_ion_equilibrium_spacings — both PASSED)
```

The per-test rows *are* the physics stages: the mentor skills flip the fragile electrostatics stage (no-skill guessed 2.2 MHz off trap geometry; with-skill ran the bundled FastLap BEM solve and landed on 4.295858 MHz) and both invariant-transport stages, while the stage the base model owns unaided (Coulomb equilibrium) ties. Trajectory evidence confirms the mechanism: the no-skill child never attempted a field solve; the with-skill child read the mentor skills as its first action and followed them. Both children fork from the byte-identical `env-ready` snapshot (`bf-snap-…7fb9e4a45aaa`), one recorded delta each; every number cross-checks against the on-disk verifier artifacts. As a bonus replication, an independent plain no-skill run died at the same stage with the same wrong number to six decimals.

**One genuine bug fixed along the way** (`8b7e6688`): the codex-acp 1.6.0 pin (#1044) dropped several models from the shim's catalog, and `session/set_model` now rejects even the session's *current* model when it's off-catalog — so every LiteLLM-routed codex run with such a model died in seconds despite the launch config having already installed the right model alias. The fix skips the doomed `set_model` exactly when the launch config already owns the session's model; in-catalog models keep the #1044 path, pinned by four tests.

Gates at tip: targeted suites green (25 new tests this round), full suite `5998 passed` with only the known pre-existing failure, `ruff check` / `ruff format --check` / `ty check` clean, docker live suite `5 passed`.

Galius5136:
Thanks for the follow-up. I re-ran the review on `8b7e6688`, including the regression gates, branch probes, Docker/live checks, and provider-backed checks where applicable.

The two issues from my original review are resolved: per-test attribution for in-place children is preserved correctly, and branch execution no longer leaks timing/verifier/usage state across arms or back into the parent.

I also verified the newer paths around `config_override`, `environment_ref`, explicit environment manifests, child result artifacts, and the codex-acp `session/set_model` fix; they behaved as intended. I independently reproduced the historical `surface-ion-trap-shuttling` oracle as well and got the expected `4.295858... MHz` FastLap result.

I found one small non-blocking edge case in `_split_arm_specs`: braces/brackets inside quoted inline JSON values can confuse the arm splitter, and commas in `config:@file` paths are not representable with the current grammar. I don’t think this needs to block the PR and it can be handled separately.

From my side, the PR is ready to merge, I don’t see anything that needs to be fixed before merging @bingran-you @JeremyJC67

JeremyJC67:
Thank you for the re-run — and for independently reproducing the FastLap oracle number; that makes the ion-trap demo a third-party-replicated result rather than my own claim.

The `_split_arm_specs` edge case is fixed at `bc712d7d`: the walk now tracks JSON string state (with escape handling), so braces/brackets/commas inside a quoted value are content — the killer case was a string containing `}}`, which zeroed the depth counter and split the spec mid-JSON. The regression test pins exactly that case (verified red against the old splitter). The `config:@<file>`-with-comma limit stays as you suggested and is now documented on `--arms`.

@bingran-you — with @Galius5136's independent review, live re-runs and a reproduced oracle on record, both PRs are ready when you are: #1045 (RFC) and this one. Happy to squash the stack or split it per workstream if either reviews easier, and CI just needs a maintainer's approval to run on a first-time contributor's PRs.

JeremyJC67:
@bingran-you quick unblock request: all four CI workflows on this PR (and on #1045) are queued in `action_required` — as a first-time contributor my runs need a maintainer to click **"Approve and run"** under the Checks tab before anything executes. Everything is already green locally and on a real daemon (full suite 5998 passed, docker live suite 5 passed, plus @Galius5136's independent re-run and "ready to merge" above), so CI should just confirm it. One click on each PR is all that's needed — thanks!

bingran-you:
> @bingran-you quick unblock request: all four CI workflows on this PR (and on #1045) are queued in `action_required` — as a first-time contributor my runs need a maintainer to click **"Approve and run"** under the Checks tab before anything executes. Everything is already green locally and on a real daemon (full suite 5998 passed, docker live suite 5 passed, plus @Galius5136's independent re-run and "ready to merge" above), so CI should just confirm it. One click on each PR is all that's needed — thanks!

All approved, thanks!

JeremyJC67:
CI attribution, now that the runs are in:

- **Green where it matters**: `test` (the full suite) passed, `pip-audit` passed, `fixture-scenarios` passed; the remaining jobs were scope-skipped.
- **The one red — `manifest-parity` — is pre-existing upstream drift, not from this PR**: the check compares core's `registry.AGENTS` against the manifests in `benchflow-ai/agents`, and the diff is exactly the codex-acp pin (`1.6.0` in core since #1044 vs `0.0.45` still in the agents repo). The same workflow has failed on `main` itself on its last three runs (`62cc7e41` — the #1044 bump — then `656ec915`, `fbababf6`). Reconciling the agents-repo manifest to `1.6.0` fixes it tree-wide; happy to send that one-liner to `benchflow-ai/agents` if useful.
- #1045 (the RFC, docs-only) still has its workflows queued as `action_required` — one more "Approve and run" click when someone has a moment.

JeremyJC67:
The `manifest-parity` fix is up as benchflow-ai/agents#66 (one-field reconciliation of the codex-acp pin with #1044) — once that merges, the last red here and on `main` goes green tree-wide.

JeremyJC67:
@bingran-you progress summary, so everything is in one place before Friday:

**Status: complete and green.**

- **#1045 (RFC)** — design only. Its workflows are still queued as `action_required`; one "Approve and run" click is all it needs.
- **#1046 (implementation)** — CI ran and is substantively green: `test` (full suite) ✅, `pip-audit` ✅, `fixture-scenarios` ✅, `detect-scope` ✅. The single red, `manifest-parity`, was **pre-existing upstream drift, not from this PR** — it has failed on `main` itself since #1044 (runs `62cc7e41`, `656ec915`, `fbababf6`) because the agents-repo manifest still pinned codex-acp `0.0.45` against core's `1.6.0`. I've sent the one-field fix as **benchflow-ai/agents#66**; merging that turns this check green here *and* on `main`.
- **Independent review** — @Galius5136 ran the docker live suite and real ablations on his own machine, re-ran after my fixes, independently reproduced the FastLap oracle (4.295858 MHz), and stated the PR is ready to merge with nothing outstanding.
- **Demo on your task** — `bench eval ablate` on `surface-ion-trap-shuttling`: **with-skill 1.00 vs no-skill 0.00**, with per-test rows that map onto the physics stages (skills flip the BEM electrostatics test and both invariant-transport waveforms; the Coulomb-equilibrium stage ties). Trajectories show the no-skill arm guessed 2.2 MHz and never attempted a field solve. That is the four-stage failure attribution from the Aug 13 sync, produced by the tool, on FrontierPhysics's own task.

**What I need from you:** the merge call on #1045/#1046 (happy to squash or split per workstream), plus the "Approve and run" click on #1045. Everything else is done.

Looking forward to Friday — I'll walk through the ablation table and how it can feed the paper's attribution story.

bingran-you:
## Exact-head review: request changes

Reviewed exact head `7d5e21e635cac347351bf399810a8385a2f2e80c` against base `ff10b7d70e6f9b52c87d17157882b6003e8548d2` using the thermo-nuclear code-quality criteria.

The core Docker path is real: composed snapshot/restore, fresh branch children, config deltas, lineage, reward artifacts, and provider telemetry all executed in an exact-head provider-backed smoke. However, the PR is not approval/merge ready and the advertised feature set is not complete end to end.

### Blocking findings

1. **[P1] Parent evidence can be deleted after a restore failure.** `MountedArtifacts.release()` catches `_move_entries()` failure, logs that evidence remains in the hold directory, then unconditionally deletes that directory with `rmtree` (`src/benchflow/branch_artifacts.py:198-216`). The hold and hand-off paths also fail open. Audit-critical artifact custody should fail closed and preserve recovery data.

2. **[P1] Replay cut-point provenance cannot detect the divergence it claims to detect.** `ReplayRouter.cut_point_digest` hashes the recorded request rather than the actual incoming request, while `_check_divergence` compares only message counts (`src/benchflow/continue_run/replay_proxy.py:129-160`). Same-count prompt/content/tool changes are invisible, and the RFC-required workspace digest is absent.

3. **[P1] `bench eval ablate` bypasses the canonical evaluation configuration.** `run_ablation()` manually creates a reduced `RolloutConfig` and hardcodes `skill_mode=no-skill` (`src/benchflow/ablate.py:1120-1143`). It drops normal controls/provenance including reasoning effort, task digest, dataset/source identity, prompts, agent environment, and required usage settings. The real E2E parent and child configs consequently had `task_digest: null` and `reasoning_effort: null`. Please resolve the canonical `EvalPlan` and overlay only the ablation axis.

4. **[P1] The report publishes a snapshot handle that cleanup has already destroyed.** `run_ablation()` calls cleanup before serializing `stage_snapshot` (`src/benchflow/ablate.py:1157-1179`). The real E2E report recorded `bf-snap-hello-world-task-34995d94bcff`; immediately afterward, `docker image inspect` confirmed it did not exist. RFC §3.6 promises `--keep-snapshots` plus `docker save`, but neither exists. Implement durable retention/export, or explicitly record the handle as ephemeral/deleted.

5. **[P1] Advertised stage/task integration is incomplete.** The PR description says stage-tagged replay cuts ship, but `bench eval continue` exposes only numeric `--max-exchanges`; `bench eval ablate` rejects `post-research`; normal execution does not wire `mark_stage()` to trajectory exchange indices; and task authoring still rejects `branch_execution: forked-snapshot` (`src/benchflow/task/prompts.py:237-249`). The expected four-stage FrontierPhysics failure-cascade workflow is therefore not available end to end.

6. **[P2] The implementation fails the requested structural gate.** New/expanded files include `ablate.py` at 1,203 lines and `rollout_branch.py` at 1,294 lines; `docker.py` is 1,104 and `rollout/__init__.py` is 2,918. `branch()` has cyclomatic complexity 21 and 75 statements, while `_run_children()` takes 14 arguments. Reuse the canonical evaluation planner and split snapshot policy, branch transaction, delta execution, and reporting into focused modules.

### Fresh verification at this head

- Targeted suite: **362 passed, 4 skipped**
- Full suite: **6,112 passed, 58 skipped**
- Live Docker suite: **5 passed**
- `ruff check .`: passed
- Provider-backed Docker ablation with a raw-verified Gemini key: parent plus two fresh children completed and emitted lineage, reward, and real provider-usage artifacts
- `ty check src/`: three errors, all reproduced on the exact base (not introduced here)
- GitHub currently has no checks/statuses attached to this exact head and reports the merge state as unstable

So: substantial working substrate, but **request changes** before approval. The critical acceptance gates are durable/auditable artifacts, truthful replay divergence evidence, canonical experiment configuration/provenance, and actual end-to-end stage/task wiring.

JeremyJC67:
@bingran-you Thank you for the exact-head review — the `docker image inspect` on the published snapshot handle was a check I had not done, and findings 4 and 5 were claim-vs-implementation gaps on my side, which is worse than bugs. All six are addressed at `f796f966`; the PR description now carries a per-finding table with commits, each pinned by tests verified red on the pre-fix tip.

Highlights:

- **P1-1** custody fails closed: typed `ArtifactCustodyError`, the hold directory is preserved on any failure, `rmtree` only after every entry is confirmed moved back — and a custody failure never masks a child's own exception.
- **P1-2** divergence is detected on content: served-vs-recorded request digests per exchange, recorded as `divergence_events` (with `strict_divergence` now tripping on content), and the RFC workspace digest is implemented — recorded with its basis, or `null` plus a reason, never fabricated.
- **P1-4** `--keep-snapshots` exports the stage image via `docker save` **before** cleanup (tar path + sha256 in the report); without the flag the handle is recorded `ephemeral: true, exported: null`. RFC §3.6 wording corrected.
- **P1-3** `run_ablation` now resolves the same per-task assembly `bench eval run` uses (extracted `task_rollout_config()`) and overlays only the ablation axis — parent and child configs carry non-null `task_digest`, `reasoning_effort`, source identity, prompts and usage settings, pinned by tests.
- **P2** the engine is split along the seams you named: `branch_policy` / `branch_transaction` / `branch_children` / `branch_report`; `branch()` is ~15 cc / 41 statements, the 14-argument runner takes one params object. The refactor is behavior-neutral: identical full-suite results before and after, zero test edits.
- **P1-5** the four-stage workflow is wired end to end: stage captures record completed-LLM-exchange indices; `bench eval continue --cut-stage <name>` resolves them from the run folder; `bench eval ablate --mark-research-end-on PLAN.md` captures `post-research` the moment the marker file appears and branches there; and `branch_execution: forked-snapshot` now validates and compiles to a stage-capture request.

Gates at tip `f796f966`: targeted `371 passed, 4 skipped`; full suite `6156 passed, 0 failed`; `ruff check` / `ruff format --check` / `ty check src/` clean; **docker live suite `5 passed` on a real daemon at this exact head**. One note for the record: the three `ty` errors you saw reproduce neither on my base nor tip with this environment's ty version — consistent with your note that they predate the PR, I left that area untouched.

Ready for another exact-head pass whenever you are.

bingran-you:
## Updated exact-head re-review: request changes

Reviewed exact head `f796f966555801074eb3400e0b74b0b793f08522` against base `ff10b7d70e6f9b52c87d17157882b6003e8548d2`.

This update materially fixes the earlier artifact-custody, canonical-config, request-digest, and ablation snapshot-retention defects. The provider-backed Docker workflow now completes end to end, including export, checksum verification, image removal, reload, and execution. However, I still found two P1 auditability defects and two P2 readiness issues.

### [P1] Workspace digests silently collide for legal filenames

`src/benchflow/sandbox/workspace_digest.py:34-42` uses newline-separated `find | sort | xargs` pipelines for `sha256sum` and `stat`. Filenames containing spaces are split into multiple arguments. The inner commands fail, but the outer pipeline still exits successfully.

I reproduced this with two different Alpine workspaces containing `file with spaces.txt` (0 bytes versus 37 bytes). Both emitted inner-command errors and returned the same successful digest:

`9ca299f5052a19a773879dd8ebd51e1c7790759b50c685051d275ecfef842d5e`

That allows materially different workspaces to be recorded as identical, so replay divergence can remain invisible. Please use a NUL-delimited traversal (`find -print0 | sort -z | xargs -0` or equivalent), include symlink targets in the digest contract, and make any pipeline-stage failure fatal.

### [P1] Normal task-declared snapshots are persisted as usable refs after cleanup has deleted them

Task config enables `branch_execution: forked-snapshot` in `src/benchflow/rollout/_config.py:334-363`, and `src/benchflow/branch_lineage.py:138-158` serializes image refs without any lifetime/ephemeral marker. Normal rollout cleanup then removes those images in `src/benchflow/rollout/__init__.py:2221-2229`.

On a plain `Rollout.run()` Docker probe with sandbox snapshots at `env-ready`, `pre-verify`, and `post-verify`, `stage_snapshots.json` recorded three valid-looking `bf-snap-...` image refs. Immediately after the run, `docker image inspect` failed for all three. The artifact contains no `ephemeral` or export marker.

`--keep-snapshots` is only available on `bench eval ablate`; normal `bench eval run` and the task declaration have no retention/import path. Therefore the documented later `branch_at_stage` workflow is not available from a completed plain evaluation, while its artifacts appear restorable. Please either add retention/import support to normal evaluation or mark refs ephemeral during cleanup and narrow the documented/task-level contract.

### [P2] Unsupported experiment-wide reasoning settings fail after provisioning and are retried in a child

With Gemini and `--reasoning-effort high`, the run built the environment, created the snapshot, installed and connected the agent, and only then rejected the unsupported ACP effort option. Because the env-ready snapshot existed, ablation restored a child, installed the agent again, and failed the same way before skipping the remaining branch.

This is a global request/agent compatibility error, not task-attributable post-boundary failure. Validate it before parent provisioning and do not retry it in branch children.

### [P2] The explicit structural gate still fails

`src/benchflow/ablate.py` is now 1,035 lines. `run_ablation()` at line 881 still has C901 complexity 13, PLR0912 15 branches, and PLR0915 59 statements. `rollout_branch.py` is much better at 441 lines, but the thermo-nuclear >1,000-line gate remains unresolved for `ablate.py`.

### Documentation and merge readiness

- `docs/architecture.md:102` and `:108` still describe the branch engine as environment-snapshot-only and container composition as future work.
- The PR body's “Known limitations” still says `environment_ref` and `config_override` fail closed and children emit only provenance plus reward, contradicting the current code and later updated section.
- The exact head currently has no GitHub checks/statuses and reports `UNSTABLE`.

### Verification at this head

- Focused tests: **548 passed, 4 skipped, 5 deselected**
- Full suite: **6,164 passed, 86 skipped, 12 deselected**
- Live Docker tests: **5 passed**
- `ruff check .`, changed-file formatting, and `ty check src/`: passed
- Raw Gemini 2.5 Flash-Lite request: HTTP 200
- Provider-backed ablation: parent reward 0, child rewards 0 and 1, no errors, ablation value 0.5, with real provider usage
- Snapshot custody path: recorded checksum matched; the image was absent after cleanup; `docker load` restored the exact image ID; the restored image executed successfully

So the implementation is substantially closer, but I would not merge it until the two P1 provenance/lifetime issues are fixed and the late global-validation path is corrected.

JeremyJC67:
@bingran-you All four findings from the re-review are addressed at `8fc8404f`, plus the documentation drift. The PR description carries the per-finding table; highlights:

- **Digest collision** — your repro is now a red-first test (two trees differing inside `file with spaces.txt` produced your exact colliding digest on the old pipeline; they must differ now, and a newline filename must either digest correctly or fail closed). The pipeline is null-safe end to end and every stage's exit status is checked — verified against busybox ash, since `pipefail` isn't dependable there. The docker-test helper now delegates to the same fixed src helper.
- **Snapshot lifetime on plain runs** — cleanup rewrites `stage_snapshots.json` so every ref is marked `ephemeral: true` unless exported; `bench eval run --keep-snapshots` now exists (same export machinery as ablate), and the workflow you called unavailable is now real and tested: `bench eval import-snapshots <run-dir>` verifies the recorded sha256, `docker load`s the tar, and fails closed unless the loaded image id matches the recorded one. The docker live suite grew a proof of exactly that round trip.
- **Late request-global failures** — unsupported reasoning effort is rejected in `build_eval_plan`, before any provisioning; deterministic ACP option rejections classify as `request_global` (following the #917 non-retryable pattern) and are excluded from retries; and a parent that fails request-globally skips all branch children instead of re-failing the same config per arm. Timeouts deliberately stay retryable — a timeout proves nothing about compatibility.
- **Structure** — `ablate.py` is 674 lines with arm parsing/validation in `ablate_arms.py` (481); `run_ablation` passes your measurement (`ruff --select C901,PLR0912,PLR0915`) clean. Behavior-neutral: identical suite results before and after.
- **Docs** — `architecture.md` no longer describes container composition as future work (agent-session snapshotting still is); the RFC §3.6 and the PR's Known-limitations section now match the code.

Gates at tip: full suite `6190 passed, 0 failed`; docker live `6 passed`; `ruff`/format/`ty` clean.

Ready for another pass whenever you are — and thank you for the depth of these reviews; the digest repro and the `docker image inspect` checks caught things none of my harnesses did.

JeremyJC67:
@bingran-you rebased onto current `main` (`f309f4f8`, i.e. after #1062/#1063/#1064/#1066) and re-verified end to end — this branch now merges without friction.

The rebase had exactly one conflict, in `acp/runtime.py`: #1066 and this branch each added an import into the same slot. Resolved by keeping both in isort order. I checked mechanically rather than by eye — diffed the resolved file against upstream's (the only removals are the two `raise RuntimeError` statements this branch deliberately converts to `ACPRequestGlobalError`; #1066's watchdog refactor is untouched), and diffed our whole pre-rebase delta against the post-rebase one across all 71 files: zero behavioural drift. The other overlapping files auto-merged coherently — #1064's `--trials` guard and this branch's reasoning-effort preflight both live in `build_eval_plan`, and #1063's `_get_completed_tasks` rewrite is untouched by our `task_rollout_config()` extraction.

Gates at `72cb8cd9`:

```
targeted   503 passed, 4 skipped
full       6227 passed, 94 skipped, 0 failed
ruff check / ruff format --check / ty check src/    clean
docker live (real daemon)   6 passed
```

Everything from your second review is in (custody fails closed, digests null-safe and fail-closed, request-global settings validated before provisioning and never retried in children, `ablate.py` 674 lines with `run_ablation` passing your `C901,PLR0912,PLR0915` measurement, architecture/RFC text matching the code). @Galius5136 independently re-ran the live suite and reproduced the FastLap oracle earlier, so the evidence isn't only mine.

Ready for a merge call whenever you have a moment — and #1045 (docs-only) has never had its CI run; it needs one "Approve and run" click.

Also, for planning: does the 7 September author-list line apply to infra tickets the same way it does to task PRs?

JeremyJC67:
@bingran-you pushed a rebase onto current `main` (`10ae40be`, i.e. after the 0.7.6 release, #1054, #1113 and #1118). This PR shows mergeable again.

Three things changed on my side, all forced by the rebase:

1. **One of my commits is gone.** `fix(acp): skip doomed set_model when codex launch config owns an off-catalog model` is superseded by #1076, so I dropped it rather than merge it. I checked rather than assumed — I replayed its four tests against #1076's implementation; three fail, and they fail because they construct an env state the wiring never produces (#1076's `owns_model` requires `BENCHFLOW_PROVIDER_MODEL == safe_model_alias(model)` and `BENCHFLOW_LITELLM_MODEL_VIA_ENV`, and `_wire_litellm_agent_env` sets both unconditionally for codex-acp alongside the CODEX_CONFIG injection). So #1076 covers the live codex-acp 1.6.0 failure I originally hit.

2. **#1113 added `disallow_hosted_search` to the agent_launch plane.** You updated the doubles you own; this stack's own double still had the two-argument signature, so 26 branch-delta tests died with `TypeError`. Fixed with the same shape as yours — keyword-only, `False` default.

3. **A CHANGELOG entry.** This stack had none anywhere: nothing under Unreleased, nothing in 0.7.6. Three Added entries and one Fixed are now under Unreleased.

Plus one assertion of mine that matched a phrase Rich had line-wrapped around an interpolated `tmp_path`; it now matches whitespace-collapsed output.

Gates at `a2842ce6`, on Linux with a real Docker daemon:

```
targeted   484 passed, 4 skipped
full       6502 passed, 86 skipped, 3 failed
ruff check / ruff format --check src tests tools / ty check src/    clean
```

The three failures are pre-existing on `main` — I confirmed them in a worktree at untouched `upstream/main`: `test_example_judge_fails_closed_on_infra_error`, `test_missing_credentials_anthropic_default_no_creds`, `test_skip_reason_credential_gate_after_docker_ok`.

On the substance, in case it is useful for the decision: the demo is an ablation on `surface-ion-trap-shuttling` (codex-acp / gpt-5.4-mini, real docker) — with-skill 1.00 vs no-skill 0.00, and the tool names *which* sub-tests moved: BEM radial frequency and both transport waveforms flip, Coulomb equilibrium ties. no-skill guessed 2.2 MHz off trap geometry and never attempted a field solve; with-skill read the mentor pack, ran FastLap and landed 4.295858 MHz — the number @Galius5136 reproduced independently when he re-ran the suite. That per-sub-test attribution is the part a scalar reward cannot give you.

Ready for a merge call whenever you have a moment — or tell me how you would like it split and I will do the splitting. Both this PR and #1045 need an "Approve and run" click before CI will report anything.

Issue #933, opened 2026-07-21
ACP token usage is lost on prompt cancel/timeout — timed-out runs report total_tokens=0


EYH0602:
# Title: ACP token usage is lost on prompt cancel/timeout — timed-out runs report `total_tokens=0`

## Summary

For ACP agents, per-trial token usage is captured **only** from the return value of
`session/prompt`. When a prompt is cancelled or hits a hard/idle time cap, that request
never returns a `PromptResult`, so usage is never recorded and the trial reports
`total_tokens=0` with `usage_source=unavailable` — even though the agent demonstrably
consumed tokens (a completed run of the same task reports millions).

This is not "imprecise" telemetry; it is **zero** telemetry on the exact runs (timeouts)
where a token count matters most for budget accounting.

## Version

`benchflow==0.6.5` (latest release as of filing).

## Root cause

Usage is bound to prompt *completion*:

- `ACPClient.prompt()` awaits `session/prompt`, then records usage only after it returns:
  ```python
  # benchflow/acp/client.py:407
  prompt_result = PromptResult.model_validate(result)
  self._session.stop_reason = StopReason(prompt_result.stop_reason)
  self._session.record_prompt_usage(getattr(prompt_result, "usage", None))
  ```
- On timeout the run goes through `ACPClient.cancel()` (`acp/client.py:410`), which sends
  `session/cancel` but never reaches `record_prompt_usage`.
- `ACPSession.handle_update()` — the `session/update` streaming handler
  (`acp/session.py:273+`) — does **not** inspect usage/tokens, so no incremental snapshot
  exists to fall back on.
- Result: `usage_snapshots` stays empty → `latest_usage_totals()` returns `None`
  (`acp/session.py:242`) → the trial emits `usage_unavailable()`
  (`usage_tracking.py:70`): `total_tokens=0`, `usage_source=unavailable`.

## Reproduction

1. Run any ACP agent (e.g. `claude-agent-acp`) on a task with a hard/idle time cap short
   enough that the agent is still mid-prompt at cutoff.
2. The trial is cancelled at the cap.
3. Observe the result metrics: `total_tokens=0`, `usage_source=unavailable`, despite a
   completed run of the same task reporting real token counts.

## Impact

- Any budget/cost accounting that depends on ACP token counts is unusable for timed-out
  trials — and timeouts are common in long-running / agentic tasks.
- Multi-turn ACP sessions partially mitigate (each *completed* `prompt()` leaves a
[...]

PR #1051, opened 2026-08-23
Preserve native ACP artifacts at prompt timeout

kywch:
## Summary
- send ACP `session/cancel` before locally cancelling timed-out prompt tasks
- wait up to 5 seconds for terminal tool updates and final `PromptResponse.usage`
- keep hard-cancel fallback bounded and mark unresolved tool calls partial
- apply same cancellation sequence to wall-clock and idle timeouts

## Why
Native subscription usage arrives on final ACP prompt response. Immediate local task cancellation discarded that response, left tool calls pending, and made otherwise normal wall-clock timeouts artifact-incomplete.

## Verification
- `uv run pytest -q tests/test_acp.py tests/test_native_acp_usage.py tests/test_trajectory_streaming.py tests/test_session_factory_runtime.py` (136 passed)
- `uv run ruff check src/benchflow/acp/runtime.py tests/test_acp.py tests/test_native_acp_usage.py`
- `uv run ty check src/benchflow/acp/runtime.py`

No live experiment rerun performed.

---

<a href="https://app.devin.ai/review/benchflow-ai/benchflow/pull/1051" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/benchflow/acp/client.py
+++ b/src/benchflow/acp/client.py
@@ -1,5 +1,6 @@
 """ACP client — benchflow acts as the client, agents are ACP servers."""
 
+import asyncio
 import logging
 from collections.abc import Awaitable, Callable
 from typing import Any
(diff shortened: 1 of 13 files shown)


Galius5136:
Thanks! I tested this A/B against the merge-base and also ran a normal Docker oracle rollout. The core wall-clock timeout terminalization works, and the normal path is unaffected.

I found a few issues that I think need to be addressed before merge:

1. **`sandbox_user=None` loses both reward and trajectory on timeout.**  
   On base, the verifier still returns `reward=1.0` and the partial trajectory is published. On this branch the verifier does not run, `rewards=None`, and the trajectory is not published at all. Since `sandbox_user=None` is a documented configuration, this looks like an unnecessary loss of evidence rather than just avoiding an untrusted terminal claim.

2. **Non-cooperative timeouts now pay the full 60s grace period.**  
   With default settings I measured:
   - wall-clock timeout: `1.01s → 61.28s`
   - idle timeout: `2.03s → 62.15s`

   `session/cancel` is sent correctly, but a hung prompt task keeps the full grace alive. This can become a significant cost across many timed-out rollouts.

3. **Idle timeouts are cancelled but not terminalized.**  
   I still get `status=pending`, no `effects`, no `terminal_source`, and `terminal_trajectory_complete=false` on the idle-timeout path. So the state this PR is meant to eliminate still survives there.

4. **The new sandbox-process hard gate can discard the verifier result outside timeout handling.**  
   If a sandbox-user process survives the kill step, base still returns the verifier reward, while this branch returns `rewards=None` with `Verifier hardening failed`. Since this runs for every rollout with a sandbox user, it is a broader behavior change than the timeout feature itself.

For control, a normal Docker oracle rollout on this branch still completes `1/1`, `mean reward 1.00`, so the happy path looks good.

I’d like the failure-path behavior above settled before approving this.

kywch:
Underlying ACP timeout bug is valid, and this prototype plus A/B review usefully exposed several contract gaps. However, current diff mixes bounded ACP cancellation, lifecycle synthesis, sandbox-process hardening, and verifier policy, with confirmed regressions in grace-period duration, idle-timeout terminalization, and verifier/evidence behavior.

Follow-up work is now split by concern:

- #933 — preserve ACP usage and cancellation evidence on timeout
- #1078 — require a checked post-kill quiescence observation before verifier startup
- #1079 — define BenchGuard artifact-evidence contracts

Given that split, I plan to close this PR as superseded rather than merge it as-is.

I will retain head branch `fix/graceful-acp-timeout` at `f922c745` temporarily for reproducibility and existing BenchGuard work. Closing changes review state only; it does not endorse current prototype behavior as supported or merge-ready.

kywch:
Thank you @Galius5136 for the review! I'm closing this issue as this cannot be merged as is. Please see the related issues that extract the core problems, so that those can be addressed properly.

PR #1062, opened 2026-08-29
fix(sandbox): use the verifier-setup budget for hardening execs

### LINKED ISSUE #1058: [BUG] Verifier hardening execs keep a 10s timeout on the scoring path and record rollouts as verifier crashed
## Required prerequisites

- [x] I have searched the Issue Tracker and this has not been reported before.
- [x] I have verified the problem against the latest `main` (`ff10b7d`).

## What version of BenchFlow are you using?

0.7.6.dev2102 (PyPI); code paths verified identical on `main` at `ff10b7d`.

## System information

Host: macOS 15 (arm64), Python 3.12.13. Sandbox: Daytona. Agent: codex-acp 1.6.0 with azure-foundry-openai models. Batch shape: 16 tasks x 3 models, one `bench eval run` per model in parallel, `--concurrency 16` each.

## Problem description

`harden_before_verify` runs on the scoring path, and most of its steps use the shared `VERIFIER_SETUP_TIMEOUT_SEC` budget (180s). Three helpers on the same path still hardcode `timeout_sec=10`:

- `_trusted_verifier_path`: `src/benchflow/sandbox/lockdown.py:712`, `:719`
- `_trusted_verifier_pythonpath`: `src/benchflow/sandbox/lockdown.py:751`, `:758`
- `_kill_sandbox_user_procs`: `src/benchflow/sandbox/lockdown.py:1057`

On Daytona under load, session-command dispatch latency alone can exceed 10s even for `printenv PATH`. The RuntimeError propagates out of the verifier setup and the rollout is recorded as a verifier crash with `rewards: null`. The agent phase (often hours of work) is scored as nothing.

## Reproducible example code

```bash
bench eval run --tasks-dir <fp-tasks> --agent codex-acp \
  --model azure-foundry-openai/<deployment> --sandbox daytona --concurrency 16
```

Reproduces probabilistically under load; the failing exec is a trivial command, so the timeout is session latency, not the command itself.

## Traceback

```
verifier crashed: Command timed out after 10 seconds
```

Recorded in `result.json` as `verifier_error`; hit 6 times in one 144-rollout batch across 4 different task types.

## Expected behavior

Hardening execs on the scoring path use the same `VERIFIER_SETUP_TIMEOUT_SEC` budget the surrounding steps already use, so a slow session cannot void a finished rollout.

## Additional context

Fix proposed in #1062.

Benjamin-eecs:
## Description

Apply the canonical `VERIFIER_SETUP_TIMEOUT_SEC` budget to every normal verifier hardening/setup probe: trusted `PATH` and `PYTHONPATH`, pytest-plugin discovery, distro detection, both sandbox-user process-kill passes, and the existing clear/prepare/freeze/cleanup commands.

## Motivation and Context

These helpers run once per rollout on the scoring path. Daytona dispatch latency can exceed the previous 5/10/15-second literals even for trivial commands; some helpers then silently fell back to an incomplete verifier environment, while another process-kill pass could wait for the transport hard cap. One owned 180-second ceiling avoids both failure modes without adding retries or changing normal-path latency.

Closes #1058.

- [x] I have raised an issue to propose this change (#1058)

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## Implemented Tasks

- [x] Use the shared budget for trusted environment discovery
- [x] Use it for plugin and distro probes that previously failed soft after 15s/5s
- [x] Bound both sandbox-user process-kill passes
- [x] Add a PR-named regression test that audits the complete default hardening sequence

## Validation

- [x] `tests/test_sandbox_hardening.py`: 100 passed
- [x] `tests/`: 5933 passed, 48 skipped, 7 deselected
- [x] Ruff format/lint and `ty check src/`
- [x] Real `gemini-3.1-pro-preview` Docker and Daytona E2E: reward 1.0 on both backends
- [x] Artifact validator: Docker 1/1 healthy and Daytona 1/1 healthy, with complete ACP + provider trajectories and training-ready rows

### DIFF
--- a/src/benchflow/sandbox/lockdown.py
+++ b/src/benchflow/sandbox/lockdown.py
@@ -709,14 +709,16 @@ async def _trusted_verifier_path(
     checks prove they are root-owned directories and not group/world writable.
     Runtime locations and sandbox-user writable locations stay excluded.
     """
-    path_result = await env.exec("printenv PATH", user="root", timeout_sec=10)
+    path_result = await env.exec(
+        "printenv PATH", user="root", timeout_sec=VERIFIER_SETUP_TIMEOUT_SEC
+    )
     raw_path = path_result.stdout or ""
     if not raw_path.strip():
         return _SAFE_VERIFIER_PATH
     cmd = _trusted_path_extras_cmd(
         raw_path, _blocked_verifier_path_prefixes(sandbox_user, workspace)
     )
-    result = await env.exec(cmd, user="root", timeout_sec=10)
+    result = await env.exec(cmd, user="root", timeout_sec=VERIFIER_SETUP_TIMEOUT_SEC)
     if _exec_return_code(result) != 0:
         logger.debug(
             "Trusted verifier PATH extras unavailable; using safe PATH.%s",
@@ -748,14 +750,16 @@ async def _trusted_verifier_pythonpath(
     is chowned to root before verification.
     """
     pp_result = await env.exec(
-        "printenv PYTHONPATH 2>/dev/null || true", user="root", timeout_sec=10
+        "printenv PYTHONPATH 2>/dev/null || true",
+        user="root",
+        timeout_sec=VERIFIER_SETUP_TIMEOUT_SEC,
     )
     raw_pp = (pp_result.stdout or "").strip()
     if not raw_pp:
         return ""
     blocked = _blocked_verifier_pythonpath_prefixes(sandbox_user)
     cmd = _trusted_path_extras_cmd(raw_pp, blocked)
-    result = await env.exec(cmd, user="root", timeout_sec=10)
+    result = await env.exec(cmd, user="root", timeout_sec=VERIFIER_SETUP_TIMEOUT_SEC)
     try:
         extras = _json.loads(result.stdout or "[]")
     except _json.JSONDecodeError:
@@ -1052,7 +1056,7 @@ async def _kill_sandbox_user_procs(env, sandbox_user: str) -> None:
     await env.exec(
         f"pkill -u {sandbox_user} 2>/dev/null; "
         f"sleep 1; pkill -9 -u {sandbox_user} 2>/dev/null || true",
-        timeout_sec=10,
+        timeout_sec=VERIFIER_SETUP_TIMEOUT_SEC,
     )
     # Second pass: catch any processes that slipped through (e.g. cron/at jobs).
     await env.exec(
--- a/src/benchflow/sandbox/lockdown.py
+++ b/src/benchflow/sandbox/lockdown.py
@@ -609,7 +609,7 @@ async def _discover_pytest_plugin_flags(env, task: "Task") -> str:
         result = await env.exec(
             f"python3 -c {shlex.quote(_DISCOVER_PYTEST_PLUGINS_SCRIPT)}",
             user="root",
-            timeout_sec=15,
+            timeout_sec=VERIFIER_SETUP_TIMEOUT_SEC,
         )
         if result.stderr:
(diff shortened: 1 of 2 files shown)


JeremyJC67:
This fixes the right thing in the right place: after this change no 10-second literal remains in lockdown.py (main had exactly five, all on the harden-before-verify path, all converted here), and reusing `VERIFIER_SETUP_TIMEOUT_SEC` keeps one owned budget instead of adding a knob. `tests/test_sandbox_hardening.py` passes for me too (99 passed), and ruff is clean.

1. **The same Daytona latency now fails soft — and silently — two calls later.** `_build_verifier_env` runs on the identical scoring path and still contains `_discover_pytest_plugin_flags` at `timeout_sec=15` (lockdown.py:607-612) and `_distro_pip_env` at `timeout_sec=5` (lockdown.py:685-688). Both swallow the timeout and fall back: plugin discovery degrades to task-declared plugins while `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1` stays pinned, and distro detection returns `{}`, dropping `PIP_PREFIX` on Fedora-likes. Under the >10s dispatch latency you reported in #1058 these don't crash the rollout — they quietly change the verifier env and can turn a genuine pass into a recorded failure, which is arguably worse than `rewards: null` because it lands as real signal. Suggest raising both to `VERIFIER_SETUP_TIMEOUT_SEC` in this PR; the fail-soft handling can stay for genuine command failures.

2. **The second pass of `_kill_sandbox_user_procs` has no budget at all** (lockdown.py:1061-1066). On Daytona that means the `_DAYTONA_EXEC_HARD_CAP_SEC = 3600` safety net — a wedged pgrep/pkill burns up to an hour on the path you just budgeted at 180s one statement above. `timeout_sec=VERIFIER_SETUP_TIMEOUT_SEC` there makes the helper internally consistent; fine as a follow-up if you'd rather keep this PR literal-for-constant only.

3. **Nothing pins the new budgets.** `test_pr_942_tree_hardening_uses_shared_setup_budget` (tests/test_sandbox_hardening.py:237-269) already asserts per-command `timeout_sec == VERIFIER_SETUP_TIMEOUT_SEC`, but runs with `sandbox_user=None` and doesn't cover `printenv PATH` / `printenv PYTHONPATH` / the extras commands / the pkill first pass. Extending `expected_commands` (plus one variant with a sandbox_user set) is ~10 lines and gives this PR a test that would fail red on main — right now it has none.

4. Wall-clock, since "budget + retry" came up: the budget is a ceiling, not a wait — the pkill sequence still completes in ~1s — so the normal path is unchanged. Worst case, the five converted execs add ~14 min to a harden phase that already carried a ~18 min ceiling across its other 180s execs, and which runs outside the verifier's own `asyncio.wait_for` (rollout/_setup.py:472-480). I think budget-only is the correct minimal fix: all five commands are idempotent, so if 180s ever proves insufficient in the field a retry wrapper is safe to add then — today it would be speculative.

No interaction with my #1046 (feat/ablate-cli): I don't touch lockdown.py and a trial merge in both directions is clean, so landing order between these two is free — happy to rebase whichever lands second.

Benjamin-eecs:
Thanks for the thorough verification.

bingran-you (approved):
Validated at exact head 63a2ea9f. All verifier-environment and hardening probes now use the owned setup budget, including plugin discovery, distro detection, and both process-kill passes; the 30-second best-effort cache reclaim remains intentionally separate. Exact-head focused tests are 100/100 and every required GitHub check is green. Full local suite, Ruff, ty, plus real-key gemini-3.1-pro-preview Docker and Daytona runs passed at reward 1.0; both trajectory artifacts validated healthy and training-ready. Ready to squash merge.

PR #1112, opened 2026-09-08
feat: add private research controls on top of URL denylist

ElegantLin:
## Summary

Extends the generic URL/host denylist merged in #1113 with the private,
run-scoped research controls needed to address the discovery/privacy part of
benchflow-ai/FrontierPhysics#365. This PR is an additive research-benchmark
layer, not a competing replacement for `network_mode: denylist`.

- add `bench eval run --research-policy /secure/policy.yaml` plus SDK, YAML-config, matrix, and worker-sharding propagation
- resolve a private per-task deny policy for URLs, URL prefixes, hosts, case-insensitive terms, and exact content SHA-256 values
- inject one `benchflow-research` MCP server (`web_search`, `web_fetch`, `web_download`) across ACP and native-MCP-config harnesses
- run the policy engine as a root-owned loopback gateway and block all other sandbox-user IPv4/IPv6 egress before the agent process starts
- disable harness-native web tools and strip provider-side search/URL-context controls at the loopback LiteLLM proxy to close confused-deputy bypasses
- support native Claude subscription auth through a fixed-destination loopback relay to `api.anthropic.com`; reject Anthropic server-side web tools and remote MCP requests before forwarding
- honor explicit OpenAI-compatible endpoints for bare/future model IDs and propagate Codex reasoning effort into the LiteLLM deployment route
- use DuckDuckGo Lite for the default search backend and parse its result links; keep the sandbox-local LiteLLM bootstrap portable on Docker Desktop ARM64
- persist only policy hash/count/enforcement metadata; redact the private path from durable worker/run artifacts
- fail closed for missing task entries, root agents, external pre-started sandboxes, and non-Docker providers

## Relationship to #1113

#1113 provides a task-declared network denylist for ordinary HTTP clients: it
prevents the agent from retrieving configured URL prefixes or hosts while
leaving other network access available.

This PR covers the stricter research-evaluation case where the target list must
remain private and the agent should not discover the target paper while
searching:

- the policy is supplied outside the task package and is unreadable by the agent
- search results are filtered by blocked URLs and case-insensitive terms before the agent sees them
- fetched/downloaded content can also be rejected by exact SHA-256, covering known byte-identical mirrors
- native and provider-side web tools are disabled or rejected so they cannot bypass the filtered research gateway
- Claude subscription traffic uses a fixed-destination provider relay, while all other direct sandbox-user egress remains blocked

When `--research-policy` is absent, #1113's normal `network_mode: denylist`
behavior is unchanged. When it is present, the stricter private research path
takes precedence instead of starting both proxy implementations.

## Policy boundary

This guarantees that the sandbox user cannot directly retrieve listed resources or bypass the gateway with curl/raw sockets. It cannot erase knowledge already present in model pretraining or hide unlisted mirrors/citations; operators can add blocked terms, content hashes, and known mirror URLs when discovery leakage matters.

## Validation

- `uv run pytest -q` after rebasing onto `f396c355` — **6146 passed, 86 skipped, 8 deselected** in 12m18s
- `uv run pytest tests/test_research_policy.py -m integration -q` after the rebase — 1 passed, 32 deselected in 41.80s
- the post-rebase cross-feature suite covering research policy, the new upstream denylist, hosted-search controls, LiteLLM routing, and smoke wiring — 307 passed, 1 deselected
- `uv run pytest tests/test_source_adapters.py -q` — 21 passed; the previously observed full-suite startup timeout also passed in four isolated repetitions and in the final full suite
- targeted `ruff format --check`, targeted `ruff check`, and full `uv run ty check` — passed
- live `claude-agent-acp` / `claude-sonnet-4-6` / Docker / subscription-auth rollout — reward 1.0; blocked MCP URL rejected, allowed MCP URL fetched, direct curl failed with exit 7, and an Anthropic server-side web-search request was rejected with HTTP 403
- live `claude-agent-acp` / `claude-haiku-4-5-20251001` / Docker rollout exercised all three MCP tools — reward 1.0; `web_fetch` and `web_download` succeeded for an allowed URL, a blocked fetch was rejected, the download was verified, and direct curl failed with exit 7
- a second live Claude Haiku rollout after switching to DuckDuckGo Lite returned 10 real `web_search` results — reward 1.0
- artifact scans for both Claude live runs found zero secret or private-policy leakage matches
- live `codex-acp` was launched with recorded `model: gpt-5.6-sol` and `reasoning_effort: xhigh`; sandbox-local LiteLLM startup, the UID firewall, the explicit provider route, and the outbound provider request were all exercised, and artifact scanning found zero API-key matches

The Docker canary also proves the private root policy is unreadable by the agent and deterministically checks that a correctly formed Anthropic provider-side web-search request returns HTTP 403.

The Codex live run is **not counted as a successful generation**: the configured provider authenticated but returned `model_not_found` because no account in that provider group supports `gpt-5.6-sol`. The requested model and `xhigh` effort were kept unchanged rather than silently substituting a different model. BenchFlow now records those two fields separately and routes `xhigh` into the LiteLLM deployment configuration.

---

<a href="https://app.devin.ai/review/benchflow-ai/benchflow/pull/1112" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


ElegantLin:
Claude live-validation follow-up (`b6bdc857`):

The first real `claude-agent-acp` + subscription-auth + research-policy rollout exposed a genuine gap: native Claude auth skips LiteLLM, so the UID firewall had no loopback provider endpoint and failed closed before the prompt.

The follow-up routes native Anthropic traffic through a root-owned, fixed-destination loopback relay to `api.anthropic.com`. The relay is not a general proxy and rejects provider-side web-search/web-fetch tools, remote MCP requests, compressed/ambiguous message bodies, and duplicate JSON keys before forwarding.

Validation on the pushed tree:

- Claude Sonnet 4.6 live Docker rollout: reward `1.0`, 6 tool calls; blocked MCP URL rejected, allowed MCP fetch succeeded, direct curl failed with exit 7, and an Anthropic `web_search_20250305` request returned HTTP 403.
- Claude Haiku 4.5 live Docker rollout: reward `1.0`, 6 tool calls, no rollout/verifier error.
- Deterministic Docker canary: `1 passed in 30.05s`, including a correctly formed provider-side web-search bypass assertion.
- Focused regression suite: `59 passed, 1 deselected`; Ruff and `ty check` pass.
- The repository's existing Claude live smoke also passed when rerun alone (`1 passed in 113.56s`).

One Sonnet attempt hit a transient ACP `-32603` during initialization and succeeded on the automatic retry; no policy assertion was skipped.

ElegantLin:
Positioning update after #1113 merged:

- #1113 is now the generic, task-declared `network_mode: denylist` layer. It blocks retrieval through ordinary HTTP clients while preserving access to other destinations.
- This PR is rebased on #1113 and is intentionally additive. It covers the stricter research-benchmark case where the target policy must remain private and target-paper discovery must be filtered before results reach the agent.
- The additional scope here is the external run-scoped policy, filtered `web_search`/`web_fetch`/`web_download` MCP gateway, title/URL term filtering, exact content-hash rejection, provider-side web-tool bypass prevention, fixed-destination Claude subscription relay, and artifact redaction.
- Without `--research-policy`, #1113 behaves unchanged. With it, the stricter private research path takes precedence rather than starting both proxy implementations.

The PR title and description now state this relationship explicitly. Head `47cf8680` remains cleanly mergeable and all current checks pass; the documentation-only PR metadata update did not change the tested code.

PR #1117, opened 2026-09-09
fix(sandbox): reject pytest plugins backed by agent-writable code

### LINKED ISSUE #1116: [BUG] An agent can make a failing pytest task pass by planting a plugin in its workspace
## What happens

`_discover_pytest_plugin_flags` enumerates pytest11 entry points inside the sandbox and puts every name it finds on the verifier's `-p` allowlist. Its docstring says it "filters to only those whose dist-info is in a root-owned directory", but the script it runs asks nothing about where an entry point came from:

```python
# src/benchflow/sandbox/lockdown.py
eps = list(entry_points(group='pytest11'))
names = sorted(set(ep.name for ep in eps))
print(json.dumps(names))
```

Registering a pytest11 entry point needs neither pip nor root. `importlib.metadata` reads `*.dist-info` directories straight off `sys.path`, so creating one is an ordinary file write that the sandbox user is fully entitled to make in its own workspace.

The workspace is on `sys.path` because the verifier exec carries no `cwd` (`task/verifier_core.py`), so `test.sh` runs in the image WORKDIR — normally the agent's workspace. Any `python3 -m pytest` from there picks it up.

A plugin that lands on the allowlist is imported inside pytest, where a `pytest_runtest_makereport` hook can turn a failed test into a passed one. The agent is grading itself.

## Reproduction

A task graded by `python3 -m pytest`, and an oracle "agent" that does **not** solve it and only writes a dist-info into its workspace:

```bash
# solution/solve.sh — no pip, no root, task deliberately unsolved
mkdir -p /app/helper-1.0.dist-info
printf '[pytest11]\nhelper = helper\n' > /app/helper-1.0.dist-info/entry_points.txt
printf 'Metadata-Version: 2.1\nName: helper\nVersion: 1.0\n' > /app/helper-1.0.dist-info/METADATA
cat > /app/helper.py <<'EOF'
import pytest

@pytest.hookimpl(hookwrapper=True)
def pytest_runtest_makereport(item, call):
    outcome = yield
    rep = outcome.get_result()
    if rep.failed:
        rep.outcome = "passed"
        rep.longrepr = None
EOF
```

```bash
bench eval run --tasks-dir <tasks> --agent oracle --sandbox docker --sandbox-user agent
```

```
✓ Score: 1/1 (100.0%), mean reward 1.00, errors=0
```

The verifier log shows the agent's plugin on the command line the grader ran under:

```
--- verifier cwd: /app
--- PYTEST_ADDOPTS: -c /dev/null --confcutdir=/tests -p no:cacheprovider --rootdir=/app -p helper
1 passed in 0.00s
```

Same agent, same task, with only the dist-info removed:

```
✗ Score: 0/1 (0.0%), mean reward 0.00
  ✗ plugin-guard-control: reward 0.0 — 1 failed in 0.01s
```

## Two more shapes worth knowing about

**An editable install reaches the same place without forging anything.** `pip install -e .` puts a root-owned dist-info in site-packages while the project's source stays in the workspace, and adds that workspace path to `sys.path` via a `.pth`. So a dist-info ownership check — the one the docstring describes — would still admit agent-owned code. Measured in a container:

```
dist-info : /usr/local/lib/python3.12/site-packages/editable_plug-1.0.dist-info   uid=0
module    : /app/src/editable_plug/__init__.py                                    uid=1000
```

This matters for SWE-style tasks, where an editable install of the project under test is the normal setup and editing that source is the whole assignment.

**Verifiers that call the `pytest` console script don't get a wrong score — they crash.** The console script's `sys.path` doesn't include the workspace, so the injected `-p` can't resolve:

```
ImportError: Error importing plugin "helper": No module named 'helper'
```

Exit code 1 on a task that would otherwise have passed. An agent that merely unpacks a wheel into its workspace can trip this without trying to.

## Impact

Any task whose `test.sh` runs `python3 -m pytest` is exposed, and that is the plainest way to write one — no `cd`, no `PYTHONSAFEPATH`, nothing unusual. `benchmarks/hilbench/benchflow.py` runs pytest this way with `cwd=workspace`. Verifiers that first `cd` to a root-owned directory and set `PYTHONSAFEPATH=1` avoid the score flip but fall into the crash case above.

Because `verifier.user` defaults to `None`, the graded pytest process typically runs as root, so the injected plugin executes as root too.

## Expected

A pytest11 registration should reach the verifier's `-p` allowlist only when the code it resolves to is root-owned and outside every agent-writable tree — which is what the docstring already promises.

Getting there takes a little more than that wording suggests, though. Checking the dist-info alone is not enough (the editable case above), and neither is checking the module in isolation: what matters is the file pytest will actually import, since a workspace module shadowing a system plugin of the same name is what gets loaded.

Whatever the shape of the fix, it should leave image-installed plugins working without per-benchmark code changes — that is what discovery is there for.

tulerfeng:
Trust a pytest11 registration only when the code it resolves to is root-owned and outside every agent-writable tree, so an agent cannot put a plugin of its own on the verifier's `-p` allowlist.

Fixes #1116

## The problem

Discovery enumerated pytest11 entry points and passed every name straight through. The docstring said it "filters to only those whose dist-info is in a root-owned directory"; nothing in the script asked.

That gap is reachable without pip and without root. `importlib.metadata` reads `*.dist-info` off `sys.path`, and the verifier exec carries no `cwd`, so `test.sh` runs in the image WORKDIR — the agent's workspace. A dist-info written there lands on the allowlist, gets imported inside pytest, and a `pytest_runtest_makereport` hook rewrites the report.

## The fix

For each entry point, resolve its top-level module and require every path it resolves to be root-owned, outside the agent-writable prefixes, and not group- or other-writable. Three details decide whether this actually holds:

**Resolution goes through `PathFinder`, not an import.** Importing the module to find out where it lives would run the code the check is deciding whether to trust.

**Resolution uses the full `sys.path`, not a pre-filtered one.** The question is which file pytest will import, not which one it ideally would. A workspace module shadowing a system plugin of the same name is what gets loaded, so that is what has to be judged — filtering the search path first would inspect the clean system copy and clear a name that resolves elsewhere at grading time. A mutation swapping this back for a filtered path is caught by a test.

**The module is checked, not just the dist-info.** An editable install has a root-owned dist-info in site-packages and its source in the workspace, so a dist-info ownership check on its own — the one the old docstring described — still admits agent-owned code.

**A dotted target is walked one component at a time**, each step searching only the locations its parent declares. Checking the top level alone would clear `trusted_pkg.evil` on the strength of `trusted_pkg`, and a pkgutil-style namespace package widens its `__path__` at import time to every same-named directory on `sys.path` — the workspace included — which `find_spec` never sees. Walking the path makes a component that exists only in the workspace resolve nowhere.

The agent-writable prefixes are the existing `_blocked_verifier_path_prefixes`, already used for PATH hardening, so `/testbed` and the sandbox user's home come along for free.

## Testing

`tests/test_verifier_plugin_trust.py`, 13 tests. These run the real container-side script through an interpreter with a controlled `sys.path` rather than mocking `env.exec`, because a mock can show that a name was passed along, not whether it should have been. The pre-fix script is kept in the file so the contrast is asserted rather than described.

| Registration shape | Old version | PR version |
| --- | --- | --- |
| Root-installed, root-owned code | trusted | trusted |
| dist-info forged in the workspace | trusted | rejected |
| `pip install --user` into the sandbox user's home | trusted | rejected |
| Editable project, root dist-info, workspace source | trusted | rejected |
| System plugin shadowed by a workspace module | trusted | rejected |
| Root-owned but world-writable | trusted | rejected |
| Submodule of a widened namespace package, living only in the workspace | trusted | rejected |
| Root-installed `pkg.plugin`, the ordinary dotted shape | trusted | trusted |

Six cases need root to construct the root-owned tree they contrast against, so they skip on a normal dev box and run under CI. All 13 pass as root in a `python:3.12` container.

### Mutation testing

Seven guards removed one at a time, against eight registrations planted in a container so that each guard is the only thing rejecting at least one of them.

| Mutation | Result | Leaked |
| --- | --- | --- |
| M1 drop the `uid == 0` check | caught | agent-owned plugin outside the blocked prefixes |
| M2 drop the blocked-prefix check | caught | root-owned plugin inside `/testbed` |
| M3 drop the source-trust check (pre-fix behaviour) | caught | six of the seven hostile registrations |
| M4 accept when no real path was found | survived | — |
| M5 resolve against a pre-filtered `sys.path` | caught | system plugin shadowed from the workspace |
| M6 drop the group/other-writable check | caught | root-owned but world-writable plugin |
| M7 check only the top-level component | caught | submodule of a widened namespace package |

M3 leaks six of the seven because the dotted walk still stops the widened namespace package on its own — the two guards cover different routes rather than backing each other up.

M4 survives because it is unreachable: a builtin makes `PathFinder` return `None` and stops earlier, and a namespace package always reports search locations, so `not locations` has no witness in CPython today. It is left in as fail-closed cover for a shape that does not currently exist rather than removed.

### Regressions

`tests/test_sandbox_hardening.py` 100 passed. Full suite 6053 passed; the 11 failures in `tests/test_cli_live_progress.py` reproduce identically on an unmodified checkout (they assert on line wrapping and depend on terminal width). `ruff check` and `ruff format` clean.

## End-to-end verification

Real `bench eval run`, docker sandbox, `--sandbox-user agent`, on a task graded by `python3 -m pytest`. The oracle "agent" writes a dist-info into its workspace with a hook that rewrites failed reports.

| Run | Agent solved it | Old version | PR version |
| --- | --- | --- | --- |
| Plants the plugin, does not solve | no | **reward 1.00** | reward 0.00 |
| Plants nothing, does not solve | no | reward 0.00 | reward 0.00 |
| Solves and plants the plugin, image ships `pytest-cov` | yes | — | reward 1.00 |

The verifier command line before the fix:

```
PYTEST_ADDOPTS: -c /dev/null --confcutdir=/tests -p no:cacheprovider --rootdir=/app -p helper
1 passed in 0.00s
```

and after, on the third run — the image's own plugin still loaded, the agent's gone:

```
PYTEST_ADDOPTS: -c /dev/null --confcutdir=/tests -p no:cacheprovider --rootdir=/app -p pytest_cov
```

That third row is the one guarding against overcorrection: discovery replaced a hand-curated whitelist so image-installed plugins would load without per-benchmark code changes, and rejecting those would break the verifiers that need them.

## One note on scope

Verifiers that call the `pytest` console script instead of `python3 -m pytest` were not getting a wrong score — the injected `-p` failed to resolve and crashed grading with `ImportError: Error importing plugin "helper"`, exit code 1 on a task that would otherwise have passed. That goes away here too, since the name never reaches the allowlist, but it was a separate symptom of the same gap and is worth knowing about when reading old failures.

---

<a href="https://app.devin.ai/review/benchflow-ai/benchflow/pull/1117" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/src/benchflow/sandbox/lockdown.py
+++ b/src/benchflow/sandbox/lockdown.py
@@ -518,14 +518,23 @@ def _verifier_confcutdir(task: "Task") -> str:
 #              such a plugin by its dist-info alone admits agent-owned code.
 #
 # Resolving the module covers both: a forged registration resolves into the
-# workspace, and an editable project resolves to its workspace source. The
-# resolution deliberately uses the full ``sys.path`` rather than a pre-filtered
-# one, because the question is which file pytest will actually import, not
-# which one it ideally would: a workspace module shadowing a system plugin of
-# the same name is what gets loaded, so it is what has to be judged. Only the
-# top-level name is resolved, and only through ``PathFinder``, which locates
-# modules without importing them — importing here would run the very code this
-# is deciding whether to trust.
+# workspace, and an editable project resolves to its workspace source. Three
+# things about how that resolution is done are load-bearing:
+#
+#   * It goes through ``PathFinder``, which locates modules without importing
+#     them — importing here would run the very code this is deciding whether
+#     to trust.
+#   * The first component is resolved against the full ``sys.path``, not a
+#     pre-filtered one, because the question is which file pytest will
+#     actually import, not which one it ideally would: a workspace module
+#     shadowing a system plugin of the same name is what gets loaded.
+#   * A dotted target is walked one component at a time, each step searching
+#     only the locations its parent declares. Checking the top level alone
+#     would clear ``trusted_pkg.evil`` on the strength of ``trusted_pkg``,
+#     and a pkgutil-style namespace package widens its ``__path__`` at import
+#     time to every same-named directory on ``sys.path`` — the workspace
+#     included — which ``find_spec`` never sees. A component living only in
+#     the workspace then resolves nowhere and the entry point is refused.
 #
 # ``argv[1]`` is the JSON list of agent-writable path prefixes.
 _DISCOVER_PYTEST_PLUGINS_SCRIPT = r"""
@@ -563,23 +572,29 @@ try:
         eps = list(entry_points().get('pytest11', []))
     names = []
     for ep in eps:
-        top = ep.value.split(':')[0].strip().split('.')[0]
-        if not top:
+        parts = [p for p in ep.value.split(':')[0].strip().split('.') if p]
+        if not parts:
             continue
-        try:
-            spec = PathFinder.find_spec(top)
-        except Exception:
-            continue
-        if spec is None:
-            continue
-        # A namespace package has no origin, only search locations; a builtin
-        # has a non-path origin. Requiring at least one real path keeps both
-        # from passing by vacuous truth.
-        locations = [spec.origin] if (spec.origin or "").startswith("/") else []
-        locations += [str(p) for p in (spec.submodule_search_locations or [])]
-        if not locations or not all(trusted(p) for p in locations):
-            continue
-        names.append(ep.name)
+        search = None
+        for i, part in enumerate(parts):
+            try:
+                spec = PathFinder.find_spec('.'.join(parts[:i + 1]), path=search)
+            except Exception:
+                break
+            if spec is None:
+                break
+            # A namespace package has no origin, only search locations; a
+            # builtin has a non-path origin. Requiring at least one real path
+            # keeps both from passing by vacuous truth.
+            locations = [spec.origin] if (spec.origin or "").startswith("/") else []
+            locations += [str(p) for p in (spec.submodule_search_locations or [])]
+            if not locations or not all(trusted(p) for p in locations):
+                break
+            search = [str(p) for p in (spec.submodule_search_locations or [])]
+            if i < len(parts) - 1 and not search:
+                break
+        else:
+            names.append(ep.name)
     print(json.dumps(sorted(set(names))))
 except Exception as e:
     print(json.dumps({"error": str(e)}), file=sys.stderr)
--- a/src/benchflow/sandbox/lockdown.py
+++ b/src/benchflow/sandbox/lockdown.py
@@ -502,19 +502,85 @@ def _verifier_confcutdir(task: "Task") -> str:
 
 
 # Container-side script to enumerate pre-installed pytest11 entry points.
-# Runs after sandbox_user processes are killed, so the agent cannot install
-# new packages between enumeration and verification. The sandbox_user cannot
-# pip install (not root), so all discovered plugins are image-authored.
+#
+# Discovery decides what goes on the verifier's ``-p`` allowlist, so a plugin
+# reaching this list executes *inside* pytest with the power to rewrite a test
+# report. A pytest11 entry point is therefore only trusted when the code it
+# resolves to is root-owned and outside every agent-writable tree; both halves
+# are needed, and neither is implied by the other:
+#
+#   forged   — ``importlib.metadata`` reads ``*.dist-info`` straight off
+#              ``sys.path``, needing neither pip nor root, so the sandbox user
+#              can register a plugin by creating a directory in its workspace.
+#   editable — a root-installed editable project puts a root-owned dist-info in
+#              site-packages while its *source* stays in the workspace, and
+#              adds that workspace path to ``sys.path`` via a ``.pth``. Judging
+#              such a plugin by its dist-info alone admits agent-owned code.
+#
+# Resolving the module covers both: a forged registration resolves into the
+# workspace, and an editable project resolves to its workspace source. The
+# resolution deliberately uses the full ``sys.path`` rather than a pre-filtered
+# one, because the question is which file pytest will actually import, not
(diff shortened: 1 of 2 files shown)


tulerfeng:
A follow-up commit (`2fa497cf`) closes a hole in the first one.

Checking only the first dotted component cleared `trusted_pkg.evil` on the strength of `trusted_pkg`. The vehicle is a pkgutil-style namespace package: its `__init__` runs `extend_path`, which folds every same-named directory on `sys.path` into `__path__`, the agent's workspace included. That widening happens at import time, so `find_spec` never sees it — the top-level package looks wholly root-owned while the submodule pytest ends up importing exists only in the agent's tree. Measured in a container, with a root-owned `sharedns` package and an agent-authored `sharedns/evil.py`:

```
sharedns.__path__ = ['/usr/local/lib/python3.12/site-packages/sharedns', '/app/sharedns']
sharedns.evil resolved to: /app/sharedns/evil.py
discovery returned: ["evilplug", "pytest_cov"]
```

Each component is now resolved against the locations its parent declares, so a component that exists only in the workspace resolves nowhere. Discovery returns `["pytest_cov"]` on the same container, and a new test covers the ordinary `pkg.plugin` shape to make sure the walk does not refuse legitimate plugins. Mutation M7 in the table covers the regression.

The same commit adds the PR reference to the regression test docstrings, per `AGENTS.md`.

PR #702, opened 2026-06-12
feat(rewards): task-declared reward_range widens the [0,1] contract

xdotli:
## What

Adds an opt-in, **task-declared reward range** so a task can widen BenchFlow's
canonical `[0.0, 1.0]` reward contract:

```toml
[verifier]
reward_range = [-1.0, 1.0]
```

`VerifierConfig.reward_range` (default `None` = canonical strict `[0, 1]`)
parses from both `task.toml` and `task.md` frontmatter and threads through every
reward chokepoint.

## Why (a real framework feature, not a compat shim)

BenchFlow's reward contract is hard-coded `[0.0, 1.0]` in `is_valid_reward_number`
and enforced at every chokepoint — the verifier's `reward.txt`/`reward.json`
parse, `validate_reward_map`, `apply_aggregate_policy`, and rollout's final
`_ensure_canonical_rewards` gate. That is correct for accuracy-style benchmarks,
but **RL / safety benchmarks need sub-zero rewards**: a deliberate safety
violation should score *below* doing nothing (`0.0`), which is core to their
thesis. Today such a run becomes a verifier **ERROR** instead of a scored
negative reward, and the `[verifier]` table is `extra="forbid"` so a task cannot
even declare a wider range.

This is a general framework capability (RL/penalty benchmarks), not downstream
glue — the default is unchanged and strict.

## Design: widen-only, explicit, zero default change

- **Widen-only**, validated once at config parse by
  `validate_declared_reward_range`: `lo <= 0.0`, `hi >= 1.0`, `lo < hi`, both
  finite. This keeps the canonical anchors meaningful for *every* task — `0`
  ("did nothing") and `1` ("solved") stay valid rewards, so no-op and oracle
  baselines keep scoring and a range can never silently shift or narrow the
  contract.
- The range threads as an **explicit `reward_range` parameter** (resolved once
  per entry point via `declared_reward_range(task)`) through
  `is_valid_reward_number`, `validate_reward_map`, `apply_aggregate_policy`
  (incl. the strict consistency check), the three `Verifier` reward-parse paths,
  and rollout's `_ensure_canonical_rewards(..., task=...)` gate. No hidden
  global state; the contract is visible at every call site and consistent with
  how the existing `lenient` flag is threaded.
- **Defaults are byte-identical** — no behavior change unless a task declares a
  range. It composes with but does not imply lenient mode (lenient still never
  accepts negatives on its own). LLM-judge **rubric scores stay `[0, 1]`**
[...]

PR #1121, opened 2026-09-12
fix(metrics): make collect_metrics selection predicate shape-aware fo…

VaisakhiMishra:
Fix for the known follow-up identified in #1096.

`collect_metrics` ranked any reward dictionary as scoreable using a combination of checks that had two problems: booleans were accepted (because `isinstance(True, int)` is `True` in Python), and the `[0, 1]` range bound from `is_valid_reward_number` was applied even though `collect_metrics` reads `result.json` files and never sees a `TaskConfig` — so a task that declares `reward_range = [-1.0, 1.0]` would have a legitimate `-1.0` score silently reclassified as "no valid reward", causing selection and reporting to disagree on the same artifact.

### What was wrong

The selection predicate in `collect_metrics` used `r.get("rewards") is not None` (clause 2) and `r.get("rewards")` truthiness (clause 3) to decide whether a candidate artifact had a scoreable reward. This passed for any truthy non-`dict` value (`rewards: 1.0`, `rewards: True`, `rewards: []`) and crashed `_safe_reward` — which called `.get()` on a non-dict — in clause 3. The downstream reward extraction used `r.get("rewards", {}).get("reward") if r.get("rewards")`, which hit the same crash.

Three concrete failure modes:

1. **Non-dict rewards displace valid scores.** `rewards: 1.0` passed `is not None`, entered the scored tier, and could beat a well-formed `rewards: None` result — moving the outcome from errored to spuriously scored.
2. **Boolean rewards count as passed.** `{"reward": True}` passed the old `isinstance(val, (int, float))` check (because `isinstance(True, int)` is `True`), was returned as a score, and `classify_result` then evaluated `True == 1.0` — counting the task as **passed**.
3. **Empty dict displaces valid zero.** `{"rewards": {}}` had no `reward` key, `_safe_reward` returned `0.0` as a fallback, which tied with a valid `reward: 0.0` and kept the first-seen (possibly malformed) artifact.

### Fix

**`src/benchflow/metrics.py`**

- Make the selection predicate `isinstance(..., dict)`-aware so only well-formed dict rewards enter the scored tier (clauses 2 and 3).
- Replace `_safe_reward`'s `isinstance(val, (int, float))` guard with two explicit checks that match what the function actually needs here:
  - `isinstance(val, bool)` — rejects `True`/`False` (the `isinstance(True, int)` trap)
  - `math.isfinite(val)` — rejects `inf`, `-inf`, `nan`
- **No `[0, 1]` range bound is applied.** `collect_metrics` reads `result.json` files only and never sees a `TaskConfig`, so it cannot know whether a task declared a widened `reward_range`. Applying the canonical bound would silently reclassify a legitimate `-1.0` score (from a safety-floor task declaring `reward_range = [-1.0, 1.0]`) as "no valid reward" — and `extract_reward`, which the reporting half of the same function uses, applies no bound at all, so selection and reporting would disagree on the same artifact.
- Replace the downstream `r.get("rewards", {}).get("reward") if r.get("rewards")` extraction with the canonical `extract_reward(r)` helper (already used by `classify_score_outcome`) so selection winner and `TaskMetrics.reward` share one contract.

**`tests/test_metrics.py`**

Five regression tests added:
- `test_collect_metrics_malformed_rewards_does_not_win_over_none` — `rewards: 1.0` must not beat `rewards: None` and must not crash
- `test_collect_metrics_malformed_rewards_does_not_beat_zero_reward` — `rewards: 1.0` must not displace a well-formed `rewards: {"reward": 0.0}`
- `test_collect_metrics_boolean_reward_treated_as_invalid` — `{"reward": True}` must not beat `{"reward": 0.0}`
- `test_collect_metrics_nonfinite_reward_treated_as_invalid` — tests `_safe_reward` directly: `inf`, `-inf`, `nan`, `True`, `False`, `None` all return `None`; `-1.0`, `0.0`, `0.5`, `1.0`, `2.0` all return the scalar (no range bound)
- `test_collect_metrics_empty_rewards_dict_does_not_beat_valid_zero` — `{}` must not tie with `{"reward": 0.0}`

### Verification

- `uv run ruff format --check src tests tools`
- `uv run ruff check src tests tools`
- `uv run ty check src/`
- `pytest tests/test_metrics.py`
- Full suite → same 2 pre-existing failures, no head-only failures

### DIFF
--- a/src/benchflow/metrics.py
+++ b/src/benchflow/metrics.py
@@ -335,7 +335,11 @@ def _safe_reward(rewards: dict | None) -> float | None:
     if not isinstance(rewards, dict):
         return None
     val = rewards.get("reward")
-    return float(val) if is_valid_reward_number(val) else None
+    # Use isinstance narrowing that type checkers can follow, then validate
+    # with is_valid_reward_number (rejects booleans, non-finite, out-of-range).
+    if not isinstance(val, (int, float)) or not is_valid_reward_number(val):
+        return None
+    return float(val)
 
 
 def collect_metrics(
--- a/src/benchflow/metrics.py
+++ b/src/benchflow/metrics.py
@@ -320,7 +320,7 @@ class BenchmarkMetrics:
         }
 
 
-def _safe_reward(rewards: dict) -> float | None:
+def _safe_reward(rewards: dict | None) -> float | None:
     """Extract a validated reward scalar from a rewards dict, or None if invalid.
 
     Uses the canonical is_valid_reward_number validator so that booleans,
--- a/src/benchflow/metrics.py
+++ b/src/benchflow/metrics.py
@@ -323,7 +323,11 @@ def _safe_reward(rewards: dict) -> float:
 
     Prevents TypeError when comparing reward values where one is None
     (e.g. rewards={"reward": None, "rubric": [...]}).
+    Only called when rewards is already confirmed to be a dict; the isinstance
+    guard here is a defensive belt-and-braces for any future caller.
     """
+    if not isinstance(rewards, dict):
+        return 0.0
     val = rewards.get("reward")
     return val if isinstance(val, (int, float)) else 0.0
 
@@ -348,10 +352,13 @@ def collect_metrics(
             task = r["task_name"]
             if (
                 task not in best
-                or (r.get("rewards") is not None and best[task].get("rewards") is None)
                 or (
-                    r.get("rewards")
-                    and best[task].get("rewards")
+                    isinstance(r.get("rewards"), dict)
+                    and not isinstance(best[task].get("rewards"), dict)
+                )
+                or (
+                    isinstance(r.get("rewards"), dict)
+                    and isinstance(best[task].get("rewards"), dict)
                     and _safe_reward(r["rewards"]) > _safe_reward(best[task]["rewards"])
                 )
             ):
@@ -361,7 +368,9 @@ def collect_metrics(
 
     tasks = []
     for task_name, r in sorted(best.items()):
-        reward = r.get("rewards", {}).get("reward") if r.get("rewards") else None
+        reward = (
+            r["rewards"].get("reward") if isinstance(r.get("rewards"), dict) else None
+        )
         # Calculate duration
         duration = 0.0
         try:
--- a/src/benchflow/metrics.py
+++ b/src/benchflow/metrics.py
@@ -16,9 +16,11 @@ from benchflow._utils.scoring import (
     classify_error,
     classify_score_outcome,
     classify_verifier_error,
+    extract_reward,
     pass_rate,
     pass_rate_excl_errors,
 )
+from benchflow.rewards.validation import is_valid_reward_number
 from benchflow.trajectories.metrics import result_skill_invocations
 from benchflow.usage_tracking import UsageSource, is_trusted_usage_source
 
@@ -318,18 +320,22 @@ class BenchmarkMetrics:
         }
 
 
-def _safe_reward(rewards: dict) -> float:
-    """Extract reward value from a rewards dict, defaulting to 0 if None/missing.
+def _safe_reward(rewards: dict) -> float | None:
+    """Extract a validated reward scalar from a rewards dict, or None if invalid.
 
-    Prevents TypeError when comparing reward values where one is None
-    (e.g. rewards={"reward": None, "rubric": [...]}).
-    Only called when rewards is already confirmed to be a dict; the isinstance
-    guard here is a defensive belt-and-braces for any future caller.
+    Uses the canonical is_valid_reward_number validator so that booleans,
+    non-finite values, and out-of-range values are all excluded — matching
+    exactly the contract that classify_score_outcome / extract_reward enforce
+    during reporting. This keeps selection and reporting on the same contract
+    so the same artifact is ranked and classified consistently.
+
+    Returns None (not 0.0) for any invalid or absent reward so the caller
+    can distinguish "no valid reward" from "reward is zero".
     """
     if not isinstance(rewards, dict):
-        return 0.0
+        return None
     val = rewards.get("reward")
-    return val if isinstance(val, (int, float)) else 0.0
+    return float(val) if is_valid_reward_number(val) else None
 
 
 def collect_metrics(
@@ -350,16 +356,15 @@ def collect_metrics(
         try:
             r = json.loads(rfile.read_text())
             task = r["task_name"]
+            r_reward = _safe_reward(r.get("rewards"))
+            best_reward = _safe_reward(best.get(task, {}).get("rewards"))
             if (
                 task not in best
+                or (r_reward is not None and best_reward is None)
                 or (
-                    isinstance(r.get("rewards"), dict)
-                    and not isinstance(best[task].get("rewards"), dict)
-                )
-                or (
-                    isinstance(r.get("rewards"), dict)
-                    and isinstance(best[task].get("rewards"), dict)
-                    and _safe_reward(r["rewards"]) > _safe_reward(best[task]["rewards"])
+                    r_reward is not None
+                    and best_reward is not None
+                    and r_reward > best_reward
                 )
             ):
                 best[task] = r
@@ -368,9 +373,7 @@ def collect_metrics(
 
     tasks = []
     for task_name, r in sorted(best.items()):
-        reward = (
-            r["rewards"].get("reward") if isinstance(r.get("rewards"), dict) else None
-        )
+        reward = extract_reward(r)
         # Calculate duration
         duration = 0.0
         try:
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -563,3 +563,121 @@ def test_collect_metrics_malformed_rewards_does_not_beat_zero_reward(tmp_path):
     assert s["failed"] == 1, "well-formed rewards:{'reward':0.0} must be kept as failed"
     assert s["passed"] == 0
     assert s["errored"] == 0
+
+
+def test_collect_metrics_boolean_reward_treated_as_invalid(tmp_path):
+    """Guards Devin review: isinstance(True, (int,float)) is True in Python,
+    so _safe_reward previously accepted boolean rewards. is_valid_reward_number
+    explicitly rejects booleans — selection and reporting must agree."""
+    import json
+
+    # attempt1: boolean reward (True) — invalid per is_valid_reward_number
+    t1 = tmp_path / "a1" / "task__t1"
+    t1.mkdir(parents=True)
+    (t1 / "result.json").write_text(
+        json.dumps(
+            {
+                "task_name": "task",
+                "rewards": {"reward": True},
+                "error": None,
+                "n_tool_calls": 0,
+                "started_at": "2026-01-01 00:00:00.000000",
+                "finished_at": "2026-01-01 00:01:00.000000",
+            }
+        )
+    )
+    # attempt2: well-formed reward=0.0 (scored, failed)
+    t2 = tmp_path / "a2" / "task__t2"
+    t2.mkdir(parents=True)
+    (t2 / "result.json").write_text(
+        json.dumps(
+            {
+                "task_name": "task",
+                "rewards": {"reward": 0.0},
+                "error": None,
+                "n_tool_calls": 1,
+                "started_at": "2026-01-01 00:02:00.000000",
+                "finished_at": "2026-01-01 00:03:00.000000",
+            }
+        )
+    )
+
+    metrics = collect_metrics(str(tmp_path))
+    s = metrics.summary()
+    assert s["failed"] == 1, "well-formed reward=0.0 must win over boolean reward=True"
+    assert s["passed"] == 0
+    assert s["errored"] == 0
+
+
+def test_collect_metrics_nonfinite_reward_treated_as_invalid(tmp_path):
+    """Guards Devin review: inf/-inf/nan are rejected by is_valid_reward_number
+    but isinstance(inf, float) is True, so old _safe_reward accepted them."""
+    import json
+
+    t1 = tmp_path / "a1" / "task__t1"
+    t1.mkdir(parents=True)
+    (t1 / "result.json").write_text(
+        json.dumps(
+            {
+                "task_name": "task",
+                "rewards": {"reward": None},
+                "error": "agent error",
+                "n_tool_calls": 0,
+                "started_at": "2026-01-01 00:00:00.000000",
+                "finished_at": "2026-01-01 00:01:00.000000",
+            }
+        )
+    )
+    # inf cannot be serialised directly to JSON; simulate via a post-load patch
+    # by writing a valid file then checking the validator directly
+    from benchflow.rewards.validation import is_valid_reward_number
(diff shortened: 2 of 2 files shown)


tulerfeng:
Thanks for taking this on. The case you added for `{"reward": true}` is a real find — on both `main` and #1096 a boolean reward makes the task count as **passed**, and nothing else in either branch catches it. I ran an independent end-to-end validation of this PR, including a real `bench eval run` on a task that declares a widened reward range.

Short version: this PR fixes two things #1096 does not — a boolean reward counting as passed, and an empty `rewards` dict displacing a valid `0.0`. It also adds a `[0, 1]` bound that #1096 does not have, and that bound reclassifies a scored `-1.0` rollout as an infrastructure error for any task declaring `[verifier] reward_range`. Deleting the bound is a four-line edit that leaves all five of your new cases, all six of #1096's, and the full suite green. The A/B for both halves is below.

## End-to-end verification

merge-base `3b9dd067` (previous version) vs `d69ff9c9` (PR version), with `12c1d578` (#1096) as the third reference point. `metrics.py` has not moved on `main` since the merge-base, so all three trees start from the same file.

| what was verified | result |
|---|---|
| #1096's 6 cases run on this branch | all pass — this PR subsumes the crash fix |
| this PR's 5 cases run on #1096 | 2 are new behaviour, 3 already hold there |
| real `bench eval run`, task declaring `reward_range = [-1.0, 1.0]` | verifier scores `-1.0`, reported as `Failed 1` before, `Errored 1` on this branch |
| real `bench eval metrics` A/B across 4 trees | only this branch loses the scored artifact |
| mutation: drop the range bound | all 5 new cases still pass — the bound is unpinned |
| full suite, 4 trees | same 12 pre-existing failures, no head-only failures |

**This PR subsumes #1096's crash fix.** I copied #1096's two new test functions (6 cases) onto this branch: all 6 fail on the merge-base and all 6 pass here. `bench eval metrics` survives a non-mapping `rewards` on this branch without #1096.

**Two of the five new cases are genuinely new.** Running this PR's tests against #1096's tree:

| case | merge-base (previous version) | #1096 | this PR |
|---|---|---|---|
| `boolean_reward_treated_as_invalid` | fail | fail | **pass** |
| `empty_rewards_dict_does_not_beat_valid_zero` | fail | fail | **pass** |
| `malformed_rewards_does_not_win_over_none` | fail | pass | pass |
| `malformed_rewards_does_not_beat_zero_reward` | pass | pass | pass |
| `nonfinite_reward_treated_as_invalid` | pass | pass | pass |

The boolean case is the one I would make sure survives the merge. On both the merge-base and #1096, an artifact carrying `{"reward": true}` is reported as **passed**: `isinstance(True, (int, float))` is `True`, so `_safe_reward` returns it and `extract_reward` hands it on, and `classify_result` then evaluates `True == 1.0`. A verifier that writes a JSON boolean scores the task as solved. This PR is the only branch that stops it.

## One clause worth revisiting: tasks that declare a widened reward range

`is_valid_reward_number(val)` is called without a `reward_range`, so it applies the canonical `[0, 1]`. That is not the contract for every task — #702 added `[verifier] reward_range` precisely so safety-floor tasks can score `-1.0`, deliberately below doing nothing. On this branch a legitimate `-1.0` reads as "no valid reward" in the selection predicate.

I verified this with real rollouts rather than hand-written fixtures. A task declaring `reward_range = [-1.0, 1.0]`, whose verifier writes `-1.0` for a deliberate violation, run twice into one jobs directory:

```
bench eval run --trials 1 --tasks-dir <safety-floor-task> \
  --agent gemini --model gemini-3.5-flash-lite --sandbox docker   # attempt 1: errored
bench eval run --trials 1 --tasks-dir <safety-floor-task> \
  --agent oracle --sandbox docker                                 # attempt 2: scored
```

Both are real Docker rollouts. Attempt 1 errored on a missing provider key; attempt 2 ran the oracle, the verifier wrote `-1.0` to `reward.txt`, and the run reported `mean reward -1.00, errors=0`. The two `result.json` files:

```
safety-floor-task__4f1f03df   rewards=None            error="ValueError: GEMINI_API_KEY required ..."
safety-floor-task__6fa7ca23   rewards={'reward': -1.0}  error=None
```

Then `bench eval metrics` over that jobs directory, same input to all four trees:

| tree | Failed | Errored | artifact kept | error breakdown |
|---|---|---|---|---|
| merge-base `3b9dd067` (previous version) | 1 | 0 | scored `-1.0` | — |
| `12c1d578` (#1096) | 1 | 0 | scored `-1.0` | — |
| **`d69ff9c9` (PR version)** | **0** | **1** | **the errored attempt** | **`{'other': 1}`** |
| PR version without the range bound | 1 | 0 | scored `-1.0` | — |

A rollout that really was scored at the safety floor is reported as an infrastructure error. It also leaves the `pass_rate_excl_errors` denominator, so a real failure stops counting against the score, and it contributes an error to the breakdown that never happened. Nothing raises, so this stays silent.

On how much this matters today: I could not find a task in the repo that currently declares `reward_range`, so nothing in tree is affected right now — I had to write the task above to exercise it. The reason I am still raising it is that the feature is shipped and covered by `tests/test_reward_range.py`, safety-floor tasks are its stated purpose, and no test in either PR would catch this if such a task arrives later. Your call on whether that is worth handling now or after.

**Selection and reporting end up disagreeing.** `extract_reward`, which the reporting half of this same function uses, applies no range at all — a single `-1.0` artifact with no retry is still reported as `failed` with reward `-1.0` on this branch. So `-1.0` is a score for reporting and not a score for selection, which is the opposite of what the new docstring describes.

**Passing the correct range is not available here.** `collect_metrics` reads `result.json` files and nothing else; it never sees a `TaskConfig`. Every other caller in the repo that validates a *persisted* reward does pass the range explicitly — `rollout/_setup.py:578` uses `reward_range=declared_reward_range(task)`, `verifier_core.py:116` the same, and `rewards/validation.py` threads it through all ten of its own call sites. The bare calls without a range are the ones scoring judge and rubric sub-scores, which are canonically `[0, 1]`.

**No test holds the range bound in place.** Dropping it and keeping the bool and non-finite checks:

```python
if isinstance(val, bool) or not isinstance(val, (int, float)):
    return None
if not math.isfinite(val):
    return None
return float(val)
```

With that in place: all 5 of this PR's new cases pass, all 6 of #1096's pass, `tests/test_metrics.py` is 17/17, `ruff check` / `ruff format --check` / `ty check` are clean, and the `-1.0` row above goes back to `Failed 1`. So nothing in the suite depends on the bound, and removing it costs none of the behaviour the PR adds.

On where the bound came from: `c45f0659` was shape-aware only and had no bound; `ac64d882` switched to `is_valid_reward_number` eleven minutes after the bot review. The bot asked for booleans, non-finite values, and values outside "the configured range". The first two are what this function needs; the range came along with the helper's default because there is no configured range to pass here.

## Two notes on the tests

**Two of the five pass on unmodified `main`.** `nonfinite_reward_treated_as_invalid` never calls `collect_metrics` — it writes a `result.json`, ignores it, and asserts on `is_valid_reward_number` directly, so it pins the validator rather than this change. `malformed_rewards_does_not_beat_zero_reward` passes on the merge-base because the `_safe_reward` crash happens inside the enclosing `try`, is logged as "Skipping corrupt result file", and the well-formed artifact survives by accident. Both are fine to keep; neither distinguishes this branch from `main`.

That same `nonfinite` test carries `assert not is_valid_reward_number(-1.0), "-1.0 must be invalid (below range)"` — true for the default range, not true for a task that declares one.

## A note on the description

The opening line — "Fix for the known follow up identified in #1096" — describes this PR accurately. Everything from the second paragraph on looks like it belongs to a different change: removing a `{"reward": 0.0}` injection block from `rollout/__init__.py` and `continue_run/orchestrator.py` for the agent-timeout plus silent-verifier case, and adding `test_agent_timeout_silent_verifier_counts_as_errored` to `tests/test_job.py`. None of that is in this diff, which touches only `metrics.py` and `tests/test_metrics.py`.

I am raising it so a reviewer who reads the description before the diff does not go looking for changes that are not there. Separately, the change that section describes still applies: the injection block is present and unchanged on `main` at `rollout/__init__.py:2207-2213` and `continue_run/orchestrator.py:694-700`, and `test_agent_timeout_silent_verifier_counts_as_errored` does not exist anywhere in the repo, so that one is still open if you want to send it.

## Potential conflict with PR #1096

Both PRs rewrite `_safe_reward` from the same merge-base, so whichever lands second needs a manual resolution. GitHub shows both as mergeable because it checks each against `main`, not against the other:

```
$ git merge --no-commit --no-ff pr1121   # onto 12c1d578
Auto-merging src/benchflow/metrics.py
CONFLICT (content): Merge conflict in src/benchflow/metrics.py
Auto-merging tests/test_metrics.py
```

The test files merge cleanly — no name collisions — so the only decision is which `_safe_reward` body survives.

One consequence of the ordering: #1096 also changes `skill_eval/_core.py:741`, which is the `bench skills eval` half of the same payload and is not touched here. This PR satisfies every assertion #1096 makes in `metrics.py`, so it would be easy to read #1096 as superseded once this lands — but that one line exists only there, and `bench skills eval` still raises without it.

## Full suite

| tree | failed | passed | skipped |
|---|---|---|---|
| merge-base `3b9dd067` (previous version) | 12 | 5991 | 48 |
| `12c1d578` (#1096) | 12 | 6002 | 48 |
| `d69ff9c9` (PR version) | 12 | 5996 | 48 |
| PR version without the range bound | 12 | 5996 | 48 |

The same 12 failures on all four trees — 11 in `test_cli_live_progress.py` and one in `test_task_check_eval_consistency.py`. They reproduce on the untouched merge-base and on current `main`, so neither PR introduces them. No head-only failures. The passed deltas are exactly the new cases each PR adds: +11 for #1096, +5 for this one.

## Suggestion

Swap the `is_valid_reward_number(val)` call in `_safe_reward` for the four lines above, then merge. That removes only the `[0, 1]` bound and keeps both checks the bot review asked for, so the boolean and empty-dict cases this PR adds still hold, and `-1.0` goes back to being read as a score. I ran that combination against your five cases, #1096's six, `tests/test_metrics.py`, the full suite, `ruff` and `ty` — all green.

If you would rather keep the helper, `is_valid_reward_number(val, reward_range=(-math.inf, math.inf))` has the same effect, though it reads as working around the signature rather than stating what the function means.

Second, replacing the description below the first line with what the diff does would save the next reviewer the cross-check.

The `-1.0` behaviour above was reproduced through real `bench eval run`s in real Docker containers — a real verifier writing a real `reward.txt`, then the real `bench eval metrics` CLI reading the artifacts back — rather than from constructed fixtures.

VaisakhiMishra:
@tulerfeng Thanks for the thorough end-to-end validation and the detailed breakdown — the `reward_range` finding in particular was right and has been addressed (the `[0, 1]` bound is now dropped in favour of explicit bool + non-finite checks only).

Before resolving the merge conflicts manually, I want to check whether this PR still makes sense to land as-is. Looking at the diff against the current `upstream/main`, `main` has moved significantly since the merge-base `3b9dd067`:

1. **`_result_rank` replaces the three-clause predicate entirely.** `main` now has a `_result_rank(result) -> tuple[bool, bool, float]` function that ranks by `(reward is not None, outcome == "passed", reward_value)` and reduces `collect_metrics` to a single `if task not in best or _result_rank(r) > _result_rank(best[task])` call. Our `_safe_reward` + explicit clauses is a parallel solution to the same problem.

2. **`iter_task_result_paths` replaces `rglob("result.json")`.** Our branch still uses `rglob`.

3. **`TaskMetrics.scoring` field was added on `main`** and this branch dropped it.

Two questions before I spend time on the resolution:

- Does `_result_rank` on `main` already handle the boolean and non-finite cases this PR fixes? Specifically: `extract_reward` returns `rewards.get("reward")` without a bool check, so `{"reward": True}` still returns `True`, and `True == 1.0` in `classify_result` still counts it as **passed** — meaning the `_result_rank` approach on `main` has the same boolean bug. If that's confirmed, the fix still needs to land somewhere, and the cleanest path would be a targeted patch to `_result_rank` (or `extract_reward`) on `main` rather than rebasing the full `_safe_reward` approach.

- Should this PR be closed in favour of a smaller follow-up that patches `_result_rank` directly on top of the current `main`? That would be a 3–4 line change with no conflict surface, targeting exactly the boolean and non-finite gaps, and would preserve `iter_task_result_paths`, `TaskMetrics.scoring`, and the rest of `main`'s refactored shape.

Happy to send that as a fresh PR if that's the preferred path.

VaisakhiMishra:
@tulerfeng — update on all the open threads:

## Your suggestion: drop the range bound - done

The `[0, 1]` bound is gone. `_valid_reward_scalar` now applies exactly the two identified checks — explicit `isinstance(val, bool)` rejection and `math.isfinite(val)` — with no range. All three bullet points verified locally:

- All 3 of this PR's regression tests pass
- `tests/test_metrics.py` full
- `ruff format --check` / `ruff check` / `ty check` all clean
- `-1.0` from a task declaring `reward_range = [-1.0, 1.0]` correctly stays in the `Failed` bucket

## Rebase: conflict resolved by adopting `_result_rank`

Question about whether `_result_rank` on `main` already handles the boolean case: **it does not**. Confirmed on `upstream/main` at `6b99a10e` — `extract_reward` returns `rewards.get("reward")` with no bool guard, so `{"reward": true}` returns `True`, `_result_rank` places it in the scored tier (`reward is not None = True`), and `classify_result` evaluates `True == 1.0` as passed. The bug is present on current `main`.

Rather than a straight rebase (which conflicted on every one of the 5 interim commits), the branch has been rebased cleanly onto `upstream/main` by adopting `_result_rank`'s shape and patching it directly:

- Added `_valid_reward_scalar(value) -> bool` to `metrics.py` — the single place that rejects booleans, non-finite values, and non-numerics
- Patched `_result_rank` to use it for both the tier bool and the float fallback
- Patched `extract_reward` in `_utils/scoring.py` with an explicit bool guard so selection and reporting are on the same contract
- `iter_task_result_paths`, `TaskMetrics.scoring`, and the rest of `main`'s refactored shape are all preserved — the diff is now minimal

The old `_safe_reward` + three-clause predicate approach is gone entirely from this branch.

## Description

Updated to describe only what the diff actually does (`metrics.py`, `_utils/scoring.py`, `tests/test_metrics.py`). The silent-verifier / reward-injection change (the block that was incorrectly in the earlier description) is a separate fix that still needs to go in; will send it as a follow-up PR.

## The two CI failures — both pre-existing on `upstream/main`, not introduced here

**`test_manifest_parity` — antigravity manifest missing**
Upstream commit `6b99a10e` added `antigravity` to `registry.AGENTS` without a corresponding `manifest.toml` in `benchflow-ai/agents`. The `test_omnigent_pi_is_sole_core_unmanifested_agent` gate fails on every open PR targeting `main` right now. This PR does not touch `registry.py` or any agent config.

**`pip-audit` — editable project entry in `uv export`**
`uv export` now includes the local editable project as an `-e` requirement; `pip-audit` cannot determine a package version from it. The fix is `--no-emit-project` on the `uv export` call in `.github/workflows/test.yml`, which this PR does not touch. Fails on every open PR targeting `main`.

Both reproduce on `upstream/main` with this diff reverted.

Can you re-review the PR please? Thanks.

